PR #128, opened 2014-11-29
Use system keyring by default when creating .oscrc

gnb:
No more storing passwords in plaintext in ~/.oscrc unless
the user explicitly requests it.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -137,16 +137,12 @@ class Osc(cmdln.Cmdln):
         except oscerr.NoConfigfile as e:
             print(e.msg, file=sys.stderr)
             print('Creating osc configuration file %s ...' % e.file, file=sys.stderr)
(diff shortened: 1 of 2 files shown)


adrianschroeter:
the various keyring implementation out there were quite broken in the past. Esp. the gnome keyring under non-Gnome environements broke very often at random times. While it worked sometimes in the beginning it suddenly stopped working, confused users and made debugging quite hard.

adrianschroeter:
since we run maintenance updates from osc git master, I do not want to have the default behaviour changed here. We should do so with osc2 instead IMHO.

Yes, storing plain text passwords on disk is not ideal, but other scm tools did the same so far. and when having access to home user partition another security issue exists anyway. Most likely you can also steal the  password anyway when the user is using it the next time. It basically only helps on stolen hard disks (which is a good reason for enabling it of course).

PR #133, opened 2015-01-09
Fix readline support in cmdln-module

bergmannf:
This was already fixed in the original module:
https://github.com/trentm/cmdln/issues/1

While this might not impact out-of-the-box `osc` client, it is useful if a plugin wants to provide an interactive interface, which is how I stumbled over it.

### DIFF
--- a/osc/cmdln.py
+++ b/osc/cmdln.py
@@ -393,47 +393,63 @@ class RawCmdln(cmd.Cmd):
         """
         self.cmdlooping = True
         self.preloop()
-        if intro is None:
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2015-01-09 04:42:47 -0800, Florian Bergmann wrote:
> While this might not impact out-of-the-box `osc` client, it is useful if a
> plugin wants to provide an interactive interface, which is how I stumbled
> over it.

Yes, this codepath is not used by osc.
Merged. Thanks!

PR #140, opened 2015-03-25
Command line handling: accept rq accept --submit-devel

DimStar77:
This will do the opposite of --no-devel and forward to a devel project
without furher asking, automatically superseding existing requests.

Comes in handy in our workflow, where we develop unstable GNOME in GNOME:Next, then submit it to GNOME:Factory and in some cases need to directly submit it to openSUSE:Factory.

Doing a large batch of submit requests, not being asked to forward every single one is mandatory

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -1109,9 +1109,13 @@ class Osc(cmdln.Cmdln):
             value = root.findtext('attribute/value')
             if value and not opts.yes:
                 repl = ''
-                print('\n\nThere are already following submit request: %s.' % \
-                      ', '.join([str(i) for i in myreqs ]))
-                repl = raw_input('\nSupersede the old requests? (y/n) ')
+                if opts.submit_devel:
+                    # In case the user passed --submit-devel, we don't ask anymore
+                    repl = 'y'
+                else:
+                    print('\n\nThere are already following submit request: %s.' % \
+                          ', '.join([str(i) for i in myreqs ]))
+                    repl = raw_input('\nSupersede the old requests? (y/n) ')
                 if repl.lower() == 'y':
                     myreqs += [ value ]
 
@@ -1865,6 +1869,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                   help='output the diff in the unified diff format')
     @cmdln.option('--no-devel', action='store_true',
                   help='Do not attempt to forward to devel project')
+    @cmdln.option('--submit-devel', action='store_true',
+                  help='Attempt to forward to devel project if defined, superseding any existing request')
     @cmdln.option('-m', '--message', metavar='TEXT',
                   help='specify message TEXT')
     @cmdln.option('-t', '--type', metavar='TYPE',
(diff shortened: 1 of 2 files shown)


adrianschroeter:
"--submit-devel" is not sounding right to me, it would also have an effect if no devel package is set.

also it sounds quite bad to me to do this automatic, what if the other request is doing further actions?

If you do not want an interactive action, you should maybe do an "osc copypac -K" instead?

DimStar77:
if there is no devel packaget set, --submit-devel has no effect (as it intercepts the queries: this package is developed at ...; do you want to forward)

why would it be bad to do such things automatic (on request of the user, as it's not default).

the GNOME Team's workflow is:
- package all things in GNOME:Next (unstable)
- submit to GNOME:Factory (during rc phase)
- submit from G:F to oS:F (rc phase, for staging integration)
--
in this time, the stable release rolls into GNOME:Next
- team submits all to GNOME:Factory from GNOME:Next
- once all is ready to accept, all pending SRs on GNOME:Factory are going to be accepted, need to be forwarded to openSUSE:Factory and superseding the existing requests from the rc phase (if it does not supersede, I don't care - factory-auto will :) )

osc copypac really does not fit in the workflow of package reviews prior to accepting.

adrianschroeter:
still, superseding unknown requests blindly may fit your/gnomes/factory workflow, but not the rest. 
given theat esp. gnome and factory are using anyway a bunch of external hacks, I do not want to get our standard code get infected by that.

Can't you solve this issue in your other scripts by heck for open requests first?
How do you handle open requests submiting exact the the same source? (killing the requests and running reviews and restart?)

DimStar77:
no too many unknown requests can be there:
the only accepted source to openSUSE:Factory is from the devel project; so also the one from the devel project would be the only one superseding it... but looking outside of the scope of oS:F, I can imagine this happening.

Same versions won't ever happen: why would this even be passing a review and be accepted into GNOME:Factory? But, again, I see this is too specific.

But then: it's a non-defautt option, not impacting a user not passing it to rq accept. Does not sound THAT intrusive

adrianschroeter:
Still, I consider "--submit-devel" as misleading. A --forward option makes sense and would be analog to the webui.

doing a second, IMHO quite bad thing together in the same action is not a good idea (superseding blindly). 

Please either implement it in your special scripts or let's discuss how we can support the gnome/factory workflow in a generic way. (currently I have no clue how you work frankly).

Btw. it was once a requirement that side submissions to factory MUST be possible for mass fixes...

DimStar77:
if --forward is acceptable, I'll gladly change the option name... 
submissions to Factory, by bypassing the devel project have cause a lot of grieve in the past and should really never ever be done. current factory-auto explicitly declines any request not coming from a devel project (and rightly so; stepping on maintainers feet is not going to give us happy contributors).

Let's see how we can approach each other to find a workable solution without me having to carry local patches forever.

marcus-h:
Short update: after a small discussion on IRC, the following workaround
can be used to "mimic" this patch:
yes y | osc rq accept ...

PR #141, opened 2015-03-26
fix TypeError

bmwiedemann:
TypeError: cannot concatenate 'str' and 'list' objects


marcus-h:
I just fixed it in commit eda2e75655eef53eb4504acbb94d94958aef79db
(using ' '.join(cmd) instead of str(cmd)).
Thanks for pointing this out!

PR #142, opened 2015-04-08
added a simple way to verify osc code base with the python security scanner bandit

vpereira:
openstack released a tool named bandit, a security scanner. I tested osc and the results looks relevant. 

more information about the tool can be found here https://wiki.openstack.org/wiki/Security/Projects/Bandit

I packaged it and you can find it here: https://build.opensuse.org/package/show/home:vpereirabr/python-bandit

the output from the osc scan, can be found here https://gist.github.com/vpereira/f54afa4ea11bb41c1a72

I integrated it in a obs project - using brp - and at build time i scan the python code.. it works.

### DIFF
--- a/run_bandit.sh
+++ b/run_bandit.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+# you can pass as argument "csv","json" or "txt" (default)
+if [ "$1" != "" ];then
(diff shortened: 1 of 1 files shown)


adrianschroeter:

Issue #39, opened 2013-07-03
osc 0.139.2: SSL Server Name Indication does not work


mattiascockburn:
Heya,

i am in the process of creating a public OBS instance for fun and proft (learning experience).

As i have only one public ip available i set up the different vhosts for my obs in nginx with ssl and am using sni.

The sites are working fine from a modern browser (chromium). I am using StartSSL as a CA which is in my local cert store.

When i am trying to connect to the api through osc i get a general ssl validation error.

I tried it with a patched version of m2crypto -> http://stackoverflow.com/questions/7961547/does-m2crypto-have-client-side-server-name-indication-sni-support but the error prevails.

It seems that the ssl code of osc is overloading some functions of m2crypto, but i am not sure where to begin to troubleshoot the problem.

PR #145, opened 2015-04-25
Add support for non-root paths in API URLs.

OlegGirko:
Currently osc supports API URLs without path only,
like https://api.example.com (if there is pathname, it's just ignored).
With this change API URLS with path, like https://example.com/api
are supported correctly.

This is useful for those who can't have OBS api, webui and main site
on different domains and buy separate X.509 certificates for them,
or just can't afford to have separate IP addresses for them
(please note that currently osc doesn't support TLS SNI).

Signed-off-by: Oleg Girko <ol@infoserver.lv>

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -382,23 +382,38 @@ cookiejar = None
 
 def parse_apisrv_url(scheme, apisrv):
     if apisrv.startswith('http://') or apisrv.startswith('https://'):
-        return urlsplit(apisrv)[0:2]
+        return urlsplit(apisrv)[0:3]
     elif scheme != None:
         # the split/join is needed to get a proper url (e.g. without a trailing slash)
-        return urlsplit(urljoin(scheme, apisrv))[0:2]
+        return urlsplit(urljoin(scheme, apisrv))[0:3]
     else:
         msg = 'invalid apiurl \'%s\' (specify the protocol (http:// or https://))' % apisrv
         raise URLError(msg)
 
 
-def urljoin(scheme, apisrv):
-    return '://'.join([scheme, apisrv])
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2015-04-24 20:09:40 -0700, Oleg Girko wrote:

<SNIP>

> (please note that currently osc doesn't support TLS SNI).
> 
That's a m2crypto issue (see https://github.com/openSUSE/osc/issues/39).

I just had a quick look your patch: IMHO, we should move the apiurl
"extraction" code from the core.py into its own method in conf.py
Then, this can be used by "is_known_apiurl" and by the "_build_opener"
function (in order to simplify the code in core.py again).
What do you think?

OlegGirko:
@marcus-h:
I've pushed a new version of my changes.

> we should move the apiurl "extraction" code from the core.py into its own method in conf.py

Known apiurl extraction code has been moved into its own `extract_known_apiurl()` function in `conf` module, according to your suggestion.

> Then, this can be used by "is_known_apiurl"

Although it's possible to change `is_known_apiurl()` to use `extract_known_apiurl()` (extracting apiurl and checking that it's equal to the original url), it's terribly inefficient, so I've decided to leave it as is.

> and by the "_build_opener" function

Moving call to `extract_known_apiurl()` into `_build_opener()` brings no improvement: `install_opener()` with the result of `_build_opener()` is called only if we already checked that there is apiurl corresponding to given url, and this check was positive. If there is already this check outside of `_build_opener()` that returns apiurl as a result, it's reasonable to not repeat the same check and extraction inside `_build_opener()`, but rather have it expecting known apiurl already.

adrianschroeter:

PR #151, opened 2015-05-10
avoid buffering stdout output on pipes (bnc#930137)

bmwiedemann:
This change is slightly tested and seems to fix the case of | tee 

https://bugzilla.opensuse.org/show_bug.cgi?id=930137

### DIFF
--- a/osc-wrapper.py
+++ b/osc-wrapper.py
@@ -5,6 +5,7 @@
 
 import locale
 import sys
+import os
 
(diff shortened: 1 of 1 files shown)


marcus-h:
Thanks! Merged. See also the comment in #bnc/bsc930137 and
commit 56a9ab60dc77f5dcbf1b29b7846c09f8d80c138c.

PR #152, opened 2015-05-13
When lxml fails to parse the _service file show error message instead of backtrace

jelly:
Print a user friendly error when there is an XML syntax error in the _service file when committing.


adrianschroeter:

PR #156, opened 2015-06-24
Fix completion for scp

bmwiedemann:
https://bugzilla.opensuse.org/show_bug.cgi?id=915037

Tested that it can complete commands and project+subproject names


marcus-h:
Merged. Thanks Bernhard and Werner!

PR #157, opened 2015-06-29
Remove hardcoded download.opensuse.org reference

jplautz:
Currently the repourls function is hardcoded to only return repourls pointing at http://download.opensuse.org/repositories/.  Update the function to pull the download_url field from the /configuration api and use it if it is defined, otherwise continue to return download.opensuse.org


jplautz:
marcus beat me to it

PR #161, opened 2015-07-21
Fix misuses of time.gmtime

andreas-schwab:


marcus-h:
Merged. Thanks!
One question about commit 71297e31f45b9da40a2b7cc862cd22bfd38cae1f:
What was wrong with the old implementation? The only issue I see
so far is that it computed the wrong build time if the build took
more than 31 days.

andreas-schwab:
The broken-down time is not suitable for time differences.  It's a representation of wall-clock time, not stop-watch time.

PR #162, opened 2015-07-29
fix an "400 Bad Request" when running against old API

seife:
Updating a whole project against an old API server (experienced with
2.3.5) leads to:
	Server returned an error: HTTP Error 400: Bad Request
	unknown parameter 'nofilename'
So just retry without nofilename if an 400 is thrown.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -3688,6 +3688,9 @@ def get_project_sourceinfo(apiurl, project, nofilename, *packages):
     try:
         si = show_project_sourceinfo(apiurl, project, nofilename, *packages)
     except HTTPError, e:
+        # old API servers do not know the 'nofilename' parameter, so retry without
+        if e.code == 400 and nofilename:
+            return get_project_sourceinfo(apiurl, project, False, *packages)
         if e.code != 414:
             raise
         if len(packages) == 1:


adrianschroeter:
not sure if we really want to make the code uglier just to support this ancient OBS version ..... why don't you use an old unmaintained osc as well? ;)

seife:
Because the old unmaintained osc is not available for SLES12 for example.
And even if it was, it would always be updated by system patches. Locking out updates is nothing I can suggest to my enterprise customers :-)

Another way would be the API announcing its capabilities and the client only using the capabilities that are available, but that solution was way beyond my non-existent python capabilities.

marcus-h:
On 2015-07-29 08:09:57 -0700, Stefan Seyfried wrote:
> Another way would be the API announcing its capabilities and the client only using the capabilities that are available, but that solution was way beyond my non-existent python capabilities.
> 
The api already (implicitly) announces its capabilities by sending the
"x-opensuse-apiversion" header, but atm osc fails to use this information.
What we could do is the following:
- for each api release x.y.z, we generate a python api_x_y_z module out of
  the existing ruby api (information like routes etc. are almost present,
  but, for example, allowed parameters cannot be easily derived yet)
- in case of a new api release, we generate wrappers for the old existing
  python api modules:
  * unchanged api calls (same "signature") are delegated to the old api
    module
  * for changed api calls, we can manually specify compat code (if possible)
- osc only uses the api exposed by the latest generated api_x_y_z module

The advantage is that the compat code is not part of the osc "core" code
anymore, but moved to a place where it actually belongs.
This might be a cool feature for osc2.

Apart from that, I merged your PR (IMHO, this change does not harm - we
have more insane compat code in osc... ;) ).

PR #164, opened 2015-08-12
Support prefer-pkg for live-build

gollub:


marcus-h:
Merged. Thanks for the PR!

PR #168, opened 2015-08-15
Add Travis CI build for Python 3.3, 3.4, 3.5-dev

bluca:
Hello,

In case it can be useful, this commit adds CI builds with Python 3. Thanks in part to the container-based infrastructure, tests across all versions are started and ran in parallel, so there is no additional time to wait for results. Travis supports 3.2, 3.3, 3.4 and release candidate builds of 3.5 (scheduled to be release in a month, so it's probably a good idea to start testing on it).

Tests on 3.2 are failing (is it supported?), so 3.2 is not included, see https://travis-ci.org/bluca/osc/jobs/75774509 for the log if you are interested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: python
 python:
   - "2.7"
+  - "3.3"
+  - "3.4"
+  - "3.5-dev"
 addons:
   apt:
     packages:


marcus-h:
On 2015-08-15 16:56:07 -0700, Luca Boccassi wrote:
> Tests on 3.2 are failing (is it supported?), so 3.2 is not included, see https://travis-ci.org/bluca/osc/jobs/75774509 for the log if you are interested.

Hmm I would call the whole python3 support "experimental". I just merged
your PR - this way we can easily see if we committed a python3 "incompatible"
change.

bluca:
Perfect, thanks!

PR #169, opened 2015-08-17
fix exception

lnussel:
TypeError: can only concatenate list (not "tuple") to list


lnussel:
important to fix https://travis-ci.org/openSUSE/osc-plugin-factory/builds/75913340

PR #173, opened 2015-09-16
Correction of grammatical errors

jengelh:


marcus-h:
Merged. Thanks! ("exists" is no python keyword:) )

jengelh:
```
osc/build.py:                if not os.path.exists(destdir):
osc/commandline.py:        if exists:
osc/commandline.py:        if not exists and (srcprj != args[0] or srcpkg != args[1]):
```
Well yeah maybe not a _keyword_, but nevertheless a commonly-observed function and variable name.

PR #174, opened 2015-09-18
Add --from-csv argument to osc-results

yaccz:
Example usage and output:

% ./osc-wrapper.py search --package hexedit | ./osc-wrapper.py results --stdin
Results of hexedit in home:adra
openSUSE_Tumbleweed  i586       succeeded
openSUSE_Tumbleweed  x86_64     succeeded
openSUSE_13.2        i586       succeeded
openSUSE_13.2        x86_64     succeeded

Results of hexedit in home:ajorgensen
RedHat_RHEL-6        i586       succeeded
RedHat_RHEL-6        x86_64     succeeded
RedHat_RHEL-5        i586       succeeded
RedHat_RHEL-5        x86_64     succeeded

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -12,6 +12,7 @@ import sys
 import time
 import imp
 import inspect
+import itertools
 try:
     from urllib.parse import urlsplit
     from urllib.error import HTTPError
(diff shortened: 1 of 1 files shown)


marcus-h:
First of all thanks for the patch!
However, the patch is quite "complicated" and not really pythonic.
For example,
+        if opts.from_csv:
+          xs = map(lambda x: x[:-1], sys.stdin.readlines())
+          xs = map(lambda x: tuple(x.split("|")), xs)

Could be written as
xs = [line.rstrip().split('|') for line in sys.stdin.readlines()]
(which is, IMHO, more readable). In general, list comprehensions
(see https://docs.python.org/2/tutorial/datastructures.html#tut-listcomps)
or generators are preferred over maps:)

Also, we could avoid the "NResultsPrintJoin" class.

IMHO, a straightforward implementation would just collect the
project, project/package data and then simply iterates over
the collected data, queries the api and prints the results.

What do you think?

yaccz:
Comprehensions yes but I don't want to rewrite the print join as that would require more invasive refactoring I don't want to get into.

yaccz:
What do you think?

yaccz:
Hello, I've pushed new code which should be more reasonable. Let me know what you think, please.

marcus-h:
On 2015-10-23 10:04:23 -0700, Jan Matějka wrote:
> Hello, I've pushed new code which should be more reasonable. Let me know what you think, please.
> 
First of all, sorry for the late response. I'll review your
changes at the end of the week (at the latest).

PR #184, opened 2016-01-24
initial documentation skeleton

cav71:
Hi,
this will start a documentation for osc the python library

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -1320,11 +1320,38 @@ class Package:
             ET.SubElement(root, 'entry', name=i, md5=todo_send[i])
         return root
 
+    @staticmethod
+    def commit_filelist(apiurl, project, package, filelist, msg='', user=None, **query):
+        """send the commitlog and the local filelist to the server"""
+        if user is None:
+            user = conf.get_apiurl_usr(apiurl)
+        query.update({'cmd': 'commitfilelist', 'user': user, 'comment': msg})
+        u = makeurl(apiurl, ['source', project, package], query=query)
+        f = http_POST(u, data=ET.tostring(filelist, encoding=ET_ENCODING))
+        root = ET.parse(f).getroot()
+        return root
+
+    @staticmethod
+    def commit_get_missing(filelist):
+        """returns list of missing files (filelist is the result of commit_filelist)"""
+        error = filelist.get('error')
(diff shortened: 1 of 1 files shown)


cav71:
Forgtten to mention:
 1. needs sphinx installed
 2. the command to generate the documentation is python setup.py build_docs

marcus-h:
First of all it's great to see that someone cares about docs!:)

Is there a specific reason why you want to use sphinx? IMHO, we
could achieve something similar (wrt. the api docs), if we use
pydoc (see, e.g., pydoc.writedoc and pydoc.writedocs).
The advantage of pydoc is that we don't need an additional dependency,
because it is part of the standard lib (in general new dependencies
are no problem, but one has to keep in mind that we build osc for
several distributions (which probably ship different versions
etc.)).

cav71:
Hi,
I think osc can be used to provide binary builds for the wheel packages and in general for python packages: in order to do that documentation is the first step in promoting osc to a wider audience.

The specific reason for sphinx is it has become the "standard" python documentation tool and the comunity at large expect documentation have some sort of consistency these days: on top of my mind I can recollect any project not using sphinx these days.

I hope this helps

cav71:
There's a readthedocs page for the built doc: http://osccav71.readthedocs.org/en/latest/

marcus-h:
On 2016-01-31 14:56:30 -0800, cav71 wrote:
> The specific reason for sphinx is it has become the "standard" python documentation tool and the comunity at large expect documentation have some sort of consistency these days: on top of my mind I can recollect any project not using sphinx these days.
> 
Ok - let's give it a try:)
Can squash your two commits into one (or at least the parts that
affect the setup.py script)?

cav71:
Hi,
just created a new pull request with a single commit (I hope, I'm still getting my head around git).

I hope this helps

Marcus Hüwe wrote:

marcus-h:
On 2016-02-06 14:00:10 -0800, cav71 wrote:
> just created a new pull request with a single commit (I hope, I'm still getting my head around git).
> 
Thanks - btw. instead of creating a new pull request, you could simply push
your local changes to the remote branch (the one that is "associated" with
this PR) again (git push --force...). The advantage is that we have all the
discussions at a "single" place:)

PR #186, opened 2016-01-31
Support build-type='none' for non-Linux systems

raimue:
osc can be used for interaction with a remote build service on platforms
other than Linux, for example OS X. However, local builds are not
supported, so we want to fail gracefully if the package maintainer
patched the default configuration to build-type = 'none'.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -5833,6 +5833,10 @@ Please submit there instead, or use --nodevelproject to force direct submission.
 
         import osc.build
 
+        if conf.config['build-type'] and conf.config['build-type'] == "none":
+            print('Error: build command has been disabled for this platform, cannot run local builds')
+            return 1
+
         if which(conf.config['build-cmd']) is None:
             print('Error: build (\'%s\') command not found' % conf.config['build-cmd'], file=sys.stderr)
             print('Install the build package from http://download.opensuse.org/repositories/openSUSE:/Tools/', file=sys.stderr)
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2016-01-31 07:44:52 -0800, Rainer Müller wrote:
> osc can be used for interaction with a remote build service on platforms
> other than Linux, for example OS X. However, local builds are not
> supported, so we want to fail gracefully if the package maintainer
> patched the default configuration to build-type = 'none'.

Hmm we could do something like that, but isn't it sufficient if we just
rely on the absense of the /usr/bin/build (or whatever is configured in
the config) script? In that you "just" need to patch error message, e.g.:

index de3a6d5..2fb1cc2 100644
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -5834,8 +5834,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         import osc.build
 
         if which(conf.config['build-cmd']) is None:
-            print('Error: build (\'%s\') command not found' % conf.config['build-cmd'], file=sys.stderr)
-            print('Install the build package from http://download.opensuse.org/repositories/openSUSE:/Tools/', file=sys.stderr)
+            print('Error: local build is not supported', file=sys.stderr)
             return 1
 
         if opts.debuginfo and opts.disable_debuginfo:

raimue:
Fair enough, I will follow your proposal then.

PR #190, opened 2016-03-02
fix 1591

ketheriel:
Fix OBS 1591 (openSUSE/open-build-service#1591)


ketheriel:
Fix OBS 1591 (openSUSE/open-build-service#1591)

PR #193, opened 2016-04-08
switch to new obsinfo service

adrianschroeter:
Hi Marcus,

this is an incompatible change I want to have on purpose (but want to hear a second opinion therefore).

It changes the used service when adding a new git repo via "osc add git:...".

I think this incompatible change is okay, because it affects only users who add a new git repo and I think the command is not widely used that so far.

This will help us a lot with disk storage on the server in future and it will help developers to work inside their git check outs (detailed docu will follow).

So I think we can risk this change and could add a config option to revert later if really needed. (But I do not believe we do).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,5 @@
 0.154
+  - switch to new obs_scm service when adding git URL's
   - set OSC_VERSION environment for source services
 
 0.153
--- a/osc/core.py
+++ b/osc/core.py
@@ -343,21 +343,27 @@ class Serviceinfo:
 
     def addSetVersion(self, serviceinfo_node):
         r = serviceinfo_node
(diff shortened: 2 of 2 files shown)


adrianschroeter:
the new obs_scm service is available via openSUSE:Tools project atm only. I plan to make it available via maintenance update together with osc update later.

marcus-h:
On 2016-04-08 00:16:32 -0700, Adrian Schröter wrote:
> this is an incompatible change I want to have on purpose (but want to hear a second opinion therefore).
> 
No objections from my side.
 
> So I think we can risk this change and could add a config option to revert later if really needed. (But I do not believe we do).

Yep - I just merged it for now.

PR #197, opened 2016-05-01
- babysitter.run: decode entities in error summary

andreas-schwab:


marcus-h:
Oops... I somehow overlooked this PR - sorry for the late
response. Btw., it's probably more sane to simply parse
the xml via etree instead of this manual xml "parsing",
but that's a different story... :)

Merged. Thanks!

PR #200, opened 2016-05-15
Fix obsolete warnings for commandline

sebix:
Remove unreplaced `${cmd_name}` in prints (same format as with other warnings) and actually print the warnings.

This is a possible output with the current version:
```bash
> osc deletepac libXScrnSaver
${cmd_name} is obsolete !
...
```

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -450,7 +450,7 @@ class Osc(cmdln.Cmdln):
     def do_addchannels(self, subcmd, opts, *args):
         """${cmd_name}: Add channels to project.
 
-        The command adds all channels which are defined to be used for a given source package. 
+        The command adds all channels which are defined to be used for a given source package.
         The source link target is used to lookup the channels. The command can be
         used for a certain package or for all in the specified project.
 
@@ -1703,7 +1703,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         args = slash_split(args)
 
         apiurl = self.get_api_url()
-        
+
         i = 0
         actionsxml = ""
         supersede = []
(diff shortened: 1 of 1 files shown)


sebix:
I updated the PR and added much more unifications and fixes, such as return values, help for commands and printing to stderr.

osc/commandline.py
```diff
@@ -4876,9 +4895,18 @@ def do_results_meta(self, subcmd, opts, *args):
...
+
+        See the help output of 'results'.
+        """
+
+        print("Command rresults is obsolete. Running 'osc results' instead",
+              file=sys.stderr)
+        print("See 'osc help results'.", file=sys.stderr)
         self.do_results('results', opts, *args)
-        sys.exit(1)
+        return 1
```
sleep-walker: I'd be consistent here and return 2 as an invalid argument return value.

sebix: Thanks, fixed

marcus-h:
First of all thanks for the patch! (and sorry for the long delay...)

Some minor remarks:
- do_deletepac: we should print the message to stderr as well
- do_results_meta: now the message is doubled - I think we can kill
  the old
  print("Command results_meta is obsolete. Please use: osc results --xml")
  line
- do_rresults: instead of returning 1, we could also do
  "return self.do_results('results', opts, *args)" (since we execute
  do_results, returning 1 might be misleading).

What do you think?

sebix:
Thanks for your feedback @marcus-h, I incorporated it and pushed it.

marcus-h:
On 2016-06-16 10:53:41 -0700, Sebastian wrote:
> Thanks for your feedback @marcus-h, I incorporated it and pushed it.
> 
Great!
Can you squash your changes into one commit? Then, we are ready to
merge:)

sebix:
@marcus-h also done :)

marcus-h:
On 2016-06-16 15:06:54 -0700, Sebastian wrote:
> @marcus-h also done :)
> 
Merged. Thanks a lot!

PR #207, opened 2016-06-01
add "osc service runall"

adrianschroeter:
This is up for discussion, but I try to simplify the interface here:
- obsolete "localrun" and "disabledrun". Still implemented, but not anymore
  recommended and documented. can hopefully removed later.
- "runall" is running all services local, also buildtime services
  - files get replaced by the service
  - can be used to get final tar balls to be used with tools like quilt

I am not 100% sure about this, please give a second thought as well...

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,7 @@
 0.155
-  -
+  - osc service runall: runs all service local, independend of the mode
+                        This allows to create local files even when disabled or
+                        just created at buildtime
 
 0.154
   - switch to new obs_scm service when adding git URL's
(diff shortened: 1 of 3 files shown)


marcus-h:
On 2016-06-01 08:07:38 -0700, Adrian Schröter wrote:
> This is up for discussion, but I try to simplify the interface here:
> - obsolete "localrun" and "disabledrun". Still implemented, but not anymore
>   recommended and documented. can hopefully removed later.
> - "runall" is running all services local, also buildtime services
>   - files get replaced by the service
>   - can be used to get final tar balls to be used with tools like quilt
> 
Looks good to me.

> I am not 100% sure about this, please give a second thought as well...

Simplifying the service ui is great! To be honest, I usually had to read
the code first before deciding which "osc service <mode>" I actually
needed;)

Ps. sorry for the late response...

PR #208, opened 2016-06-02
Add revision control options to meta project commands

kkeil:
Since the API controller functions are now fixed and support this, add revision control options to the osc meta calls
`osc meta prj|prjconf -r|--revision REV`
to checkout any revision and
`osc meta prj|prjconf [-m|--message TEXT] [-e] [-F file]`
to add an optional commit message.
It is compatible to older API releases.
Tested with 0.154.0 because of the broken master.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -740,6 +740,10 @@ class Osc(cmdln.Cmdln):
     @cmdln.option('-F', '--file', metavar='FILE',
                         help='read metadata from FILE, instead of opening an editor. '
                         '\'-\' denotes standard input. ')
+    @cmdln.option('-r', '--revision', metavar='REV',
+                        help='checkout given revision instead of head revision. For prj and prjconf meta only')
+    @cmdln.option('-m', '--message', metavar='TEXT',
+                        help='specify log message TEXT. For prj and prjconf meta only')
     @cmdln.option('-e', '--edit', action='store_true',
                         help='edit metadata')
     @cmdln.option('-c', '--create', action='store_true',
@@ -774,6 +778,8 @@ class Osc(cmdln.Cmdln):
         the --file switch. If the argument is '-', input is taken from stdin:
         osc meta prjconf home:user | sed ... | osc meta prjconf home:user -F -
 
+        For meta prj and prjconf updates optional commit messages can be applied with --message.
+
         When trying to edit a non-existing resource, it is created implicitly.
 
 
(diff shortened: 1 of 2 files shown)


marcus-h:
Hmm I'm a bit puzzled about the change in show_project_meta - excerpt:

+def show_project_meta(apiurl, prj, rev=None):
+    if rev:
+        query = {}
+        query['rev'] = rev
+        url = makeurl(apiurl, ['source', prj, '_project', '_meta'], query)

The _meta file is versioned in a tree that is tracked in the _project.mrev
file instead of the _project.rev file. Hence, if the above call reaches the
backend, it should yield a 404 (see getfile in bs_srcserver). Something
like "query = {'meta': '1'}" should fix it.

However, it seems that the above call doesn't reach the backend at all (at
least when talking with api.o.o):

marcus@linux:~> curl -u Marcus_H https://api.opensuse.org/source/home:Marcus_H/_project/_meta
Enter host password for user 'Marcus_H':
<status code="unknown_package">
  <summary>home:Marcus_H/_project</summary>
</status>
marcus@linux:~>

IMHO, this looks as if it is routed to update_package_meta instead of
get_file (see source_controller.rb) (although, the routes.rb looks ok
to me).

So the questions are:)
- did this ever work for you?
- do I miss something obvious?

Ps. sorry for the late response...

marcus-h:
On 2016-06-08 18:51:25 +0200, Marcus Hüwe wrote:
> Hmm I'm a bit puzzled about the change in show_project_meta - excerpt:
> 
> +def show_project_meta(apiurl, prj, rev=None):
> +    if rev:
> +        query = {}
> +        query['rev'] = rev
> +        url = makeurl(apiurl, ['source', prj, '_project', '_meta'], query)
> 
> The _meta file is versioned in a tree that is tracked in the _project.mrev
> file instead of the _project.rev file. Hence, if the above call reaches the
> backend, it should yield a 404 (see getfile in bs_srcserver). Something
> like "query = {'meta': '1'}" should fix it.
> 
Ok... my fault: it is routed to getpackage instead of getfile... so the
code is correct.

> However, it seems that the above call doesn't reach the backend at all (at
> least when talking with api.o.o):
> 
> marcus@linux:~> curl -u Marcus_H https://api.opensuse.org/source/home:Marcus_H/_project/_meta
> Enter host password for user 'Marcus_H':
> <status code="unknown_package">
>   <summary>home:Marcus_H/_project</summary>
> </status>
> marcus@linux:~>
> 
> IMHO, this looks as if it is routed to update_package_meta instead of
> get_file (see source_controller.rb) (although, the routes.rb looks ok
> to me).
> 
b696773 is deployed on api.o.o, which doesn't have routes.rb fix.

> So the questions are:)
> - did this ever work for you?

Now I'm pretty sure it did/does

> - do I miss something obvious?
> 
Of course... ;)

Sorry for the confusion and thanks for the patch!

PR #209, opened 2016-06-12
return different return codes for different errors

sleep-walker:
babysitter eats all interesting reasons why osc exited and returns just 3 values:
0 for success, 1 for error and 2 for wrong parameters/options

This patch is attempt to address the problem, I tried to select some sane values, organize it a bit but you may have better ideas.

### DIFF
--- a/osc/babysitter.py
+++ b/osc/babysitter.py
@@ -78,7 +78,7 @@ def run(prg, argv=None):
         return 1
     except KeyboardInterrupt:
         print('interrupted!', file=sys.stderr)
-        return 1
+        return 130
     except oscerr.UserAbort:
         print('aborted.', file=sys.stderr)
         return 1


marcus-h:
On 2016-06-12 12:48:29 -0700, Tomáš Čech wrote:
> babysitter eats all interesting reasons why osc exited and returns just 3 values:
> 0 for success, 1 for error and 2 for wrong parameters/options
> 
> This patch is attempt to address the problem, I tried to select some sane values, organize it a bit but you may have better ideas.

Hmm we could do something like this. Do you have a specific use case
where you need the return values? For example, in case of an IOError
one might be more interested in the errno value instead of the mere
fact that an IOError occurred. Or why is it helpful to distinguish
an IOError from an OSError?
(Note: I'm not arguing against your patch - I'm just trying to understand
the use case:) ).

sleep-walker:
Thank you for your review.

> Hmm we could do something like this. Do you have a specific use case where you need the return values? For example, in case of an IOError one might be more interested in the errno value instead of the mere fact that an IOError occurred. Or why is it helpful to distinguish an IOError from an OSError?

I came to osc code when I wasn't able to distinguish between osc failure and keyboard interrupt in return code for my wrapper.

Then I spotted that all the information is available but unused so I made generic patch. I'd rather provide more information than less and let user to use it or ignore. I agree that printing errno for IOError could be useful but not for me. There could also be error message for ExtRuntimeError or print answer when BadStatusLine happens.

Would you rather me to create simple patch with changing return code for KeyboardInterrupt only?

marcus-h:
On 2016-06-13 10:53:56 -0700, Tomáš Čech wrote:
> Thank you for your review.
> 
> > Hmm we could do something like this. Do you have a specific use case where you need the return values? For example, in case of an IOError one might be more interested in the errno value instead of the mere fact that an IOError occurred. Or why is it helpful to distinguish an IOError from an OSError?
> 
> I came to osc code when I wasn't able to distinguish between osc failure and keyboard interrupt in return code for my wrapper.
> 
> Then I spotted that all the information is available but unused so I made generic patch. I'd rather provide more information than less and let user to use it or ignore. I agree that printing errno for IOError could be useful but not for me.

The same might also apply for HTTPError - it could be useful to return
the http status code.

> There could also be error message for ExtRuntimeError or print answer when BadStatusLine happens.
> 
Definitely.

> Would you rather me to create simple patch with changing return code for KeyboardInterrupt only?
> 
Yes, let's start with this simple case first. If one needs more fine
grained error codes, we should reconsider this (e.g., think about the
grouping again and which information we could encode in the retval).
On the one hand, this approach is less "user-friendly" (provide less
information), but on the other hand, chances are small that we break
existing scripts if we introduce the more fine-grained error codes.

What do you think?

sleep-walker:
I'm fine with the incremental approach even though I don't believe that anyone sane checks `$? -eq 1` instead of `$? -ne 0`. I'll prepare another patch.

PR #210, opened 2016-06-23
Add instructions for installing requirements

hayderimran7:
Added how to install the pre-requirements . These are recommended imho because i had several issues related to versioning and openssl settings when installing the requirements from pip. 
Installing from zypper worked like charm after adding the suse oss repo.

### DIFF
--- a/README
+++ b/README
@@ -7,6 +7,19 @@ Patches can be submitted via
  * or the official Git repository on Github:
    https://github.com/openSUSE/osc
 
+PRE-REQUISITES:
+
+Prior to installation, following pre-requistes are required to be installed.  
+  * libcurl-devel
+  * libcurl4
+  * python-pycurl
+  * python-urlgrabber
+  * python-m2crypto
+
+To install the requirements:  
+  $ zypper ar http://download.opensuse.org/distribution/13.2/repo/oss/suse/ python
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2016-06-22 17:40:54 -0700, Imran Hayder wrote:
> Added how to install the pre-requirements . These are recommended imho because i had several issues related to versioning and openssl settings when installing the requirements from pip. 
> Installing from zypper worked like charm after adding the suse oss repo.

Hmm why do you need the curl stuff? Installing python-urlgrabber and
python-M2Crypto (which is actually optional) should be sufficient.

hayderimran7:
@marcus-h `libcurl4` is needed of python-pycurl

marcus-h:
On 2016-06-23 14:33:56 -0700, Imran Hayder wrote:
> @marcus-h `libcurl4` is needed of python-pycurl
> 
Yes, python-pycurl is required by python-urlgrabber. Installing
python-urlgrabber via zypper will automatically install python-pycurl
etc. (at least on a recent Tumbleweed installation).

What concrete problem did you face?

lethliel:
This is now very old. There is no python-urlgrabber anymore and this should no longer be a problem with newer distributions.

PR #213, opened 2016-07-02
Return different return code on KeyboardInterrupt

sleep-walker:
As I haven't found any good source for choice of exit code, I took value
130 found here:

http://tldp.org/LDP/abs/html/exitcodes.html


marcus-h:
Merged. Thanks!

PR #214, opened 2016-07-05
mention "key" in no-verify help text

bmwiedemann:
to make it easier to find with
man osc | grep key


sayalilunkad:
+1

marcus-h:
Merged. Thanks!

PR #216, opened 2016-07-28
Check for existance of .oscrc before collecting alias

olafhering:
Signed-off-by: Olaf Hering <olaf@aepfle.de>


marcus-h:
Merged. Thanks!

Issue #220, opened 2016-08-16
misspelling of 'separate' in --seperate-requests flag


jmcdough:
'seperate' is a misspelling of 'separate', so the flag name should be changed or made a synonym.

marcus-h:
Fixed via b9a53e3527506e3e9327de7faa24433f4ae18893 and
b5679dee818cb1300f1a955fb1c93f9c56962b2f.

PR #221, opened 2016-08-16
Fix spelling of --separate-requests

jmcdough:
Just a simple spelling fix.  Issue #220


marcus-h:
Merged. Thanks! For now, we also support the wrong spelling
(just for backward compatibility), but it'll be removed in the
future (see commit b5679dee818cb1300f1a955fb1c93f9c56962b2f).

PR #223, opened 2016-08-17
get_prefer_pkgs: don't ignore debuginfo packages

andreas-schwab:
This has been part of the original implementation, but it doesn't make
sense.  It should be possible to build with debuginfo packages belonging
to the preferred packages, if you want to debug something in the build
root.


marcus-h:
On 2016-08-17 00:57:54 -0700, Andreas Schwab wrote:
> This has been part of the original implementation, but it doesn't make
> sense.

Yes. Merged. Thanks!

PR #225, opened 2016-08-22
travis: Automatically deploy tags on pypi

### LINKED ISSUE #183: Upload to PyPI
Is it possible to upload this to PyPI - https://pypi.python.org/pypi/osc ?

ping @toabctl

toabctl:
osc should be available on pypi for easy installation via pip.
And travis can automatically deploy on pypi so do the setup for this.

Closes #183

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,21 @@
 language: python
 python:
-  - "2.7"
-  - "3.3"
-  - "3.4"
-  - "3.5-dev"
+- '2.7'
+- '3.3'
+- '3.4'
+- 3.5-dev
 addons:
   apt:
     packages:
     - diffstat
 sudo: false
 script: cd tests; python suite.py
+deploy:
+  provider: pypi
+  user: suse
+  password:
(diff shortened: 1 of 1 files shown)


marcus-h:
Is there anything else that needs to be done apart from merging this
PR? What is this user/password pair in the travis.yml? I assume it
it isn't possible that a random person uses these credentials to
submit a new osc release to pypi, right? (I have to admit that I
haven't considered the travis docs yet...)

toabctl:
@marcus-h I think there is nothing else needed. You need to tag a release and then it should work. The username ```suse``` is i.e. also used for the [kiwi project](https://github.com/SUSE/kiwi/blob/master/.travis.yml) .
And it's not possible to use the username/password pair for others. See https://docs.travis-ci.com/user/encryption-keys/

marcus-h:
On 2016-08-22 23:00:49 -0700, Thomas Bechtold wrote:
> And it's not possible to use the username/password pair for others. See https://docs.travis-ci.com/user/encryption-keys/
> 
Thanks for the explanation. Merged. Let's see what happens:)
Thanks a lot!

PR #227, opened 2016-09-08
Fix exception handling in osc my

kaltsi:
The exception handler did nothing and one side effect was that
using osc my --user unknownuser would dump out a huge list
of open requests.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -6894,6 +6894,9 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                     # skip it ... try again with old style below
                     pass
 
+                # other errors are handled elsewhere
+                raise e
+
         res = get_user_projpkgs(apiurl, user, role_filter, exclude_projects,
                                 'project' in what, 'package' in what,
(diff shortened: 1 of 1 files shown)


marcus-h:
Hmm with this patch the old-style/backward compatible code wouldn't be
executed anymore in case of status 400 (hmm... maybe it is time to get
rid of this codepath). I just fixed it in commit
e9cc00098738462e0bdc7327231c7473b162739e.

kaltsi:
Hmm wouldn't the 'pass' in 400 case mean that it leaves the exception block and goes to the backward compatible code?

But your fix is probably clearer.

marcus-h:
On 2016-09-13 03:55:45 -0700, Juha Kallioinen wrote:
> Hmm wouldn't the 'pass' in 400 case mean that it leaves the exception block and goes to the backward compatible code?
> 
No, "pass" is just a no operation. (The previous code just catched all
HTTPError exceptions.)

kaltsi:
All right, learned something :)

PR #228, opened 2016-09-09
FIX results watch returns in blocked state

ssedlmaier:
At the moment only the project status is taken into account when determining when to stop watching a build.
This leads to wrong behavior when a package is in 'blocked' for a longer time.
In this state the project status and code is 'published' but the package remains at 'blocked'.
With this additional check this problem is fixed.


marcus-h:
Merged. Thanks a lot! Btw., I just pushed a small follow-up
commit (07e0628d9eb7853c5daef8915bf4b1c5058b32a0), which checks that
"pkg" is not None.

ssedlmaier:
I'm glad I could help!

PR #230, opened 2016-09-21
Add 'wipe' subcommand to delete buildroots

dwmh:
Uses the new --wipe option of build to delete build-roots.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -6105,6 +6105,10 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                   help='Path to the buildroot')
     @cmdln.option('-o', '--offline', action='store_true',
                   help='Use cached data without contacting the api server')
+    @cmdln.option('--wipe', action='store_true',
+                  help='Delete the build root instead of chrooting into it')
+    @cmdln.option('-f', '--force', action='store_true',
+                  help='Do not ask confirmation for wipe')
     def do_chroot(self, subcmd, opts, *args):
         """${cmd_name}: opens a shell inside of the build root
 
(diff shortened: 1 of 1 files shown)


dwmh:
Corresponding build pull request:
https://github.com/openSUSE/obs-build/pull/295

dwmh:
Added missing command line completion

marcus-h:
Hmm I'm not sure if this deserves an own command. What about adding
a "--wipe" option to the build command (and drop the "--roots" option)?

+            else:
+                # Delete default build root.
+                build_root = conf.config['build-root']
+                while '%' in build_root:
+                    build_root = build_root[:build_root.rfind('/')]
+                build_roots.append(build_root)

IMHO, that's dangerous. Just consider the build-root
"/home/user/build-root-%(project)-%(repo)-%(arch)"... in this case
it would delete /home/user.

dwmh:
I guess we could change it to just fail if current dir is not a package dir, then we can remove that piece. We are using the build-root validation, so we never really ran into this issue with our configuration.

I think the build command already has too much stuff in it, so I'd rather have wipe as a separate command. After all, it does not really build anything.

marcus-h:
On 2016-09-21 04:29:24 -0700, Matias Hilden wrote:
> I think the build command already has too much stuff in it, so I'd rather have wipe as a separate command. After all, it does not really build anything.
> 
Hmm good point:)
What do you think about "osc chroot --wipe"? (I'm just a bit hesitant to
add yet another command to the osc ui)

dwmh:
Yeah, chroot --wipe would make more sense. I'll rewrite this a bit.

dwmh:
It's now 'osc chroot --wipe'

marcus-h:
On 2016-09-28 05:51:07 -0700, Matias Hilden wrote:
> It's now 'osc chroot --wipe'
> 
Thanks! LGTM.
Minor remarks:
- use the print function instead of sys.stdout.write
- use whitespaces around the operators (makes it easier to read;) )

Anyway... these are no showstoppers:)
I'll merge it once the corresponding obs-build PR got merged.

Thanks again for the patch!

marcus-h:
While merging I noticed that the commit message should be updated
as well (something like 'Add "--wipe" option to "chroot" command'...).
Sorry for requesting yet another change... :/ but I'll merge it
afterwards!:)

Ps. if you get this message twice feel free to ignore one of them (not sure
if my mail reply will eventually appear...)

marcus-h:
On 2016-09-30 11:17:59 +0200, Marcus Hüwe wrote:
> On 2016-09-28 05:51:07 -0700, Matias Hilden wrote:
> > It's now 'osc chroot --wipe'
> > 
> Thanks! LGTM.
> Minor remarks:
> - use the print function instead of sys.stdout.write
> - use whitespaces around the operators (makes it easier to read;) )
> 
> Anyway... these are no showstoppers:)
> I'll merge it once the corresponding obs-build PR got merged.
> 
While merging I noticed that the commit message should be updated
as well (something like 'Add "--wipe" option to "chroot" command'...).
Sorry for requesting yet another change... :/ but I'll merge it
afterwards!:)

tuminoid:
Build counterpart was merged, please test and merge this too.

marcus-h:
On 2016-10-27 09:13:36 -0700, Tuomo Tanskanen wrote:
> Build counterpart was merged, please test and merge this too.
> 
Merged. Thanks!

PR #235, opened 2016-10-01
[UNTESTED] pass apiurl to source services (#234)

### LINKED ISSUE #234: FR: osc should pass apiurl to source services
Some source services need to retrieve values from `~/.oscrc`; for example `tar_scm` needs to know the user's email so it can include it in the generated changes which are automatically prepended to the changelog.  In order to do this lookup correctly, it needs to know the apiurl of the BS instance currently in use, in order to use the right section of `~/.oscrc`.

A real world example demonstrating the need for this feature request is described in https://github.com/openSUSE/obs-service-tar_scm/pull/119.

aspiers:
[**NOTE**: I haven't tested this yet, or at least I don't *think* I've tested it, but then [I can't even remember coding it](https://github.com/openSUSE/osc/issues/234#issuecomment-250940279) so who knows ... ;-)  It looks reasonable to me though.]

Some source services need to know the apiurl, e.g. to lookup values in the `~/.oscrc` config file.

Closes #234.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -297,6 +297,7 @@ class Serviceinfo:
             self.services.append(data)
 
     def getProjectGlobalServices(self, apiurl, project, package):
+        self.apiurl = apiurl
         # get all project wide services in one file, we don't store it yet
         u = makeurl(apiurl, ['source', project, package], query='cmd=getprojectservices')
         try:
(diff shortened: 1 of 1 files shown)


marcus-h:
LGTM. Merged. Thanks!

Issue #84, opened 2014-04-20
Avoid duplicates due to a second OBS instance


jobermayr:
"osc build" should know that PROJECT on api.opensuse.org and openSUSE.org:PROJECT on pmbs-api.links2linux.de (https://pmbs.links2linux.de/project/show/openSUSE.org) are the same ...

Btw. if packages from openSUSE.org:PROJECT get fetched via api.opensuse.org->pmbs-api.opensuse.org->MYHOST it should also go directly (api.opensuse.org->MYHOST) to avoid unnecessary traffic on second OBS instance ...

dmach:
The linked PR was rejected and it seems that this cannot be easily implemented without hardcoding values in osc, which wouldn't work in a general case. Closing.

PR #237, opened 2016-10-03
Fetch upstream packages on 3rd party OBS directly from openSUSE api.

jobermayr:
Fixes:
https://pmbs-api.links2linux.de/build/openSUSE.org:openSUSE:Factory/snapshot/x86_64/_repository?...&view=cpio: HTTP Error 500: Internal Server Error

Also a start for: https://github.com/openSUSE/osc/issues/84

### DIFF
--- a/osc/fetch.py
+++ b/osc/fetch.py
@@ -110,7 +110,11 @@ class Fetcher:
         query = ['binary=%s' % quote_plus(i) for i in pkgs]
         query.append('view=cpio')
         try:
-            url = makeurl(apiurl, ['build', project, repo, arch, package], query=query)
+            if project.startswith('openSUSE.org'):
+                sys.stdout.write("Found upstream project ...\r")
+                url = makeurl('https://api.opensuse.org', ['build', project[13:], repo, arch, package], query=query)
+            else:
+                url = makeurl(apiurl, ['build', project, repo, arch, package], query=query)
             sys.stdout.write("preparing download ...\r")
             sys.stdout.flush()
             with tempfile.NamedTemporaryFile(prefix='osc_build_cpio') as tmparchive:


adrianschroeter:
Please don't hardcode project names and remoteurls. I see the value, but it should be reported by the server directly from where to download.

Also, the 500 crash should be fixed in first place, so you would not need this workaround hack.

marcus-h:
On 2016-10-03 23:36:51 -0700, Adrian Schröter wrote:
> Also, the 500 crash should be fixed in first place, so you would not need this workaround hack.
> 
Already "fixed" in commit bdec2046003419f3bc4c2f31f43313ef976adade.

jobermayr:
> Please don't hardcode project names and remoteurls.

I see hardcoding is not allowed within osc tool: https://github.com/openSUSE/osc/blob/master/osc/conf.py#L98

Also [Admin Guide ](http://openbuildservice.org/help/manuals/obs-admin-guide/obs.cha.installation_and_configuration.html#_online_configuration) says:
> If you want to use an interconnect to another OBS instance to reuse the build targets you can do this as Admin [...]
`<project name="openSUSE.org">`
`[...]`
`<remoteurl>https://api.opensuse.org/public</remoteurl>`

So this PR simply implements what is preferred ...

marcus-h:
On 2016-10-04 10:05:24 -0700, Johannes Obermayr wrote:
> > Please don't hardcode project names and remoteurls.
> 
> I see hardcoding is not allowed within osc tool: https://github.com/openSUSE/osc/blob/master/osc/conf.py#L98
> 
That's just a default that can be overriden.

> Also [Admin Guide ](http://openbuildservice.org/help/manuals/obs-admin-guide/obs.cha.installation_and_configuration.html#_online_configuration) says:
> > If you want to use an interconnect to another OBS instance to reuse the build targets you can do this as Admin [...]
> `<project name="openSUSE.org">`
> `[...]`
> `<remoteurl>https://api.opensuse.org/public</remoteurl>`
> 
> So this PR simply implements what is preferred ...
> 
The problem is that it does not work in the general case, because
one could set up an "openSUSE.org" project that is configured with
a different remoteurl (!= api.o.o) or has no remoteurl element.
Additionally, this change implies that every PMBS user needs an
api.o.o account as well.

PR #238, opened 2016-10-04
Osc plugin for project level builds

dwmh:
Note that this is a plugin for osc, so you'll probably want to create a separate repository for this if you want to take it in.


marcus-h:
On 2016-10-04 07:35:55 -0700, Matias Hilden wrote:
> Note that this is actually a plugin for osc, so you'll probably want to create a separate repository for this if you want to take it in.

What's your concrete use case(s) for this feature?

I just had a quick glance at your patch, but I think the following
scenario won't work (this no criticism - just to check if I got the
code right:) ):

###
Assume we have the following packages:

On the server:
Pkg1:
Pkg2: Requires: Pkg3
Pkg3:

No package on the server provides "foo".

Next, modify Pkg1 and Pkg3 locally (add a provides and a requires).

Locally:
Pkg1: Provides: foo
Pkg2: Requires: Pkg3
Pkg3: Requires: foo

Let's assume we call "osc prjmake" and "os.listdir" returns the
packages in the following order: Pkg1, Pkg2, Pkg3

Pkg1 and Pkg2 are resolvable, but Pkg3 isn't (maybe the code in
buildorder.py should check for an "error" element). In this situation,
calc_buildorder will compute the following order:

Pkg1, Pkg3, and then Pkg2

The funny thing is that this yields the expected results (even though
Pkg3 was initially unresolvable).
However, if we assume that "os.listdir" returns the packages in
the order Pkg2, Pkg3, Pkg1, then it'll compute the following order:

Pkg3, Pkg2, and then Pkg1

In this case the "prjmake" will fail, because the build of "Pkg3"
fails (since it is unresolvable).
###


Is this a use case/scenario that you want to support?

dwmh:
This plugin kind of provides an alternative for having a home project where you verify your changes don't cause dependent breakages. The advantage is that you can use your local machine to build the project without having to wait for workers.

The assumption is that the project you are working on already exists in a reasonably stable state on the server side.

The plugin does the dependency resolution based on bdep and subpack entries in buildinfos, so that already sets some limitations for its use. A package being unresolvable on server side is an unhandled error that causes prjmake to behave incorrectly, I just overlooked it since that does not happen in our use cases. So no, the use case you described is not supported and should result in a failure.

It's also possible produce incorrect build orders by locally changing Provides for multiple packages, since the plugin relies on the server to solve those dependencies.

I will document these restrictions and add some kind of error check for the unresolvable case.

dwmh:
Unresolvable packages are still allowed if they are fixed by your local changes. Using a combination of Provides and Requires not matching package names can still cause issues.

marcus-h:
On 2016-10-07 02:33:20 -0700, Matias Hilden wrote:
> Unresolvable packages are still allowed if they are fixed by your local changes.
> 
Hmm

Pkg1: BuildRequires: PkgLocallyAdded
PkgLocallyAdded:

Assume os.listdir returns Pkg1, PkgLocallyAdded
then the current code should break (the calculated order is
Pkg1, PkgLocallyAdded as well).

In order to support the special case of unresolvable "package names",
you would probably need several iterations:

- build resolvable packages
- calculate new order/buildinfos while taking the previously built
  packages into account
- go to the first step again

Ps. s/Requires/BuildRequires/ in my previous post:)

dwmh:
True, the code would have to do what the scheduler does, so the buildorder would have to be recalculated after each build with the produced build results. For the above example, the code would produce an error since the server does not know about the local package (and preferred packages are not passed to the buildinfo query).

However, the current naive way of calculating the buildorder works most of time, at least in our case.

adrianschroeter:
We have meanwhile the pbuild tool for handling multipackage builds. It is independend of OBS, but it can be used to configured to use OBS repositories. We see this as a superior solution here.

http://opensuse.github.io/obs-build/pbuild.html

PR #239, opened 2016-10-06
improved watched output of osc results

lethliel:
before joining the results on a osc r -w a newline is printed now
IMHO this makes the output more easy to read.


marcus-h:
On 2016-10-07 01:08:39 -0700, Marco Strigl wrote:
> I fixed the indents an rebased
> 
Merged. Thanks!

PR #242, opened 2016-11-14
Support running osc non-interactively (in e.g. scripts).

dwmh:
Provide a global option for setting non-interactive mode and disable
prompts when creating a new config file if the new option was given by the
user.

Co-Authored-By: Juha Kallioinen <juha.kallioinen@ericsson.com>

### DIFF
--- a/osc-wrapper.py
+++ b/osc-wrapper.py
@@ -35,6 +35,7 @@ except NameError:
 # remains fully buffered (see PyFile_SetBufSize (Objects/fileobject.c))).
 if not os.isatty(sys.stdout.fileno()):
     sys.stdout = os.fdopen(sys.stdout.fileno(), sys.stdout.mode, 1)
+    sys.argv.insert(1, "--non-interactive")
 
 osccli = commandline.Osc()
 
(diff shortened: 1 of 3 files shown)


marcus-h:
Hmm IMHO "--non-interactive" is a bit misleading here: I would have expected that osc
works in some "non-interactive" mode for _all_ commands (using some "defaults",
where it makes sense). Can't you simply check these special conditions within your script?
(Or if you just want to abort if osc tries to read something from stdin you could simply close stdin
when starting osc (for instance, in bash you could simply do "exec 0<&- osc..."))

trhd:
The idea is that the global --non-interactive option would indeed make all osc subcommands non-interactive. Its non-trivial however to identify all the situations that might end up asking some kind of input from the user and I would thus start by adding the option and then incrementally add checks for it where needed.

It would probably be be possible to work around this from the calling script as well but as we have had more than one use case for this kind of functionality it was seen best to add the functionality into osc itself.

trhd:
Feel free to close this PR if this is something that doesn't interest you.

dmach:
This PR has been inactive for a long time, closing.
@marcus-h @lethliel please reopen it if there's still interest in getting this change in.

PR #243, opened 2016-11-14
Ignore build descr directory in build-root substitution for local packages

dwmh:
If build-root contains %(package) substitution, --local-package builds
would substitute absolute path there. This is different than the rule used
in osc chroot (uses relative path), causing the chroot to fail by default.

This commit removes the directory part from both build-root substitutions.


marcus-h:
Merged. Thanks!

PR #245, opened 2016-11-15
Backport a fix from Python 2.7 upstream for urllib2.

trhd:
This change makes basic authentication fail after first unsuccessful
attempt (instead of automatically retrying) and thus prevents unnecessary
failed login attempts.

The problem can at least be reproduced on Ubuntu 14.04 with Python 2.7.6.

This change also allows printing the contents of an HTTP 401 message to
the user.

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -42,6 +42,7 @@ import os
 import re
 import sys
 import ssl
+import warnings
 
 try:
     from http.cookiejar import LWPCookieJar, CookieJar
@@ -57,7 +58,7 @@ except ImportError:
     from httplib import HTTPConnection, HTTPResponse
     from StringIO import StringIO
     from urlparse import urlsplit
-    from urllib2 import URLError, HTTPBasicAuthHandler, HTTPCookieProcessor, HTTPPasswordMgrWithDefaultRealm, ProxyHandler
+    from urllib2 import URLError, HTTPBasicAuthHandler, HTTPCookieProcessor, HTTPPasswordMgrWithDefaultRealm, ProxyHandler, AbstractBasicAuthHandler
     from urllib2 import AbstractHTTPHandler, build_opener, proxy_bypass, HTTPSHandler
 
 from . import OscConfigParser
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2016-11-15 06:02:31 -0800, Hemmo Nieminen wrote:
> This change makes basic authentication fail after first unsuccessful
> attempt (instead of automatically retrying) and thus prevents unnecessary
> failed login attempts.
> 
From a very quick glance the patch looks ok to me in general, but the
question is if we really need it: the present OscHTTPBasicAuthHandler
classes are just workarounds for _broken_ urllib2 versions. Luckily,
"recent" python releases don't need such fixes anymore. So, I would
rather like to remove this "cruft" entirely from the conf module...

Your patch just fixes a "cosmetic" (IMHO) issue: reducing the number
of auth retries from 5 (iirc?) to 1.

Do you have a specific use-case for this change?

> The problem can at least be reproduced on Ubuntu 14.04 with Python 2.7.6.
> 
> This change also allows printing the contents of an HTTP 401 message to
> the user.

Hmm what do you exactly mean? In case of an unsuccessful auth a HTTPError
with code 401 should be raised (even with an "ancient" python version (iirc)).

trhd:
On 09:34 Wed 16 Nov, Marcus Hüwe wrote:
> On 2016-11-15 06:02:31 -0800, Hemmo Nieminen wrote:
> > This change makes basic authentication fail after first unsuccessful
> > attempt (instead of automatically retrying) and thus prevents unnecessary
> > failed login attempts.
> > 
> From a very quick glance the patch looks ok to me in general, but the
> question is if we really need it: the present OscHTTPBasicAuthHandler
> classes are just workarounds for _broken_ urllib2 versions. Luckily,
> "recent" python releases don't need such fixes anymore. So, I would
> rather like to remove this "cruft" entirely from the conf module...

I understand, but in my experience enterprise world does seem to lag
behind _a bit_ in these kinds of things.

> Your patch just fixes a "cosmetic" (IMHO) issue: reducing the number
> of auth retries from 5 (iirc?) to 1.
>
> Do you have a specific use-case for this change?

Yes. As a single osc invocation causes 6 failed login attempts an
unsuspecting user can get his account very easily locked
(authentication being done against an LDAP server) causing major
inconveniences.

> > The problem can at least be reproduced on Ubuntu 14.04 with Python 2.7.6.

I'm using this setup to try this now as it has been a while since I
wrote the actual patch.

> > This change also allows printing the contents of an HTTP 401 message to
> > the user.
> 
> Hmm what do you exactly mean? In case of an unsuccessful auth a HTTPError
> with code 401 should be raised (even with an "ancient" python version (iirc)).

This change allows displaying the custom status/error messages the OBS
server sends to the client (as some osc commands do).

E.g.

# with invalid username/password defined in ~/.oscrc
$ ~/osc/osc-wrapper.py -d -ci -m dslfk

(without the patch):
[snip]
WWW-Authenticate: basic realm="API login"
X-Opensuse-Errorcode: authentication_required
X-Opensuse-Runtimes: {"view":0.905635,"db":0.358123,"backend":0,"xml":0}
Vary: Accept-Encoding
Cache-Control: no-cache
X-Request-Id: c129674c-3bdf-43bf-a03b-954e18603d9a
X-Runtime: 0.141938
X-Powered-By: Phusion Passenger 4.0.33
Status: 401 Unauthorized
Cache-Control: public
Connection: close
Transfer-Encoding: chunked
Content-Type: application/xml; charset=utf-8

(with the patch):
[snip]
WWW-Authenticate: basic realm="API login"
X-Opensuse-Errorcode: authentication_required
X-Opensuse-Runtimes: {"view":0.89353,"db":0.33147,"backend":0,"xml":0}
Vary: Accept-Encoding
Cache-Control: no-cache
X-Request-Id: 4db6d190-48ce-489c-a6ed-5ad75014142c
X-Runtime: 0.191473
X-Powered-By: Phusion Passenger 4.0.33
Status: 401 Unauthorized
Cache-Control: public
Connection: close
Transfer-Encoding: chunked
Content-Type: application/xml; charset=utf-8

<status code="authentication_required">
  <summary>Unknown user 'lksjdflfkdsj' or invalid password</summary>
</status>

marcus-h:
On 2016-11-17 01:02:17 -0800, Hemmo Nieminen wrote:
> On 09:34 Wed 16 Nov, Marcus Hüwe wrote:
> > On 2016-11-15 06:02:31 -0800, Hemmo Nieminen wrote:
> > > This change makes basic authentication fail after first unsuccessful
> > > attempt (instead of automatically retrying) and thus prevents unnecessary
> > > failed login attempts.
> > > 
> > From a very quick glance the patch looks ok to me in general, but the
> > question is if we really need it: the present OscHTTPBasicAuthHandler
> > classes are just workarounds for _broken_ urllib2 versions. Luckily,
> > "recent" python releases don't need such fixes anymore. So, I would
> > rather like to remove this "cruft" entirely from the conf module...
> 
> I understand, but in my experience enterprise world does seem to lag
> behind _a bit_ in these kinds of things.
> 
Ok.

> > Your patch just fixes a "cosmetic" (IMHO) issue: reducing the number
> > of auth retries from 5 (iirc?) to 1.
> >
> > Do you have a specific use-case for this change?
> 
> Yes. As a single osc invocation causes 6 failed login attempts an
> unsuspecting user can get his account very easily locked
> (authentication being done against an LDAP server) causing major
> inconveniences.
> 
I agree:)

Could you add an "import warnings" to the patch? Otherwise,
```
+                            warnings.warn("Basic Auth Realm was unquoted",
+                                          UserWarning, 2)
```
won't work. The rest looks ok to me. Thanks in advance!

marcus-h:
Merged. Thanks a lot for this patch!

Ps., if you update a PR (forced push), please add a comment
to the PR that you updated it (it seems that github sends
no notification email in this case...).

marcus-h:
Just a test (please ignore this, if you receive this message)

PR #247, opened 2016-11-16
Alter osc update operation's default behaviour.

trhd:
Make update operation behave as checkout does and do not checkout linked
packages from the same project (by default) during a project wide update.

Default behaviour can be altered with the link expansion options.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -4720,7 +4720,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                     # (a) update all packages
                     args += prj.pacs_have
                     # (b) fetch new packages
-                    prj.checkout_missing_pacs(expand_link = not opts.unexpand_link)
+                    prj.checkout_missing_pacs(opts.expand_link, opts.unexpand_link)
                     args.remove(arg)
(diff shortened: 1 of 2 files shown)


marcus-h:
Makes sense:) Merged.
Thanks a lot for your patch!

Btw., I tweaked your changes a bit in commit
25ccfde43c8a1fd958d5179ff0e38a136cad14d9 to avoid the http request in
Project.checkout_missing_pacs.

Btw2., maybe we should also checkout a package in case of a linkerror
(that's what "osc co..." does by default). What do you think?

Thanks again!

trhd:
> Btw., I tweaked your changes a bit in commit
> 25ccfde to avoid the http request in
> Project.checkout_missing_pacs.

Looks good to me.

> Btw2., maybe we should also checkout a package in case of a linkerror
> (that's what "osc co..." does by default). What do you think?

I'd say that the state of the links should probably not affect osc's behaviour w.r.t. expanding/skipping the links. Perhaps checkout's behaviour should be altered instead so that it too will skip the erroneous link (unless explicitly told not to)?

PR #252, opened 2016-12-08
Add osc password command

trhd:


marcus-h:
This can be already done via the "config" command:
"osc config apiurl/alias pass <newval>" or
"osc config apiurl/alias pass --no-echo"

trhd:
Yes but as changing one's password is such a common thing to do we deemed this feature would be useful. Especially for those new to using osc.

marcus-h:
On 2016-12-08 23:39:45 -0800, Hemmo Nieminen wrote:
> Yes but as changing one's password is such a common thing to do we deemed this feature would be useful. Especially for those new to using osc.
> 
I agree, but I don't think it deserves an own command.
@adrianschroeter: any opinion on this?

Btw., if your users need this, you could also ship it as an osc
plugin.

PR #253, opened 2016-12-08
Fetch ssl cert if the spec file asks for it

lnussel:
Allows building packages like e.g. ovmf locally the same as on the
server

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -636,6 +636,18 @@ def main(apiurl, opts, argv):
     prefer_pkgs = {}
     build_descr_data = open(build_descr).read()
 
+    if re.search(r'^#\s*needssslcertforbuild\s*$', build_descr_data, re.MULTILINE):
+        try:
+            url = makeurl(apiurl, ['source', prj, '_project', '_sslcert'], 'meta=1')
+            cert = ''.join(osc.core.http_GET(url).readlines())
+            with open(os.path.dirname(build_descr)+'/_projectcert.crt', 'w') as fh:
+                fh.write(cert)
+        except HTTPError as e:
+            if e.code == 404:
+                print('WARNING: spec file requests ssl cert but %s does not have one.'%prj)
+                print('Build may fail, use osc signkey to create a key')
+                pass
+
     # XXX: dirty hack but there's no api to provide custom defines
     if opts.without:
         s = ''


marcus-h:
```
+            url = makeurl(apiurl, ['source', prj, '_project', '_sslcert'], 'meta=1')
+            cert = ''.join(osc.core.http_GET(url).readlines())
```
Hmm does work for you? If I understand it correctly, the GET should always
result in a 404, because the backend doesn't expose the _sslcert file (see,
for instance, BSRevision::lsrev).

marcus-h:
On 2016-12-08 19:57:13 +0100, Marcus Hüwe wrote:
> Hmm does work for you? If I understand it correctly, the GET should always
> result in a 404, because the backend doesn't expose the _sslcert file (see,
> for instance, BSRevision::lsrev).

(via this route and the /getsslcert route is not accessible from the
outside)

marcus-h:
On 2016-12-08 19:57:13 +0100, Marcus Hüwe wrote:
> ```
> +            url = makeurl(apiurl, ['source', prj, '_project', '_sslcert'], 'meta=1')
> +            cert = ''.join(osc.core.http_GET(url).readlines())
> ```
> Hmm does work for you? If I understand it correctly, the GET should always
> result in a 404, because the backend doesn't expose the _sslcert file (see,
> for instance, BSRevision::lsrev).

Arghs... this is complete nonsense... I mixed up the _sslcert file with the
_signkey file - sorry. However, it seems that most projects don't have a
_sslcert file (for instance, openSUSE:Factory has no _sslcert file). The
_sslcert file is only created/updated (during createkey/extendkey) if the
project already has a _sslcert file.

lnussel:
Oh it does have one. If it's not exposed via api it's a server setup error. Factory and distros are specially treated on server side in obs, the api doesn't know about it.

marcus-h:
On 2016-12-12 02:39:26 -0800, Ludwig Nussel wrote:
> Oh it does have one. If it's not exposed via api it's a server setup error. Factory and distros are specially treated on server side in obs, the api doesn't know about it.
> 
That's probably the backend code path with length(<signkey>) <= 2. Is
there a reason why the e.g. factory _sslcert is (explicitly) not committed
to the _project's meta?

lnussel:
mls fears it gets out of sync

marcus-h:
On 2016-12-12 07:50:12 -0800, Ludwig Nussel wrote:
> mls fears it gets out of sync
> 
Hmm ok. Do you still want to have this PR merged?

lnussel:
Yes, at least it works for rings and staging

marcus-h:
Is this urgent? Otherwise, I'll wait for
https://github.com/openSUSE/open-build-service/issues/2472.

lnussel:
not urgent. If we have the new api it's perfectly fine to reject this and implement against the new api instead

lethliel:
openSUSE/open-build-service#2472 has been closed. This PR is now 3 years old.

@lnussel: can this be closed now?

lnussel:
I guess it needs to be reimplemented using the api but yes

lethliel:
Ok. Then this will be closed. Thanks for the info.

PR #254, opened 2016-12-15
Add support for Kerberos based authentication

trhd:
The osc part of the Kerberos support PR in open-build-service: https://github.com/openSUSE/open-build-service/pull/2467

Note the potential license issue with urllib2_kerberos. This could perhaps be avoided by removing it from the osc's source tree and relying the environment to provide it?

### DIFF
--- a/osc-wrapper.py
+++ b/osc-wrapper.py
@@ -38,5 +38,6 @@ if not os.isatty(sys.stdout.fileno()):
 
 osccli = commandline.Osc()
 
+babysitter.run_baby_sitter = True
 r = babysitter.run(osccli)
 sys.exit(r)
--- a/osc/babysitter.py
+++ b/osc/babysitter.py
@@ -42,6 +42,8 @@ except ImportError:
 
 # the good things are stolen from Matt Mackall's mercurial
 
+# this is set True from osc-wrapper.py
+run_baby_sitter = False
 
 def catchterm(*args):
     raise oscerr.SignalInterrupt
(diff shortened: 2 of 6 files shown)


trhd:
A Travis misconfiguration?

trhd:
I don't really expect this to be merged as it is...don't really know Travis well enough to understand what the problem with the urlgrabber module is.

There is also the possible license issue with the urllib2_kerberos.

marcus-h:
Sorry for the late response... I had no time so far to have a closer
look at this. Currently, it is quite "cumbersome" to add additional
authentication methods to osc. I'm planning to make this a bit more
easier (needs some refactorings, cleanups etc.). One idea is to
use python distribute's/setuptools' entry point mechanism. Then,
osc could quite easily support 3rd party/additional auth methods
(see for instance [1]).

Example:
- your osc-auth-kerberos plugin registers at some "osc_config_auth"
  entry point
- instead of asking for a username/password, osc prompts for the
  preferred auth method (osc simply reads the "osc_config_auth"
  entry point)
- next, the user supplies the data needed by the selected auth
  method

This way, we could easily add new auth methods for whatever the
api supports/could support in the future (api token based auth,
oauth etc.).

Do you think something like would be useful?


[1] https://github.com/openSUSE/osc2/blob/master/osc2/cli/plugin.py

trhd:
Sure, why not. Should surely make it easier to implement new authentication methods. A bit afraid though that you might end up writing plugin specific entry points for the hooks as different authentication methods may end up requiring very different kinds of changes to the osc.

PR #257, opened 2017-01-17
RFC: show build results of multibuild packages by default

adrianschroeter:
Be in sync with webui. Also it is assumed that a user wants to see
all build results of the source change he did and not only for a subset.

This turns around the default behaviour, so please comment on this :)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,6 @@
 0.157
   - add unpublish command (requires OBS 2.8)
+  - results: show multibuild results by default
 
 0.156
   - highlight scheduled jobs with dispatch problems (due to constraints)
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -4977,8 +4977,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                         help='Show results only for specified architecture(s)')
     @cmdln.option('-v', '--verbose', action='store_true', default=False,
                         help='more verbose output')
-    @cmdln.option('-m', '--multibuild', action='store_true', default=False,
-                        help='Show results for all packages in multibuild')
+    @cmdln.option('--no-multibuild', action='store_true', default=False,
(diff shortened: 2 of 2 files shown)


marcus-h:
On 2017-01-17 07:47:53 -0800, Adrian Schröter wrote:
> Be in sync with webui. Also it is assumed that a user wants to see
> all build results of the source change he did and not only for a subset.
> 
> This turns around the default behaviour, so please comment on this :)

In general, I've no objection against it (especially since it makes things
more consistent), but it breaks the ui:/
But... for the sake of consistency, let's do it (better now than later...)

marcus-h:
Btw.,
+    @cmdln.option('-m', '--no-multibuild', action='store_true', default=False,
+                        help='Disable results for all direct affect packages inside of the project')

Do we want to keep the short option? (The help text sounds a bit strange)

adrianschroeter:
right, we should remove that one then

adrianschroeter:
another option is to enable it only for multibuild by default, but not for locallink. So it would not be incompatible (to the last released osc via maintenance), because it would affect only packages using the new _multibuild feature.

Downside is that the user would need to understand the difference of local linked packages and multibuild ....

adrianschroeter:
@mlschroe @lethliel any opinion about this?

lethliel:
I have nothing against changing the default behavior. 

@marcus-h But why would it break the ui?

marcus-h:
On 2017-01-17 08:23:59 -0800, Marco Strigl wrote:
> @marcus-h But why would it break the ui? 
> 
The option, which is part of the 0.156.0 release, is renamed. In an
upcoming release, the absence of this option might break, for instance,
existing scripts.
But since the multibuild feature is rather new, I don't expect any
real-world problems:)

lethliel:
LGTM

PR #258, opened 2017-01-17
Fix URLGrabError exception handling.

OlegGirko:
`URLGrabError` class has no `exception` attribute, use `strerror` to print error instead.


marcus-h:
Merged. Thanks for the patch! (Btw., the URLGrabError instance might
have an exception attribute (in some cases it is added it on-the-fly
in the "fetch" module).)

PR #260, opened 2017-01-26
[RFC] improved build recipe guessing

lethliel:
A try to improve the build recipe guessing.

**Prerequisites:**

- there need to be multiple build recipes (e.g. pack1.spec pack1.dsc PKGBUILD)
- when calling `obs build` a repository and an arch needs to be given
- it doesn't work with multiple build recipes of the same type

**How the build recipe is determined:** 

1. get the build configuration (based on repository)
2. decide, based on repotype or patterntype, which recipe to use
3. Look for a recipe file in directory for this type

Please share your thoughts about this approach and if this is a thing we want.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -5742,6 +5742,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         """helper to parse the repo, arch and build description from args"""
         import osc.build
         import glob
+        import tempfile
         arg_arch = arg_repository = arg_descr = None
         if len(args) < 3:
             # some magic, works only sometimes, but people seem to like it :/
@@ -5786,15 +5787,18 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                 # only persist our own repos
                 Repo.tofile(repolistfile, repositories)
 
+        no_repo = False
         repo_names = sorted(set([r.name for r in repositories]))
         if not arg_repository and repositories:
             # XXX: we should avoid hardcoding repository names
             # Use a default value from config, but just even if it's available
             # unless try standard, or openSUSE_Factory, or openSUSE_Tumbleweed
+            no_repo = True
             arg_repository = repositories[-1].name
             for repository in (conf.config['build_repository'], 'standard', 'openSUSE_Factory', 'openSUSE_Tumbleweed'):
                 if repository in repo_names:
                     arg_repository = repository
+                    no_repo = False
                     break
 
         if not arg_repository:
@@ -5818,22 +5822,37 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         elif not arg_descr:
             msg = None
             if len(descr) > 1:
-                # guess/prefer build descrs like the following:
-                # <pac>-<repo>.<ext> > <pac>.<ext>
-                # no guessing for arch's PKGBUILD files (the backend does not do any guessing, too)
+                if no_repo:
+                    raise oscerr.WrongArgs("Repository is missing. Cannot guess build description without repository")
+                apiurl = self.get_api_url()
+                project = store_read_project('.')
+                bc = get_buildconfig(apiurl, project, arg_repository)
+                with tempfile.NamedTemporaryFile() as f:
+                    f.write(bc)
+                    f.flush()
+                    recipe = return_external('/usr/lib/build/queryconfig', '--dist', f.name, 'type')
+                recipe = recipe.strip()
+                if recipe == 'arch':
+                    recipe = 'PKGBUILD'
                 pac = os.path.basename(os.getcwd())
                 if is_package_dir(os.getcwd()):
                     pac = store_read_package(os.getcwd())
-                extensions = ['spec', 'dsc', 'kiwi', 'livebuild']
-                cands = [i for i in descr for ext in extensions if i == '%s-%s.%s' % (pac, arg_repository, ext)]
+                if recipe == 'PKGBUILD':
+                    cands = [d for d in descr if d.startswith(recipe)]
+                else:
+                    cands = [d for d in descr if d.endswith('.' + recipe)]
+                if len(cands) > 1:
+                    repo_cands = [d for d in cands if d == '%s-%s.%s' % (pac, arg_repository, recipe)]
+                    if repo_cands:
+                        cands = repo_cands
+                    else:
+                        pac_cands = [d for d in cands if d == '%s.%s' % (pac, recipe)]
+                        if pac_cands:
+                            cands = pac_cands
                 if len(cands) == 1:
                     arg_descr = cands[0]
-                else:
-                    cands = [i for i in descr for ext in extensions if i == '%s.%s' % (pac, ext)]
-                    if len(cands) == 1:
-                        arg_descr = cands[0]
                 if not arg_descr:
-                    msg = 'Multiple build description files found: %s' % ', '.join(descr)
+                    msg = 'Multiple build description files found: %s' % ', '.join(cands)
             elif not ignore_descr:
                 msg = 'Missing argument: build description (spec, dsc, kiwi or livebuild file)'
                 try:
--- a/osc/core.py
+++ b/osc/core.py
@@ -7310,6 +7310,32 @@ def run_external(filename, *args, **kwargs):
             raise
         raise oscerr.ExtRuntimeError(e.strerror, filename)
 
+def return_external(filename, *args, **kwargs):
+    """Executes the program filename via subprocess.check_output.
+
+    *args are additional arguments which are passed to the
+    program filename. **kwargs specify additional arguments for
+    the subprocess.check_output function.
+    if no args are specified the plain filename is passed
+    to subprocess.check_output (this can be used to execute a shell
+    command). Otherwise [filename] + list(args) is passed
+    to the subprocess.check_output function.
+
+    Returns the output of the command.
+
+    """
+    if args:
+        cmd = [filename] + list(args)
+    else:
+        cmd = filename
+
+    try:
+        return subprocess.check_output(cmd, **kwargs)
+    except OSError as e:
+        if e.errno != errno.ENOENT:
+            raise
+        raise oscerr.ExtRuntimeError(e.strerror, filename)
+
 # backward compatibility: local role filtering
 def filter_role(meta, user, role):
     """


marcus-h:
On 2017-01-26 02:37:45 -0800, Marco Strigl wrote:
> **How the build recipe is determined:** 
> 
> 1. get the build configuration (based on repository)
> 2. decide, based on repotype or patterntype, which recipe to use

Hmm we should use "Type:" for this (in case of a missing type, we
could use the same fallback as in Build::read_config). Regardless
of which *type we use, we would actually need to parse the buildconfig
file (that is we have to do the macro expansion etc.). Alternatively,
we could simply use the first/last "Type:" that is present (or maybe
first/last unconditional "Type:") (I think that's ok for a start,
because parse_repoarchdescr is just a heuristic).
(Maybe we could move the code for this into a new function or
encapsulate it in a new Buildconfig class).

> Please share your thoughts about this approach and if this is a thing we want.

Makes sense
I just had a quick glance at the code - see comments below.

> @@ -5786,11 +5787,13 @@ Please submit there instead, or use --nodevelproject to force direct submission.
>                  # only persist our own repos
>                  Repo.tofile(repolistfile, repositories)
>  
> +        no_repo = 0

Using "False" would be more pythonic (and later no_repo = True).

> +                if no_repo == 1:
> +                    raise oscerr.WrongArgs("Repository is missing. Cannot gues recipe without repository")
"guess":) and maybe "build description"? (see below)

> +                    if (repotype ==  'suse'):
> +                        recipe='spec'

Cosmetic remark: no parentheses and whitespaces around "=":)

> @@ -5829,7 +5858,14 @@ Please submit there instead, or use --nodevelproject to force direct submission.
>                  if len(cands) == 1:
>                      arg_descr = cands[0]
>                  else:
> -                    cands = [i for i in descr for ext in extensions if i == '%s.%s' % (pac, ext)]
> +                    if ('recipe' in locals()):

Arghs:) just do something like "recipe = None" or "recipe = ''" (in
the correct scope) and then "if recipe is not None: ..." etc.

> +                        if (recipe == 'spec' or recipe == 'arch'):
> +                            recipe_r = re.compile(r"\.spec")

Hmm what's spec/arch specific here and why is "\.spec" hardcoded?

> +                            if len(recipe_r.findall(" ".join(descr))) > 1:

Maybe it is "easier" to use a list comprehension + endswith instead of a
regex: if len([d for d in descr if d.endswith('.' + recipe)]) > 1
(just for the sake of contradiction, you could have a file called "foo.spec"
and a second file called "foo.spec.nobody.would.do.this.dsc" - in this case
the regex approach would "fail") (note: we would also need an extra check
for PKGBUILD).

> +                                raise oscerr.WrongArgs("More than one recipe file found for repository type. Please specify recipe")

Probably we should use the term "build description" instead of "recipe"
(at least that's how it is currently called in osc).

> +                        cands = [i for i in descr if i == '%s.%s' % (pac, recipe) or i == recipe]

Hmm can't we simply take the descr for which .endswith(descr) holds? (If we
extend the previous check to all recipe types, then there's exactly one
file for which this holds).

> +                    else:
> +                        cands = [i for i in descr for ext in extensions if i == '%s.%s' % (pac, ext)]

lethliel:
I now use queryconfig to get the build type. 
Switched from Regex to a list comprehension

marcus-h:
On 2017-01-31 00:01:43 -0800, Marco Strigl wrote:
> I now use queryconfig to get the build type. 

Great! The patch looks really good - I have just some minor
remarks:)


> +                if no_repo == True:

style: "if no_repo: ..."

> +                    raise oscerr.WrongArgs("Repository is missing. Cannot guess build desctiption without repository")

"description"

<SNIP>

> +                store_write_string(os.curdir, '_build.conf', bc +'\n')

Hmm let's not "pollute" the storedir. I think the best is to use
a tempfile. E.g., something along these lines

with tempfile.NamedTemporaryFile() as f:
    f.write(bc)
    ...
    recipe = return_external(*cmd)
    ...

The advantage is that the tempfile is automatically unlinked (even
if some exception is raised).

> @@ -5829,7 +5843,13 @@ Please submit there instead, or use --nodevelproject to force direct submission.

Hmm here we could be a bit more clever and do a recipe type filtering:

# filter by recipe type
cands = [d for d in descr if d.endswith('.' + recipe)]
if len(cands) > 1:
    # filter by pac, repo, and recipe type
    cands = [d for d in cands if d == '%s-%s.%s' % (pac, arg_repository, recipe)]

And now the code below...

>                  if len(cands) == 1:
>                      arg_descr = cands[0]
>                  else:
> -                    cands = [i for i in descr for ext in extensions if i == '%s.%s' % (pac, ext)]
> +                    if recipe:
> +                        if (recipe == 'spec' or recipe == 'dsc'):

Hmm IMHO the check below is recipe type independent and should be
always executed.

> +                            if len([d for d in descr if d.endswith('.' + recipe)]) > 1:
> +                                raise oscerr.WrongArgs("More than one build description found for repository type. Please specify build description")
> +                        cands = [i for i in descr if i == '%s.%s' % (pac, recipe) or i == recipe]
> +                    else:
> +                        cands = [i for i in descr for ext in extensions if i == '%s.%s' % (pac, ext)]

If we can guarantee that "recipe" is always something reasonable (e.g.,
queryconfig did not fail (see below)), then we could remove the else part
(dead code).

> +def return_external(filename, *args, **kwargs):
> +    """Executes the program filename via subprocess.Popen.
> +
> +    *args are additional arguments which are passed to the
> +    program filename. **kwargs specify additional arguments for
> +    the subprocess.call function.

This should be subprocess.Popen, but so far **kwargs are unused (maybe
just pass it to the Popen constructor?)

> +    if no args are specified the plain filename is passed
> +    to subprocess.call (this can be used to execute a shell

subprocess.Popen

> +    command). Otherwise [filename] + list(args) is passed
> +    to the subprocess.call function.
> +

subprocess.Popen constructor

> +    Returns the output of the command.
> +
> +    """
> +    if args:
> +        cmd = [filename] + list(args)
> +    else:
> +        cmd = filename
> +
> +    try:
> +        ret = subprocess.Popen(cmd, stdout=subprocess.PIPE)

Instead of subprocess.Popen, we could also use subprocess.check_ouput
(advantage: it checks the return code) [1]. For our use-case above, it
would be guaranteed that the recipe (type) is something "reasonable".

[1] https://docs.python.org/2/library/subprocess.html#subprocess.check_output

lethliel:
- The idea with the tempfiles is very good. Done :) 
- I also switched to check_output, which is a nicer function than Popen
- refactored build description filtering
- fixed some spelling errors and wording

lethliel:
changed return_external call

marcus-h:
On 2017-02-01 00:57:31 -0800, Marco Strigl wrote:
> - The idea with the tempfiles is very good. Done :) 
> - I also switched to check_output, which is a nicer function than Popen
> - refactored build description filtering

It seems the <pac>.<recipe> case was dropped (see below). Otherwise
it looks great!

> @@ -5825,13 +5840,16 @@ Please submit there instead, or use --nodevelproject to force direct submission.
>                  if is_package_dir(os.getcwd()):
>                      pac = store_read_package(os.getcwd())
>                  extensions = ['spec', 'dsc', 'kiwi', 'livebuild']

Unused - so let's remove it:)

> -                cands = [i for i in descr for ext in extensions if i == '%s-%s.%s' % (pac, arg_repository, ext)]
> +                if recipe == 'PKGBUILD':
> +                    cands = [d for d in descr if d.startswith(recipe)]
> +                else:
> +                    cands = [d for d in descr if d.endswith('.' + recipe)]
> +                if len(cands) > 1:
> +                    cands = [d for d in cands if d == '%s-%s.%s' % (pac, arg_repository, recipe)]
> +                    if len([d for d in descr if d.endswith('.' + recipe)]) > 1:
> +                        raise oscerr.WrongArgs("More than one build description found for repository type. Please specify build description")

Hmm the if statement makes no sense?! (if we reach this if statement in
the non PKGBUILD case, we will always raise an exception). Just remove
it:)

>                  if len(cands) == 1:
>                      arg_descr = cands[0]
> -                else:
> -                    cands = [i for i in descr for ext in extensions if i == '%s.%s' % (pac, ext)]
> -                    if len(cands) == 1:
> -                        arg_descr = cands[0]

We sill need something like this for the <pac>.<recipe> case. It could
be integrated into the len(cands) > 1 above. E.g., something along
these lines

if len(cands) > 1:
    tmp_cands = cands
    cands = [d for d in cands if d == '%s-%s.%s' % (pac, arg_repository, recipe)]
    if not cands:
        cands = [d for d in tmp_cands if d == '%s.%s' % (pac, recipe)]

>                  if not arg_descr:
>                      msg = 'Multiple build description files found: %s' % ', '.join(descr)
>              elif not ignore_descr:

Ps. why do we need the "no_repo" handling?:)

lethliel:
> Hmm the if statement makes no sense?! (if we reach this if statement in the non PKGBUILD case, we will 
> always raise an exception). Just remove it:)

The exception is just thrown if there are more than one build description of one type. So I will change to check if there is a repository based spec file with 
`tmp_cands = [d for d in cands if d == '%s-%s.%s' % (pac, arg_repository, recipe)]`

and use this one. (If there is only one)


> It seems the \<pac\>.\<recipe\> case was dropped (see below)

So if there is a build description file which matches exactly the package name we should use it to build even if there are more than one build description file? 
I am fine with that. Just want to be sure. 

> why do we need the "no_repo" handling?

It is needed if you have multiple repositories in a project and don't tell osc which one to use.

lethliel:
I now implemented something like this:

```
if len(cands) > 1:
    repo_cands = cands
    pac_cands = cands
    repo_cands = [d for d in cands if d == '%s-%s.%s' % (pac, arg_repository, recipe)]
    if repo_cands:
        cands = repo_cands
    else:
        pac_cands = [d for d in cands if d == '%s.%s' % (pac, recipe)]
        if pac_cands:
            cands = pac_cands
    if len(cands) > 1:
        raise oscerr.WrongArgs("More than one build description found for repository type. Please specify build description")
```

which looks if there is a build description based on repo. (e.g. mypack-openSUSE_Leap_42.2.spec). 
If none found it looks for a build description matching exactly the package name (e.g. mypack.spec).
If none found it raises an error

marcus-h:
On 2017-02-01 04:52:59 -0800, Marco Strigl wrote:
> > Hmm the if statement makes no sense?! (if we reach this if statement in the non PKGBUILD case, we will 
> > always raise an exception). Just remove it:)
> 
> The exception is just thrown if there are more than one build description of one type.

Ah ok, if that is your intention then

if len(cands) > 1:
    raise...

would have been more concise;)

> > It seems the \<pac\>.\<recipe\> case was dropped (see below)
> 
> So if there is a build description file which matches exactly the package name we should use it to build even if there are more than one build description file? 
> I am fine with that. Just want to be sure. 
> 
Yep.

> > why do we need the "no_repo" handling?
> 
> It is needed if you have multiple repositories in a project and don't tell osc which one to use. 
> 
parse_repoarchdescr tries to guess a repo, too. If no suitable repo
was found, it will error out before. At this point "arg_repository"
contains a "valid" value.

marcus-h:
On 2017-02-01 05:22:45 -0800, Marco Strigl wrote:
> I now implemented something like this:
> 
> ```
> if len(cands) > 1:
>     repo_cands = cands
>     pac_cands = cands

Both assignments can be removed.

>     repo_cands = [d for d in cands if d == '%s-%s.%s' % (pac, arg_repository, recipe)]
>     if repo_cands:
>         cands = repo_cands
>     else:
>         pac_cands = [d for d in cands if d == '%s.%s' % (pac, recipe)]
>         if pac_cands:
>             cands = pac_cands
>     if len(cands) > 1:
>         raise oscerr.WrongArgs("More than one build description found for repository type. Please specify build description")

The error checking is done later in the code.

> which looks if there is a build description based on repo. (e.g. mypack-openSUSE_Leap_42.2.spec). 
> If none found it looks for a build description matching exactly the package name (e.g. mypack.spec).
> If none found it raises an error
> 
Yep that's how it is supposed to be:)

(Sorry for being picky... :/)

lethliel:
Yeah :) 
removed the assignments. 
Let the error handling be done later in the code. (already existed. You are right.) But i changed the joined output from **descr** to **cands**. I think it fits better in this case.

I don't have a problem with someone being picky. As I am still a python noob it is the best way to learn.
So. Don't worry :)

marcus-h:
On 2017-02-01 10:16:22 -0800, Marco Strigl wrote:
> Yeah :) 
> removed the assignments. 
> Let the error handling be done later in the code. (already existed. You are right.) But i changed the joined output from **descr** to **cands**. I think it fits better in this case.
> 
Looks good:)

What about the "no_repo" thing?:) (see also the last comment in
https://github.com/openSUSE/osc/pull/260#issuecomment-276672316)

>                  pac = os.path.basename(os.getcwd())
> -                if is_package_dir(os.getcwd()):
> -                    pac = store_read_package(os.getcwd())

We should keep this.

> index 9bb66fb..813113b 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -7310,6 +7310,32 @@ def run_external(filename, *args, **kwargs):

<SNIP>

> +    if no args are specified the plain filename is passed
> +    to subprocess.call (this can be used to execute a shell

subprocess.check_output

lethliel:
> We should keep this.

I thought you said in https://github.com/openSUSE/osc/pull/260#issuecomment-276637457 it is unused. I might have misunderstood you there :) 

I will put it back in. 

> What about the "no_repo" thing?:) (see also the last comment in #260 (comment))

The repository guessing works fine if you have just one repository. But if you have multiple repos it always takes the last, when invoking `osc build`. So I am aware that I am bypassing the repository guessing here, but if you have multiple build description and multiple repositories (arch / debian / spec) a repo guessing makes no sense IMHO

I have the following repositories: 

```
xUbuntu_16.10       i586
xUbuntu_16.10       x86_64
openSUSE_Leap_42.2  x86_64
openSUSE_Leap_42.1  x86_64
Debian_8.0          i586
Debian_8.0          x86_64
Arch_Extra          i586
Arch_Extra          x86_64
Arch_Core           i586
Arch_Core           x86_64
```

and it always takes the last.

marcus-h:
On 2017-02-02 00:50:08 -0800, Marco Strigl wrote:
> > We should keep this.
> 
> I thought you said in https://github.com/openSUSE/osc/pull/260#issuecomment-276637457 it is unused. I might have misunderstood you there :) 
> 
Ah sorry - I just wanted to give a bit context there... but yes that made
it quite ambiguous:)

> I will put it back in. 
> 
> > What about the "no_repo" thing?:) (see also the last comment in #260 (comment))
> 
> The repository guessing works fine if you have just one repository. But if you have multiple repos it always takes the last, when invoking `osc build`.

It depends on the available repos - taking the last one is just a
fallback if the configured "build_repository" is not present and
no other heuristic worked.

> So I am aware that I am bypassing the repository guessing here, but if you have multiple build description and multiple repositories (arch / debian / spec) a repo guessing makes no sense IMHO
> 
Ok. What about the following compromise?:)
If one of the heuristics match, we set "no_repo" back to "False".

> I have the following repositories: 
> 
> ```
> xUbuntu_16.10       i586
> xUbuntu_16.10       x86_64
> openSUSE_Leap_42.2  x86_64
> openSUSE_Leap_42.1  x86_64
> Debian_8.0          i586
> Debian_8.0          x86_64
> Arch_Extra          i586
> Arch_Extra          x86_64
> Arch_Core           i586
> Arch_Core           x86_64
> ```
> 
> and it always takes the last. 
> 
If you run "osc config set general build_repository Debian_8.0" before,
it'll select Debian_8.0.

lethliel:
> Ok. What about the following compromise?:) If one of the heuristics match, we set "no_repo" back to "False".

Yes :) no_repo is now set to false if one of the heuristics matches. 

> If you run "osc config set general build_repository Debian_8.0" before, it'll select Debian_8.0.

Oh. I wasn't aware of that. Thanks for the information.

marcus-h:
On 2017-02-02 03:28:50 -0800, Marco Strigl wrote:
> > Ok. What about the following compromise?:) If one of the heuristics match, we set "no_repo" back to "False".
> 
> Yes :) no_repo is now set to false if one of the heuristics matches. 
> 
Great! Merged. Thanks a lot!

PR #263, opened 2017-02-08
fixes osc checkout error with checkout_rooted=1

lethliel:
This fixes a bug reported from Martin Wilck (see bsc#1012592)
He submitted the patch via bugzilla. Just added the following part: 

```
if root_dots == '.':
     root_dots = ''
```

Otherwise root_dots would be `...` and the new path would be **...**/\<proj\>/\<pack\>/

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -4799,16 +4799,19 @@ def checkout_package(apiurl, project, package,
                 # if project roots were previously inconsistent
                 root_dots = "../../"
             if is_project_dir(root_dots):
+                oldproj = store_read_project(root_dots)
                 if conf.config['checkout_no_colon']:
-                    oldproj = store_read_project(root_dots)
                     n = len(oldproj.split(':'))
                 else:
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2017-02-08 05:02:30 -0800, Marco Strigl wrote:
> This fixes a bug reported from Martin Wilck (see bsc#1012592)
> He submitted the patch via bugzilla. Just added the following part: 
> 
> ```
> if root_dots == '.':
>      root_dots = ''
> ```
> 
> Otherwise root_dots would be `...` and the new path would be **...**/\<proj\>/\<pack\>/
> 
LGTM. Merged. Thanks!

(Ps. in case of a multi line commit message there should be an empty
line between the "summary" (first line) and "description")

lethliel:
> (Ps. in case of a multi line commit message there should be an empty line between the "summary" (first line) and "description")

ok. I will consider it in the future.

Issue #264, opened 2017-02-11
osc fails if $PAGER variable contains command + arguments


peti:
For example:

~~~
PAGER="less -R" osc sr -d devel:languages:haskell:lts:7 HaXml devel:languages:haskell
less -R: No such file or directory
~~~

Most other programs interpret this setting correctly, i.e. this is the default setting of `$PAGER` on some distributions.

PR #265, opened 2017-02-11
Lex the contents of $PAGER to support pager commands with arguments.

peti:
Closes https://github.com/openSUSE/osc/issues/264.


marcus-h:
Merged. Thanks a lot!

PR #271, opened 2017-02-27
added syntax check if no token opt is set

lethliel:
Check if the user accidentally typed

`osc token create project package`

instead of 

`osc token --create project package`

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -725,6 +725,8 @@ class Osc(cmdln.Cmdln):
             fd = urlopen(req, data=None)
             print(fd.read())
         else:
+            if args and args[0] in ['create', 'delete', 'trigger']:
+                raise oscerr.WrongArgs("Did you mean --" + args[0] + "?")
             # just list token
             for data in streamfile(url, http_GET):
                 sys.stdout.write(data)


marcus-h:
Hmm this change breaks the complete "else" part (the "token list").

Something like

elif args:
    raise oscerr.WrongArgs('Unexpected arguments: \'%s\'' % ', '.join(args))
else:
    # old else part...

should do the trick.

lethliel:
On Wed, Mar 01, 2017 at 10:56:20AM -0800, Marcus Hüwe wrote:
> Hmm this change breaks the complete "else" part (the "token list").
>

Ups. You are right. Silly mistake. 
 
> Something like
> 
> elif args:
>     raise oscerr.WrongArgs('Unexpected arguments: \'%s\'' % ', '.join(args))
> else:
>     # old else part...
> 

I changed it to:

if args and args[0] in ['create', 'delete', trigger']

marcus-h:
On 2017-03-02 00:05:47 -0800, Marco Strigl wrote:
> On Wed, Mar 01, 2017 at 10:56:20AM -0800, Marcus Hüwe wrote:
> > Something like
> > 
> > elif args:
> >     raise oscerr.WrongArgs('Unexpected arguments: \'%s\'' % ', '.join(args))
> > else:
> >     # old else part...
> > 
> 
> I changed it to:
> 
> if args and args[0] in ['create', 'delete', trigger'] 
> 
Ok. Merged. Thanks!

PR #273, opened 2017-03-07
Add missing whitespaces in commandline options

FelixZhang:


marcus-h:
Merged. Thanks for the patch!

PR #275, opened 2017-03-09
cat/less/blame command: default to expand to stay in sync with checkout

adrianschroeter:
this allows us also to drop the nasty fallback code

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,5 @@
 0.158
-  -
+  - cat/less/blame command: default to expand to stay in sync with checkout
 
 0.157
   - add unpublish command (requires OBS 2.8)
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2017-03-09 03:03:30 -0800, Adrian Schröter wrote:
> this allows us also to drop the nasty fallback code

Hmm we could do that and for the blame command it makes a lot of
sense (IMHO). Now my usual complaint;) we break the existing UI.

An alternative would be to introduce a dedicated blame command
(instead of an alias) with its own defaults (but this would
increase the "inconsistency" among commands (e.g., blame and
cat/less)...).

Now choose between the devil and the deep blue sea...

adrianschroeter:
merged by accident ... but we agreed on it anyway on IRC, IIRC

PR #278, opened 2017-03-09
Don't ask api for defined services in local mode

k0da:
Signed-off-by: Dinar Valeev <k0da@opensuse.org>


marcus-h:
Hmm this fix does not conform to the intended semantics of
"local"/"localrun". Even in a local run, we have to take the
project services into account.
I just fixed it in commit 9d0725b. Anyway, thanks for the PR!

PR #279, opened 2017-03-14
Add plugin directory to search path

lnussel:
Some osc plugins ship with extra modules that are not globally installed
as python modules. So expand search path.


marcus-h:
Merged. Thanks!

PR #284, opened 2017-04-07
filter excluded repos for package by default

lethliel:
This changes the default behavior to **not** show the excluded repos. 
It is more clearer, especially in large multibuild environments.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -4995,6 +4995,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                         help='generate output in CSV format')
     @cmdln.option('', '--format', default='%(repository)s|%(arch)s|%(state)s|%(dirty)s|%(code)s|%(details)s',
                         help='format string for csv output')
+    @cmdln.option('--show-excluded', action='store_true',
+                        help='show repos that are excluded for this package')
     def do_results(self, subcmd, opts, *args):
         """${cmd_name}: Shows the build results of a package or project
 
@@ -5044,7 +5046,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
 
         kwargs = {'apiurl': apiurl, 'project': project, 'package': package,
                   'lastbuild': opts.last_build, 'repository': opts.repo,
-                  'arch': opts.arch, 'wait': opts.watch}
+                  'arch': opts.arch, 'wait': opts.watch, 'showexcl': opts.show_excluded}
         if opts.multibuild_package:
             opts.no_multibuild = False
             kwargs['multibuild_packages'] = opts.multibuild_package
--- a/osc/core.py
+++ b/osc/core.py
@@ -5543,9 +5543,12 @@ def get_results(apiurl, project, package, verbose=False, printJoin='', *args, **
     r = []
     printed = False
     multibuild_packages = kwargs.pop('multibuild_packages', [])
+    show_excluded = kwargs.pop('showexcl', False)
     for results in get_package_results(apiurl, project, package, **kwargs):
         r = []
         for res, is_multi in result_xml_to_dicts(results):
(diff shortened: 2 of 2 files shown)


marcus-h:
On 2017-04-07 01:06:26 -0700, Marco Strigl wrote:
> This changes the default behavior to **not** show the excluded repos. 
> It is more clearer, especially in large multibuild environments.
> 
Hrm yes, we can do that (let's hope this does not confuse users if
they, for instance, just upload a *.spec and try to build against a
debian repo (in this case "osc results" shows nothing))

> index 44083fa..49f35df 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -5543,9 +5543,12 @@ def get_results(apiurl, project, package, verbose=False, printJoin='', *args, **
>      r = []
>      printed = False
>      multibuild_packages = kwargs.pop('multibuild_packages', [])
> +    show_excluded = kwargs.pop('showexcl')

This will break all callers except the one in do_results:

show_excluded = kwargs.pop('showexcl', False)

>      for results in get_package_results(apiurl, project, package, **kwargs):
>          r = []
>          for res, is_multi in result_xml_to_dicts(results):
> +            if show_excluded == None and res['code'] == 'excluded':

Hmm in this case it is better to use "not show_excluded" (this takes
showexcl=False/None into account and results in a "cleaner" api)
(otherwise "show_excluded is None" instead of "show_excluded == None").

lethliel:
On 04/07/2017 11:01 AM, Marcus Hüwe wrote:
> On 2017-04-07 01:06:26 -0700, Marco Strigl wrote:
>> This changes the default behavior to **not** show the excluded repos.
>> It is more clearer, especially in large multibuild environments.
>>
> Hrm yes, we can do that (let's hope this does not confuse users if
> they, for instance, just upload a *.spec and try to build against a
> debian repo (in this case "osc results" shows nothing))
> 

I talked to a view people and all think it is clearer if the excluded repos are not 
shown by default. But If the users will get confused we can change it back and blame 
the change on me ;)

>> index 44083fa..49f35df 100644
>> --- a/osc/core.py
>> +++ b/osc/core.py
>> @@ -5543,9 +5543,12 @@ def get_results(apiurl, project, package, verbose=False, printJoin='', *args, **
>> r = []
>> printed = False
>> multibuild_packages = kwargs.pop('multibuild_packages', [])
>> + show_excluded = kwargs.pop('showexcl')
> 
> This will break all callers except the one in do_results:
> 
> show_excluded = kwargs.pop('showexcl', False)
> 
>> for results in get_package_results(apiurl, project, package, **kwargs):
>> r = []
>> for res, is_multi in result_xml_to_dicts(results):
>> + if show_excluded == None and res['code'] == 'excluded':
> 
> Hmm in this case it is better to use "not show_excluded" (this takes
> showexcl=False/None into account and results in a "cleaner" api)
> (otherwise "show_excluded is None" instead of "show_excluded == None").
> 

Yes. Forgot to set the default value for the pop method. 
Changed the condition to "not show_excluded"

marcus-h:
On 2017-04-07 02:27:15 -0700, Marco Strigl wrote:
> On 04/07/2017 11:01 AM, Marcus Hüwe wrote:
> > On 2017-04-07 01:06:26 -0700, Marco Strigl wrote:
> >> This changes the default behavior to **not** show the excluded repos.
> >> It is more clearer, especially in large multibuild environments.
> >>
> > Hrm yes, we can do that (let's hope this does not confuse users if
> > they, for instance, just upload a *.spec and try to build against a
> > debian repo (in this case "osc results" shows nothing))
> > 
> 
> I talked to a view people and all think it is clearer if the excluded repos are not 
> shown by default. But If the users will get confused we can change it back and blame 
> the change on me ;)
> 
Merged. Thanks again for the PR!

Issue #236, opened 2016-10-02
--no-cleanup unaffective with --separate-requests


Vogtinator:
Although I made submit requests with `osc sr --no-cleanup --seperate-requests -m "Update to 5.8.0 (unreleased tarballs, may not be final)"`, the source packages got removed after accepting.

lethliel:
fixed in #285

PR #285, opened 2017-04-11
fixes issue #236

lethliel:
fixes issue #236 

If **--no-cleanup** is set the source package will not be deleted anymore using the **--separate-requests** for a project


marcus-h:
Merged. Thanks a lot!

PR #287, opened 2017-04-19
fixes issue #175 / invalid input at prompt

### LINKED ISSUE #175: `osc sr` "supersede?" prompt continues with default (presumably) option even if given invalid input
```
yac@linux-apw9.site % osc sr devel:tools:scm bs-update openSUSE:Factory
Warning: failed to fetch meta data for 'openSUSE:Factory' package 'bs-update' (new package?) 
There are already the following submit request: 332259.
Supersede the old requests? (y/n/c) ?
Default log message was not changed. Press 'c' to continue.
a)bort, c)ontinue, e)dit: a
aborted.
```

I was curious about the "c" option, assuming it refers to "cancel" I was wondering what gets cancelled here so I tried "?" for help.

lethliel:
fixes #175 

The user must enter a valid option to proceed.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -4002,9 +4002,11 @@ def edit_text(data='', delim=None, suffix='.txt', template=''):
         (fd, filename) = tempfile.mkstemp(prefix='osc-editor', suffix=suffix)
         os.close(fd)
         mtime = os.stat(filename).st_mtime
+        ri_err = False
         while True:
-            file_changed = _edit_message_open_editor(filename, data, mtime)
-            msg = open(filename).read()
+            if not ri_err:
+                file_changed = _edit_message_open_editor(filename, data, mtime)
+                msg = open(filename).read()
             if delim:
                 msg = msg.split(delim)[0].rstrip()
             if msg and file_changed:
@@ -4019,6 +4021,11 @@ def edit_text(data='', delim=None, suffix='.txt', template=''):
                 elif ri in 'cC':
                     break
                 elif ri in 'eE':
+                    ri_err = False
+                    pass
+                else:
+                    print("%s is not a valid option." % ri)
+                    ri_err = True
                     pass
     finally:
         os.unlink(filename)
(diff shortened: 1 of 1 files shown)


marcus-h:
Sorry for the long delay... I just reviewed the second part of the
patch (will do the other part ASAP).

> @@ -4417,6 +4424,10 @@ def check_existing_requests(apiurl, src_project, src_package, dst_project,
>          print('There are already the following submit request: %s.' % \
>                ', '.join([i.reqid for i in reqs]))
>          repl = raw_input('Supersede the old requests? (y/n/c) ')
> +        if repl.lower() not in ['c', 'y', 'n']:
> +            print('%s is not a valid option.' % repl.lower())

Maybe no .lower()? (just print out what the user entered...)

> +            check_existing_requests(apiurl,src_project, src_package, dst_project,
> +                                    dst_package)

Hmm probably a return missing (otherwise it's "useless")? Nevertheless, we
shouldn't recurse here (no need to do the http request again).
Maybe something like this

while repl.lower() not in [...]:
    print('foo')
    repl = raw_input(...)

>          if repl.lower() == 'c':
>              print('Aborting', file=sys.stderr)
>              raise oscerr.UserAbort()
> @@ -4435,6 +4446,10 @@ def check_existing_maintenance_requests(apiurl, src_project, src_packages, dst_p
>          print('There are already the following maintenance incident request: %s.' % \
>                ', '.join([i.reqid for i in reqs]))
>          repl = raw_input('Supersede the old requests? (y/n/c) ')
> +        if repl.lower() not in ['c', 'y', 'n']:
> +            print('%s is not a valid option.' % repl.lower())
> +            check_existing_maintenance_requests(apiurl,src_project, src_package, dst_project,
> +                                    release_project)

See above.

lethliel:
No problem. Take your time. 

I've made the changes.

marcus-h:
Merged. Thanks a lot!

PR #290, opened 2017-04-19
allow to pass any option to the build command

bmwiedemann:
so that we do not have to patch osc for every parameter added to build
e.g. "--vm-custom-opt" in my case

similar in nature to gcc'c -Wl, option


marcus-h:
Sorry for the long delay! LGTM. Merged. Thanks!

PR #291, opened 2017-04-21
fixed chroot to use proc

### LINKED ISSUE #288: osc chroot must mount procfs
```
Name        : osc
Version     : 0.157.2
Release     : 1.1
Distribution: openSUSE Tumbleweed
```

`osc chroot` currently does not set up procfs, which causes problems like
```
$ osc chroot
running: sudo chroot /var/tmp/jng/openSUSE_Factory-x86_64 su - abuild
$ cd /usr/src/packages/SOURCES/; rpmbuild -bi --short-circuit blah.spec
[...]
+ /usr/lib/rpm/brp-compress
+ /usr/lib/rpm/brp-suse
calling /usr/lib/rpm/brp-suse.d/brp-05-permissions
setting / to root:root 0755. (wrong owner/group abuild:abuild)
setting /usr/ to root:root 0755. (wrong owner/group abuild:abuild)
calling /usr/lib/rpm/brp-suse.d/brp-15-strip-debug
/usr/lib/rpm/brp-suse.d/brp-15-strip-debug: line 33: /dev/fd/62: No such file or directory
/usr/lib/rpm/brp-suse.d/brp-15-strip-debug: line 47: /dev/fd/62: No such file or directory
error: Bad exit status from /var/tmp/rpm-tmp.IOvCKJ (%install)

abuild@a4:/home/abuild> namei /dev/fd/62
f: /dev/fd/62
 d /
 d dev
 l fd -> /proc/self/fd
   d /
   d proc
     self - No such file or directory
```

lethliel:
fixes #288 

proc is now mounted (similar to build script). 
If the mount fails an error is displayed, but you get chrooted nevertheless.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -9,6 +9,7 @@ from . import cmdln
 from . import conf
 from . import oscerr
 import sys
+import signal
 import time
 import imp
 import inspect
@@ -6272,6 +6273,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             osc chroot [OPTS] (REPOSITORY = build_repository (config option), ARCH = hostarch, BUILD_DESCR is detected automatically)
         ${cmd_option_list}
         """
+
         if len(args) > 3:
             raise oscerr.WrongArgs('Too many arguments')
         if conf.config['build-type'] and conf.config['build-type'] != "lxc":
@@ -6325,13 +6327,38 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         sucmd = suwrapper.split()[0]
         suargs = ' '.join(suwrapper.split()[1:])
         if suwrapper.startswith('su '):
+            mntproc = [sucmd, '%s mount -n -tproc none %s/proc' % (suargs, buildroot)]
+            umntproc = [sucmd, '%s umount %s/proc' % (suargs, buildroot)]
             cmd = [sucmd, '%s chroot "%s" su - %s' % (suargs, buildroot, user)]
         else:
+            mntproc = [sucmd, 'mount', '-n', '-tproc' , 'none', '%s/proc' % buildroot]
+            umntproc = [sucmd, 'umount', '%s/proc' % buildroot]
             cmd = [sucmd, 'chroot', buildroot, 'su', '-', user]
             if suargs:
+                mntproc[1:1] = suargs.split()
+                umntproc[1:1] = suargs.split()
                 cmd[1:1] = suargs.split()
+
+        #signal handler for chroot procfs umount
+        def umount_proc(signum = None, frame = None, ret=1):
+            subprocess.call(umntproc)
+            sys.exit(ret)
+        for sig in [signal.SIGTERM, signal.SIGINT, signal.SIGHUP, signal.SIGQUIT]:
+            signal.signal(sig, umount_proc)
+
+        print('mounting proc: %s' % ' '.join(mntproc))
+        mount_err = -1
+        mount_err = subprocess.call(mntproc)
+        if mount_err > 0:
+            print('There was an error mounting proc. Please check mountpoints in chroot')
         print('running: %s' % ' '.join(cmd))
-        os.execvp(sucmd, cmd)
+        retval = 0
+        try:
+            retval = subprocess.call(cmd)
+        finally:
+            if not mount_err or mount_err == 32:
+                print('unmounting %s/proc ...' % buildroot)
+                umount_proc(ret=retval)
 
 
     @cmdln.option('', '--csv', action='store_true',


marcus-h:
On 2017-04-21 01:22:13 -0700, Marco Strigl wrote:
> fixes #288 
> 
> proc is now mounted (similar to build script). 

In the long-term, the best is if we outsource the chrooting
to "build --shell..." command (but the obs-build script needs a
few minor changes for this, IIRC). Then we get chrooting into
a vm for free...

Minor remarks:
- in the non "su" case we should also take suargs into account
- we should always umount proc again (probably the best is to
  in a try-finally). E.g. something like
  mount_err = -1
  try:
      mount_err = subprocess...
      ...
  finally:
      if not mount_err:
         # umount code here

As usual: thanks for the patch!:)

lethliel:
On 04/21/2017 01:54 PM, Marcus Hüwe  wrote:
> On 2017-04-21 01:22:13 -0700, Marco Strigl wrote:
>> fixes #288
>>
>> proc is now mounted (similar to build script).
> 
> In the long-term, the best is if we outsource the chrooting
> to "build --shell..." command (but the obs-build script needs a
> few minor changes for this, IIRC). Then we get chrooting into
> a vm for free...

Yes. In the long-term this will be the best solution...

> 
> Minor remarks:
> - in the non "su" case we should also take suargs into account

Yes. I see :) 

> - we should always umount proc again (probably the best is to
> in a try-finally). E.g. something like
> mount_err = -1
> try:
> mount_err = subprocess...
> ...
> finally:
> if not mount_err:
> # umount code here
> 

This will not work. The problem is the os.execvp call which does not come back.

lethliel:
@marcus-h:

I changed the call of chroot from execvp to subprocess.call. 
Now I can call a finally block :) 

Do you see any downsides on this implementation?

marcus-h:
On 2017-04-21 05:36:02 -0700, Marco Strigl wrote:
> On 04/21/2017 01:54 PM, Marcus Hüwe  wrote:
> > - we should always umount proc again (probably the best is to
> > in a try-finally). E.g. something like
> > mount_err = -1
> > try:
> > mount_err = subprocess...
> > ...
> > finally:
> > if not mount_err:
> > # umount code here
> > 
> 
> This will not work. The problem is the os.execvp call which does not come back. 
> 
Arghs right:/ - good catch!

lethliel:
On 04/21/2017 03:38 PM, Marcus Hüwe  wrote:
> On 2017-04-21 05:36:02 -0700, Marco Strigl wrote:
>> On 04/21/2017 01:54 PM, Marcus Hüwe wrote:
>> > - we should always umount proc again (probably the best is to
>> > in a try-finally). E.g. something like
>> > mount_err = -1
>> > try:
>> > mount_err = subprocess...
>> > ...
>> > finally:
>> > if not mount_err:
>> > # umount code here
>> >
>>
>> This will not work. The problem is the os.execvp call which does not come back.
>>
> Arghs right:/ - good catch!

I changed it a little bit. Now the unmounting works. But i had to switch from execvp to subprocess.call. Does this have any downsides?

lethliel:
On 04/21/2017 03:51 PM, Marcus Hüwe  wrote:
> The problem is the following:
> - user runs osc chroot...
> - user closes terminal (e.g. she didn't type "exit" to
> leave the chroot)
> 
> In this case the finally block isn't executed and we might
> end up with several proc mounts for the same chroot (if the
> above pattern is repeated). To do this properly, we probably
> have register some signal handlers (e.g. for SIGHUP etc.)...
>

You are right. The implementation of signal handlers would be the 
cleanest way. I will do it and rebase when I am ready.

lethliel:
I've implemented the signal handler. It unmounts $buildroot/proc on SIGHUP, SIGTERM, SIGINT and SIGQUIT.

It does not check if the mount point is there. It makes no sense to me. Just unmount it. If ii is not there nothing will happen.

marcus-h:
On 2017-04-24 02:37:05 -0700, Marco Strigl wrote:
> I've implemented the signal handler. It unmounts $buildroot/proc on SIGHUP, SIGTERM, SIGINT and SIGQUIT.
> 
Cool! We should install the signal handlers once the "umountproc"
variable was defined (after the else part) (otherwise, it could lead
to a NameError exception).
Maybe use sys.exit(1) instead of sys.exit(0)?

(Btw., you could also rename "handler" to "umount_proc" and call it in
the "normal" control flow (instead of the subprocess.call(umount_proc)
calls) (then an additional flag for the sys.exit behavior is needed).
But the current code is perfectly ok.)

>              if suargs:
> +                mntproc[1:1] = suargs.slpit()

Typo;)

> It does not check if the mount point is there. It makes no sense to me. Just unmount it. If ii is not there nothing will happen. 
> 
Yep, sounds reasonable.

nickbroon:
:+1:

nickbroon:
Any plans to merge this soon?

marcus-h:
On 2017-05-24 03:03:55 -0700, Nick Brown wrote:
> Any plans to merge this soon?
> 
Uh - I haven't noticed that the patch changed after my
last comments (it would be nice if github would send a
notification in this case... so for now an explicit
"ping" is needed;) ).
Anyway... I'll review it until sunday (at the latest).

marcus-h:
Looks almost good, except that we always exit with status 1. Maybe
we can do something like this:
def umount_proc(signum=None, frame=None, ret=1):
    subprocess.call(...)
    sys.exit(ret)

And in the finally clause simply call "umount_proc(ret=0)".

(alternatively, we could also change umount_proc so that
it only calls sys.exit if ret != 0).

lethliel:
@marcus-h: I made the changes from your last comment and rebased the pull request. Can you please have a look? Thanks.

marcus-h:
On 2017-06-07 00:47:39 -0700, Marco Strigl wrote:
> @marcus-h: I made the changes from your last comment and rebased the pull request. Can you please have a look? Thanks. 
> 
Great! Unfortunately, 2 small changes are still needed (sorry, I missed
them in the previous reviews:/):

>              if suargs:
> +                mntproc[1:1] = suargs.split()
>                  cmd[1:1] = suargs.split()

We should do the same for umntproc.

> -        os.execvp(sucmd, cmd)
> +        try:
> +            subprocess.call(cmd)
> +        finally:
> +            if not mount_err or mount_err == 32:
> +                print('unmounting %s/proc ...' % buildroot)
> +                umount_proc(ret=0)

We should also take the retval of the subprocess.call into account.
Something like:

ret = 0
try:
    ret = subprocess.call(cmd)
finally:
    if ...:
        ...
        umount_proc(ret=ret)

Sorry again for the "back and forth"...

lethliel:
@nickbroon @marcus-h Sorry. I was on parental leave and had no time to check.
I made the changes Marcus suggested. I think this should be finally ready to merge.

marcus-h:
On 2017-07-31 12:53:41 +0000, Marco Strigl wrote:
> @nickbroon @marcus-h Sorry. I was on parental leave and had no time to check.
> I made the changes Marcus suggested. I think this should be finally ready to merge.
> 
Merged. Thanks!

PR #292, opened 2017-04-23
Using OSC on OpenBSD

andreasbaumann:
Didn't really know where to put those things. There is a README for which packages to install
on OpenBSD. And `/etc/ssl/cert.pem` is an additional location where certificates are stored
on OpenBSD.

I just tested basic osc functionality like checkout, checkin, update..

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,38 @@
+0.158
+  - cat/less/blame command: default to expand to stay in sync with checkout
+  - add support for highly experimental native appimage.yml support
+  - show package status when repository configuration is broken
+
+0.157.2
+  - add compat code for older apis that do not support multibuild
+  - fix ssl.connection instantiation in case of old M2Crypto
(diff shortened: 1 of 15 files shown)


marcus-h:
Hi,

first of all thanks the PR!

On 2017-04-23 10:19:44 -0700, Andreas Baumann wrote:
> Didn't really know where to put those things. There is a README for which packages to install
> on OpenBSD.

Hmm are those install instructions really needed (especially the extra
README.OpenBSD file)? I mean they do not differ too much from the "normal"
linux install instructions (see the README file).

What do you think?

> And `/etc/ssl/cert.pem` is an additional location where certificates are stored
> on OpenBSD.
> 
This part looks good to me.

andreasbaumann:
Yes, I agree, the build instructions are more or less common sense. For sure an extra
file is exagerated. Maybe an operating system section in README itself?

PR #295, opened 2017-05-08
fix test suite for new request creator handling

lethliel:
fixes test suite for new release


marcus-h:
I fixed it a bit differently in commit 7d3f8cd. Serializing an
empty creator attribute might lead to validation errors (in the
future?).

Thanks for pointing this out!

PR #306, opened 2017-07-07
Add compatibility with Debian's obs-build

bluca:
In Debian and Ubuntu build is renamed to obs-build for disambiguation
purposes.
Add a simple check to use the correct paths if running on Debian and
use /usr/bin/obs-build and /usr/lib/obs-build if so.

This is a constant source of errors and confusion for a lot of our users. Hopefully this very simple fix can put an end to that.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -12,6 +12,7 @@ import sys
 import time
 import imp
 import inspect
+import os
 try:
     from urllib.parse import urlsplit
     from urllib.error import HTTPError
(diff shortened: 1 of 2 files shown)


marcus-h:
Merged. Thanks for the PR!
(ps., actually, the "import os" in commandline.py is not needed,
but I prefer your explicit import as well:) ).

bluca:
Thanks! I prefer explicit imports too.

I have now forwarded the patch to Debian as well:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=867680

bluca:
I've also sent a request to make osc depend on "build | obs-build" instead of "build", so that it can be installed with Debian's version too:

https://build.opensuse.org/request/show/509034

PR #316, opened 2017-08-15
add --package-filter option to osc review list

lethliel:
This will filter the requests you get to review based on
the tgt_package in the action element.

This helps reviewers who want just to review a subset of requests or
do **not** want to review some packages.

Examples:

```
# osc review list --interactive --package-filter='^python.*' -G opensuse-review-team
# osc review list --interactive --package-filter='^(?!ghc).*' -G opensuse-review-team
```

@marcus-h: Can you review please ;) It would be better to get the filtered packages returned by the api. But AFAIK the api does not support regular expressions at this point. So this is the fastest thing I could come up with. 

As always: Feedback very welcome.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -2029,6 +2029,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                         help='exclude target project from request list')
     @cmdln.option('--involved-projects', action='store_true',
                         help='show all requests for project/packages where USER is involved')
+    @cmdln.option('--target-package-filter', metavar='TARGET_PACKAGE_FILTER',
+                        help='only list requests for the packages matching the package filter. A (python) regular expression is expected.')
     @cmdln.option('--source-buildstatus', action='store_true',
                         help='print the buildstatus of the source package (only works with "show" and the interactive review)')
     @cmdln.alias("rq")
@@ -2312,17 +2314,29 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             ## Workaround: fetch all, and filter on client side.
 
             ## FIXME: date filtering should become implemented on server side
+
+
+            if opts.target_package_filter:
+                filter_pattern = re.compile(opts.target_package_filter)
             for result in results:
-                if days == 0 or result.state.when > since or result.state.name == 'new':
-                    if (opts.interactive or conf.config['request_show_interactive']) and not opts.non_interactive:
-                        ignore_reviews = subcmd != 'review'
-                        request_interactive_review(apiurl, result, group=opts.group,
-                                                   ignore_reviews=ignore_reviews,
-                                                   source_buildstatus=source_buildstatus)
+                filtered = False
+                for action in result.actions:
+                    if action.type == 'group' or not opts.target_package_filter:
+                        continue
+                    if not filter_pattern.match(action.tgt_package):
+                        filtered = True
+                        break
+                if not filtered:
+                    if days == 0 or result.state.when > since or result.state.name == 'new':
+                        if (opts.interactive or conf.config['request_show_interactive']) and not opts.non_interactive:
+                            ignore_reviews = subcmd != 'review'
+                            request_interactive_review(apiurl, result, group=opts.group,
+                                                       ignore_reviews=ignore_reviews,
+                                                       source_buildstatus=source_buildstatus)
+                        else:
+                            print(result.list_view(), '\n')
                     else:
-                        print(result.list_view(), '\n')
-                else:
-                    skipped += 1
+                        skipped += 1
             if skipped:
                 print("There are %d requests older than %s days.\n" % (skipped, days))
 


marcus-h:
On 2017-08-15 06:04:03 -0700, Marco Strigl wrote:
> @marcus-h: Can you review please ;) It would be better to get the filtered packages returned by the api. But AFAIK the api does not support regular expressions at this point. So this is the fastest thing I could come up with. 
> 
In general LGTM - some minor remarks below:)

- We should probably call it --target-package-filter (since we just
  filter based on the target package)
- Is there a specific reason why we only support this for reviews?
  IMHO, it also makes sense for the other cases as well (then we
  can simply move the filtering into commandline.py (somewhere
  around line 2272))

> +    @cmdln.option('--package-filter', metavar='PACKAGE_FILTER',
> +                        help='only list requests for the packages matching the package filter.')

Maybe "... for target packages matching..."?

> @@ -4264,7 +4264,13 @@ def build_by(xpath, val):
>      for root in collection.findall('request'):
>          r = Request()
>          r.read(root)
> -        requests.append(r)
> +        if package_filter:
> +            filter_pattern = re.compile(package_filter)
> +            for action in r.actions:
> +                if (filter_pattern.match(action.tgt_package)):

- We should skip actions with type "group", because these actions
  have no "tgt_package" attribute
- Superfluous parentheses:)

> +                    requests.append(r)

We should add a "break" here (otherwise a request with multiple
actions might be added several times)

> +        else:
> +            requests.append(r)

osc/commandline.py
```diff
@@ -2312,17 +2314,27 @@ def do_request(self, subcmd, opts, *args):
...
-                    if (opts.interactive or conf.config['request_show_interactive']) and not opts.non_interactive:
-                        ignore_reviews = subcmd != 'review'
-                        request_interactive_review(apiurl, result, group=opts.group,
-                                                   ignore_reviews=ignore_reviews,
-                                                   source_buildstatus=source_buildstatus)
+                filtered = 0
+                for action in result.actions:
+                    if action.type == 'group' or not opts.target_package_filter:
+                        continue
+                    filter_pattern = re.compile(opts.target_package_filter)
```
darix: your first example is a glob match but you use regexp compile here. maybe the help text should mention that you expect regexp format. and maybe also that it is python regexp.

lethliel: I updated the example and the help text. The example worked nevertheless.

lethliel:
@marcus-h: 

Yes you are right. 

- I moved the filtering to commandline.py so that not only reviews are supported. 
- Also changed the naming to target-package-filter
- skip action type == 'group'

osc/commandline.py
darix: did it actually match the correct submitrequests? iirc you had python* which means it would match things like pythonnnnnnnnnnnn.

also wouldnt '^python.*' be enough? or do you really need the group?

lethliel: As already discussed in IRC it matched the right submitrequests. The group is not really needed. I will remove the grouping from the example to avoid confusion.

marcus-h:
On 2017-08-16 08:50:17 +0000, Marco Strigl wrote:
> - I moved the filtering to commandline.py so that not only reviews are supported. 
> - Also changed the naming to target-package-filter

Thanks! Can you also update the commit message?

> - skip action type == 'group'
> 

Some minor remarks below (I'm also fine if you ignore them except
the True/False thing:) )

> @@ -2312,17 +2314,27 @@ Please submit there instead, or use --nodevelproject to force direct submission.
>              ## Workaround: fetch all, and filter on client side.
>  
>              ## FIXME: date filtering should become implemented on server side
> +
>              for result in results:
> -                if days == 0 or result.state.when > since or result.state.name == 'new':
> -                    if (opts.interactive or conf.config['request_show_interactive']) and not opts.non_interactive:
> -                        ignore_reviews = subcmd != 'review'
> -                        request_interactive_review(apiurl, result, group=opts.group,
> -                                                   ignore_reviews=ignore_reviews,
> -                                                   source_buildstatus=source_buildstatus)
> +                filtered = 0

Hmm let's make filtered a boolean:) (filtered = False)

> +                for action in result.actions:
> +                    if action.type == 'group' or not opts.target_package_filter:
> +                        continue
> +                    filter_pattern = re.compile(opts.target_package_filter)

Hmm there's no need to compile the pattern in each iteration (ideally,
we only do this once (e.g., before the "for result in results:" loop)).

> +                    if not (filter_pattern.match(action.tgt_package)):

Superfluous parentheses

> +                        filtered = 1

filtered = True

> +                        break
> +                if filtered == 0:

if filtered:

> +                    if days == 0 or result.state.when > since or result.state.name == 'new' or filtered == 0:
> +                        if (opts.interactive or conf.config['request_show_interactive']) and not opts.non_interactive:
> +                            ignore_reviews = subcmd != 'review'
> +                            request_interactive_review(apiurl, result, group=opts.group,
> +                                                       ignore_reviews=ignore_reviews,
> +                                                       source_buildstatus=source_buildstatus)
> +                        else:
> +                            print(result.list_view(), '\n')
>                      else:
> -                        print(result.list_view(), '\n')
> -                else:
> -                    skipped += 1
> +                        skipped += 1
>              if skipped:
>                  print("There are %d requests older than %s days.\n" % (skipped, days))

Btw., we could also this without moving much of the existing code (and
saving a level of indentation). For instance, something like

filter_pattern = None
if opts.target_package_filter:
    filter_pattern = re.compile(opts.target_package_filter)
for result in results:
    filtered = False
    if filter_pattern is not None:
        # here your new filter code
    if filtered:
        continue
    # here the old existing code
    if days...

lethliel:
> Hmm let's make filtered a boolean:) (filtered = False)
>

Done. Also removed a useless if condition inside the 'not filtered case'. 
>> + if not (filter_pattern.match(action.tgt_package)):
> 
> Superfluous parentheses

Removed. Just a question: Do they hurt? Or is this just a design question?
 
> Btw., we could also this without moving much of the existing code (and
> saving a level of indentation). For instance, something like
> 

I would like to keep it in the way it is. If find it more readable.

marcus-h:
On 2017-08-17 01:23:18 -0700, Marco Strigl wrote:
> > Hmm let's make filtered a boolean:) (filtered = False)
> >
> 
> Done. Also removed a useless if condition inside the 'not filtered case'. 
> >> + if not (filter_pattern.match(action.tgt_package)):
> > 
> > Superfluous parentheses
> 
> Removed. Just a question: Do they hurt? Or is this just a design question?
>  
No. Yes, just for cosmetic reasons:)

> > Btw., we could also this without moving much of the existing code (and
> > saving a level of indentation). For instance, something like
> > 
> 
> I would like to keep it in the way it is. If find it more readable. 
> 
Ok. Merged. Thanks!

PR #318, opened 2017-08-17
Check digest of modified files

bmwiedemann:
this reverts a semantic change introduced by commit
48a35fed91d557fd89eed4297d2516776434d837
that caused files with updated timestamps to always be in "M" state.


marcus-h:
On 2017-08-16 21:32:37 -0700, Bernhard M. Wiedemann wrote:
> this reverts a semantic change introduced by commit
> 48a35fed91d557fd89eed4297d2516776434d837
> that caused files with updated timestamps to always be in "M" state.

Arghs sorry! Merged. Thanks for patch!

PR #320, opened 2017-08-21
renamed get_group to get_group_meta.

lethliel:
renamed get_group to get_group_meta.  

@marcus-h: please review.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -909,7 +909,7 @@ class Osc(cmdln.Cmdln):
                 if r:
                     sys.stdout.write(''.join(r))
             elif cmd == 'group':
-                r = get_group(apiurl, group)
+                r = get_group_meta(apiurl, group)
                 if r:
                     sys.stdout.write(''.join(r))
             elif cmd == 'pattern':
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2017-08-21 07:54:40 +0000, Marco Strigl wrote:
> renamed get_group to get_group_meta.  
> 
Hmm can we keep get_group for backward compatibility, which simply
delegates to get_group_meta? Just in case it is used by some
plugin...

lethliel:
> Hmm can we keep get_group for backward compatibility, which simply
> delegates to get_group_meta? Just in case it is used by some
> plugin...

Of course. I added the function get_group again and just forwarded it to get_group_meta.

marcus-h:
On 2017-08-21 08:39:07 +0000, Marco Strigl wrote:
> > Hmm can we keep get_group for backward compatibility, which simply
> > delegates to get_group_meta? Just in case it is used by some
> > plugin...
> 
> Of course. I added the function get_group again and just forwarded it to get_group_meta.
> 
Thanks! Merged.

PR #324, opened 2017-08-25
show user comments in interactive mode

lethliel:
show comments for the request in interactive mode.


lethliel:
@marcus-h: please review

marcus-h:
On 2017-08-25 10:01:46 +0000, Marco Strigl wrote:
> show comments for the request in interactive mode.

LGTM. Merged. Thanks! (Remark: an alternative would be
to move the print_comments call into print_request.)

PR #325, opened 2017-08-30
added function to get rpmlint log content

lethliel:
Provide function to show the content of rpmlint.log.
The core function get_rpmlint_log will later be used in the interactive
request mode to provide the rpmlint logs to the reviewer.

I add the osc rpmlint | rpmlintlog | lint command also.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -5211,6 +5211,26 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         print("See 'osc help prjresults'.", file=sys.stderr)
         return 2
 
+    @cmdln.alias('rpmlint')
+    @cmdln.alias('lint')
+    def do_rpmlintlog(self, subcmd, opts, *args):
+        """${cmd_name}: Shows the rpmlint logfile
+
+        Shows the rpmlint logfile to analyse if there are any problems
+        with the spec file and the built binaries.
+
+        usage:
+            osc rpmlintlog project package repository arch
+        """
+        apiurl = self.get_api_url()
+
+        if len(args) == 4:
+            project, package, repository, arch = args
+        else:
+            raise oscerr.WrongArgs('please provide project package repository arch.')
+
+        print(get_rpmlint_log(apiurl, project, package, repository, arch))
+
     @cmdln.alias('bl')
     @cmdln.alias('blt')
     @cmdln.alias('buildlogtail')
--- a/osc/core.py
+++ b/osc/core.py
@@ -7548,4 +7548,9 @@ def delete_comment(apiurl, cid):
     ret = ET.fromstring(f.read()).find('summary')
     return ret.text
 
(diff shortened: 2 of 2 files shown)


marcus-h:
On 2017-08-30 15:13:08 +0000, Marco Strigl wrote:
> The core function get_rpmlint_log will later be used in the interactive
> request mode to provide the rpmlint logs to the reviewer.
> 

> +        Shows the rpmlint logfile to analyse if there are any problems
> +        with the spec file.

It also checks the built binaries.

> +def get_rpmlint_log(apiurl, proj, pkg, repo, arch):
> +    u = makeurl(apiurl, ['build', proj, repo, arch, pkg, 'rpmlint.log'])
> +    try:
> +        f = http_GET(u)
> +        return f.read()
> +    except HTTPError as e:
> +        if e.code == 404:
> +            return('no rpmlint.log found')
> +        else:
> +            return('Error in http request: ' + str(e.code))

Hmm on the one hand, this error handling for a "core" function is a
bit problematic, because potential callers have no "real" clue what
went wrong in case of an error. On the other hand it comes handy for
the use in the interactive review code path...
What about the following:
- no exception handling in get_rpmlint_log
- later, add a "safe_get_rpmlint_log" function in
  "request_interactive_review" (cf. "safe_change_request_state")

(btw., useless parentheses around the return statements;) )

The rest: LGTM

lethliel:
> 
> It also checks the built binaries.
>

You are right. Added to description
 
> What about the following:
> - no exception handling in get_rpmlint_log
> - later, add a "safe_get_rpmlint_log" function in
> "request_interactive_review" (cf. "safe_change_request_state")
>

I removed the error handling from the core function. I just thouht that
it would be a better message for the user that there is no rpmlint.log instead
of just returning a 404. 

But as you said I will implement a safe_get_rpmlint_log function to handle
the 404 if called from request_interactive_review. 

I think the command 'osc rpmlintlog' is not issued directly by anyone.

> (btw., useless parentheses around the return statements;) )
> 

Those sneaky parentheses

marcus-h:
On 2017-09-01 08:42:45 +0000, Marco Strigl wrote:
> > What about the following:
> > - no exception handling in get_rpmlint_log
> > - later, add a "safe_get_rpmlint_log" function in
> > "request_interactive_review" (cf. "safe_change_request_state")
> >
> 
> I removed the error handling from the core function. I just thouht that
> it would be a better message for the user that there is no rpmlint.log instead
> of just returning a 404. 
> 
That's what the 404 basically says:)

> But as you said I will implement a safe_get_rpmlint_log function to handle
> the 404 if called from request_interactive_review. 
> 
> I think the command 'osc rpmlintlog' is not issued directly by anyone.
> 
Why did we add it then?:) I think it could be useful...

Btw., I just did a small follow-up commit b07ba53 (just a bit "usability").

Thanks for the PR!

PR #326, opened 2017-08-31
osc tr: show the time when the event happened

DimStar77:
It can be handy to have this information at hand - otherwise one needs an extra query on jobhist to find it out (preferably I'd also like to see 'who' caused the event, but that info is not stored at the moment)


marcus-h:
Merged. Thanks a lot!

PR #327, opened 2017-08-31
submitreq entire project fix

jnweiger:
`osc submitreq TARGETPROJ`
bails out with unbound variable t. This is true. The variable should be target_project.
Am I the first user to try that feature? :-)


marcus-h:
On 2017-08-31 05:59:52 -0700, Jürgen Weigert wrote:
> `osc submitreq TARGETPROJ`
> bails out with unbound variable t. This is true. The variable should be target_project.

Merged. Thanks for the PR!

> Am I the first user to try that feature? :-)

Hehe:)

PR #331, opened 2017-09-04
show lintlog (li) in interactive mode

lethliel:
Show rpmlintlog in interactive mode.
Gives the user a list with possible repo/arch combinations he can choose from.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -7126,6 +7126,40 @@ def request_interactive_review(apiurl, request, initial_cmd='', group=None,
             print('Try -f to force the state change', file=sys.stderr)
         return False
 
+    def safe_get_rpmlint_log(src_actions):
+        lintlogs = []
+        for action in src_actions:
+            print('Type %s:' % action.type)
+            disabled = show_package_disabled_repos(apiurl, action.src_project, action.src_package)
+            for repo in get_repos_of_project(apiurl, action.src_project):
+                if disabled is None or repo.name not in disabled:
+                    lintlog_entry = {
+                                      'proj': action.src_project,
+                                      'pkg': action.src_package,
+                                      'repo': repo.name,
+                                      'arch': repo.arch
+                                      }
+                    lintlogs.append(lintlog_entry)
+                    print('(%i) %s/%s/%s/%s' % ((len(lintlogs)-1), action.src_project, action.src_package, repo.name, repo.arch))
+        if not lintlogs:
+            print('No possible rpmlintlogs found')
+            return False
+        while True:
+            try:
+                lint_n = int(raw_input('Number of rpmlint log to examine (0 - %i): ' % (len(lintlogs)-1)))
+                lintlogs[lint_n]
+                break
+            except (ValueError, IndexError):
+                print('Invalid rpmlintlog index. Please choose between 0 and %i' % (len(lintlogs)-1))
+        try:
+            print(get_rpmlint_log(apiurl, **lintlogs[lint_n]))
+        except HTTPError as e:
+            if e.code == 404:
+                print('No rpmlintlog for %s %s' % (lintlogs[lint_n]['repo'],
+                      lintlogs[lint_n]['arch']))
+            else:
+                raise e
+
     def print_request(request):
         print(request)
 
@@ -7149,10 +7183,10 @@ def request_interactive_review(apiurl, request, initial_cmd='', group=None,
         # actions which have sources + buildresults
         src_actions = editable_actions + request.get_actions('maintenance_release')
         if editable_actions:
-            prompt = 'd(i)ff/(a)ccept/(d)ecline/(r)evoke/(b)uildstatus/c(l)one/(e)dit/co(m)ment/(s)kip/(c)ancel > '
+            prompt = 'd(i)ff/(a)ccept/(d)ecline/(r)evoke/(b)uildstatus/rpm(li)ntlog/c(l)one/(e)dit/co(m)ment/(s)kip/(c)ancel > '
         elif src_actions:
             # no edit for maintenance release requests
-            prompt = 'd(i)ff/(a)ccept/(d)ecline/(r)evoke/(b)uildstatus/c(l)one/co(m)ment/(s)kip/(c)ancel > '
+            prompt = 'd(i)ff/(a)ccept/(d)ecline/(r)evoke/(b)uildstatus/rpm(li)ntlog/c(l)one/co(m)ment/(s)kip/(c)ancel > '
         editprj = ''
         orequest = None
         if source_buildstatus and src_actions:
@@ -7197,6 +7231,8 @@ def request_interactive_review(apiurl, request, initial_cmd='', group=None,
                 create_comment(apiurl, 'request', comment, request.reqid)
             elif repl == 'b' and src_actions:
                 print_source_buildstatus(src_actions)
+            elif repl =='li' and src_actions:
+                safe_get_rpmlint_log(src_actions)
             elif repl == 'e' and editable_actions:
                 # this is only for editable actions
                 if not editprj:


lethliel:
@marcus-h: 
My solution is just an idea. I am open to your thoughts.

marcus-h:
On 2017-09-04 15:33:31 +0000, Marco Strigl wrote:
> Show rpmlintlog in interactive mode.
> Gives the user a list with possible repo/arch combinations he can choose from.
> 
> +    def safe_get_rpmlint_log(src_actions):
> +        for action in src_actions:
> +            lintlogs = {}
> +            count = 0
> +            disabled = show_package_disabled_repos(apiurl, action.src_project, action.src_package)
> +            for repo in get_repos_of_project(apiurl, action.src_project):
> +                if (disabled is None) or ((disabled is not None) and (repo.name not in disabled)):

Could be simplified:
if disabled is None or repo.name not in disabled:
    ...

> +                    count = count + 1
> +                    lintlogs[count] = {}
> +                    lintlogs[count]['proj'] = action.src_project
> +                    lintlogs[count]['pkg'] = action.src_package
> +                    lintlogs[count]['repo'] = repo.name
> +                    lintlogs[count]['arch'] = repo.arch
> +                    print('(%i) %s/%s/%s/%s' % (count, action.src_project, action.src_package, repo.name, repo.arch))
> +            lint_n = int(raw_input(('Number of rpmlint log to examine (1 - %i): ' % count)))

Hmm maybe catch a ValueError here?

> +            try:
> +                print(get_rpmlint_log(apiurl,
> +                      lintlogs[lint_n]['proj'],
> +                      lintlogs[lint_n]['pkg'],
> +                      lintlogs[lint_n]['repo'],
> +                      lintlogs[lint_n]['arch']))

Could be shortened to print(get_rpmlint_log(apiurl, **lintlogs[lint_n])),
but the current call is, of course, also ok...

> +            except HTTPError as e:
> +                print('No rpmlintlog for %s %s' % (lintlogs[lint_n]['repo'],
> +                      lintlogs[lint_n]['arch']))

Hmm the message makes only sense if e.code == 404.

Maybe we should also catch an IndexError exception or check if
lint_n is within the bounds?


The "behavior" of safe_get_rpmlint_log looks a bit odd to me: the
user has to make a choice for each src_action. Maybe it is more
appropriate to list all possible choices and let the user select
exactly _one_? What do you think?

> @@ -7149,10 +7174,10 @@ def print_source_buildstatus(src_actions, newline=False):
>          # actions which have sources + buildresults
>          src_actions = editable_actions + request.get_actions('maintenance_release')
>          if editable_actions:
> -            prompt = 'd(i)ff/(a)ccept/(d)ecline/(r)evoke/(b)uildstatus/c(l)one/(e)dit/co(m)ment/(s)kip/(c)ancel > '
> +            prompt = 'd(i)ff/(a)ccept/(d)ecline/(r)evoke/(b)uildstatus/(li)ntlog/c(l)one/(e)dit/co(m)ment/(s)kip/(c)ancel > '

Maybe call it rpm(li)ntlog?

lethliel:
> On 09/04/2017 08:01 PM, Marcus Hüwe  wrote:
> 
> Could be simplified:
> if disabled is None or repo.name not in disabled:
> ...

Ok.

> 
>> + lint_n = int(raw_input(('Number of rpmlint log to examine (1 - %i): ' % count)))
> 
> Hmm maybe catch a ValueError here?
> 

Done.
 
> Could be shortened to print(get_rpmlint_log(apiurl, **lintlogs[lint_n])),
> but the current call is, of course, also ok...
>

I forget about this option every time ;)
 
>> + except HTTPError as e:
>> + print('No rpmlintlog for %s %s' % (lintlogs[lint_n]['repo'],
>> + lintlogs[lint_n]['arch']))
> 
> Hmm the message makes only sense if e.code == 404.
> 
> Maybe we should also catch an IndexError exception or check if
> lint_n is within the bounds?
> 

Not needed. The check if lint_n is within valid bounds is now done 
right after the lint_n = raw_input(....)

> 
> The "behavior" of safe_get_rpmlint_log looks a bit odd to me: the
> user has to make a choice for each src_action. Maybe it is more
> appropriate to list all possible choices and let the user select
> exactly _one_? What do you think?
>

Rearranged a little bit. Now all src_actions.types are shown at once.

 
>> @@ -7149,10 +7174,10 @@ def print_source_buildstatus(src_actions, newline=False):
>> # actions which have sources + buildresults
>> src_actions = editable_actions + request.get_actions('maintenance_release')
>> if editable_actions:
>> - prompt = 'd(i)ff/(a)ccept/(d)ecline/(r)evoke/(b)uildstatus/c(l)one/(e)dit/co(m)ment/(s)kip/(c)ancel > '
>> + prompt = 'd(i)ff/(a)ccept/(d)ecline/(r)evoke/(b)uildstatus/(li)ntlog/c(l)one/(e)dit/co(m)ment/(s)kip/(c)ancel > '
> 
> Maybe call it rpm(li)ntlog?
>

Yes. I think at some point we have to think of a new approach to the interactive mode.

lethliel:
@marcus-h: can you please have a look again?

marcus-h:
On 2017-09-25 03:42:32 -0700, Marco Strigl wrote:
> @marcus-h: can you please have a look again? 
> 
Yep:)

> +    def safe_get_rpmlint_log(src_actions):
> +        count = 0
> +        for action in src_actions:
> +            lintlogs = {}
> +            print('Type %s:' % action.type)
> +            disabled = show_package_disabled_repos(apiurl, action.src_project, action.src_package)
> +            for repo in get_repos_of_project(apiurl, action.src_project):
> +                if disabled is None or repo.name not in disabled:
> +                    count = count + 1
> +                    lintlogs[count] = {}
> +                    lintlogs[count]['proj'] = action.src_project
> +                    lintlogs[count]['pkg'] = action.src_package
> +                    lintlogs[count]['repo'] = repo.name
> +                    lintlogs[count]['arch'] = repo.arch
> +                    print('(%i) %s/%s/%s/%s' % (count, action.src_project, action.src_package, repo.name, repo.arch))

The lintlogs dict will only contain the data of the last iteration
and, hence, the subsequent code breaks when using a "key" of a
prior iteration. Instead of a dict, I would simply use a list.

> +        lint_n = int(raw_input(('Number of rpmlint log to examine (1 - %i): ' % count)))

Hmm maybe we could/should use a range between 0 and count - 1 (this would
be consistent with the "choose review code"). But that's a minor detail...

Superfluous parentheses.

> +        while lint_n < 1 or lint_n > count:
> +            print('Invalid rpmlintlog index. Please choose between 1 and %i' % count)
> +            lint_n = int(raw_input(('Number of rpmlint log to examine (1 - %i): ' % count)))

I think we should handle a ValueError exception.

Superfluous parentheses.

lethliel:
> The lintlogs dict will only contain the data of the last iteration
> and, hence, the subsequent code breaks when using a "key" of a
> prior iteration. Instead of a dict, I would simply use a list.
> 

I switched to a list and also got rid of the counter

>> + lint_n = int(raw_input(('Number of rpmlint log to examine (1 - %i): ' % count)))
> 
> Hmm maybe we could/should use a range between 0 and count - 1 (this would
> be consistent with the "choose review code"). But that's a minor detail...
> 

is now from 0 to (len(lintlogs) - 1)

> Superfluous parentheses.

My friends :) 

> 
>> + while lint_n < 1 or lint_n > count:
>> + print('Invalid rpmlintlog index. Please choose between 1 and %i' % count)
>> + lint_n = int(raw_input(('Number of rpmlint log to examine (1 - %i): ' % count)))
> 
> I think we should handle a ValueError exception.
>

I know handle a ValueError for anything else entered than an integer and an 
IndexError for not existing index entered.

 
> Superfluous parentheses.

Again my friends.

marcus-h:
On 2017-09-26 06:32:59 -0700, Marco Strigl wrote:
> > The lintlogs dict will only contain the data of the last iteration
> > and, hence, the subsequent code breaks when using a "key" of a
> > prior iteration. Instead of a dict, I would simply use a list.
> > 
> 
> I switched to a list and also got rid of the counter
> 
Thanks! For the second part: you got almost rid of "counter";) (there
are still 2 (useless) occurrences)

> >> + lint_n = int(raw_input(('Number of rpmlint log to examine (1 - %i): ' % count)))
> > 
> > Hmm maybe we could/should use a range between 0 and count - 1 (this would
> > be consistent with the "choose review code"). But that's a minor detail...
> > 
> 
> is now from 0 to (len(lintlogs) - 1)
> 
Thanks.

> >> + while lint_n < 1 or lint_n > count:
> >> + print('Invalid rpmlintlog index. Please choose between 1 and %i' % count)
> >> + lint_n = int(raw_input(('Number of rpmlint log to examine (1 - %i): ' % count)))
> > 
> > I think we should handle a ValueError exception.
> >
> 
> I know handle a ValueError for anything else entered than an integer and an 
> IndexError for not existing index entered.
> 
Actually, your initial solution, which did the explicit bounds check was
a bit nicer, but ok...

> +        while True:
> +            try:
> +                lint_n = int(raw_input('Number of rpmlint log to examine (0 - %i): ' % (len(lintlogs)-1)))
> +                t=lintlogs[lint_n]

Hmm "t" is not needed.

> +                break
> +            except ValueError:
> +                print('Invalid rpmlintlog index. Please choose between 0 and %i' % (len(lintlogs)-1))
> +            except IndexError:
> +                print('Invalid rpmlintlog index. Please choose between 0 and %i' % (len(lintlogs)-1))

Could be merged:

except (ValueError, IndexError):
    print(...)

but the current code is also ok.

Once, the "unused" variables (counter, t) are removed, we are ready
to merge:)

Btw., maybe we should also take an empty "lintlogs" list into account
(e.g., a check before the while loop).

lethliel:
@marcus-h 

- removed useless vars t and count
- check if lintlog is empty before while loop

marcus-h:
On 2017-09-26 18:18:11 +0000, Marco Strigl wrote:
> @marcus-h 
> 
> - removed useless vars t and count
> - check if lintlog is empty before while loop
> 
Thanks! (Actually, a simple "return" instead of "return False"
would have been sufficient. However, I merged it nonetheless)

PR #335, opened 2017-09-18
Add pacsuffix hint using BinaryType from build config

### LINKED ISSUE #321: osc always says 100% cache miss when building debian packages with spec files
Whenever doing `osc build` for building Debian packages that use a spec file as a source, it always has a 100% cache miss and redownloads everything unnecessarily.

For example, when building `cscript` for Debian 9.0 from `home:Pharaoh_Atem:cscript`, it always returns this message:
```
100.0% cache miss. 0/153 dependencies cached.
```

It can be reproduced by doing the following:
```
$ osc co home:Pharaoh_Atem:cscript
$ cd home:Pharaoh_Atem:cscript/cscript
$ osc build Debian_9.0 x86_64
....
$ osc build Debian_9.0 x86_64
```
Each run will always have a 100% cache miss.

I believe it has to do with [how package suffixes](https://github.com/openSUSE/osc/blob/master/osc/build.py#L124-L134) are determined, but I'm not sure how this can be fixed properly.

Conan-Kudo:
This adds a hint for package suffix using the BinaryType from the build config

This fixes #321.

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -21,7 +21,7 @@ except ImportError:
 from tempfile import NamedTemporaryFile, mkdtemp
 from osc.fetch import *
 from osc.core import get_buildinfo, store_read_apiurl, store_read_project, store_read_package, meta_exists, quote_plus, get_buildconfig, is_package_dir, dgst
-from osc.core import get_binarylist, get_binary_file, run_external, raw_input
+from osc.core import get_binarylist, get_binary_file, run_external, return_external, raw_input
 from osc.util import rpmquery, debquery, archquery
 import osc.conf
 from . import oscerr
@@ -90,7 +90,7 @@ if hostarch == 'parisc':
 class Buildinfo:
     """represent the contents of a buildinfo file"""
 
-    def __init__(self, filename, apiurl, buildtype = 'spec', localpkgs = []):
+    def __init__(self, filename, apiurl, buildtype = 'spec', localpkgs = [], binarytype = 'rpm'):
         try:
             tree = ET.parse(filename)
         except:
@@ -117,12 +117,13 @@ class Buildinfo:
             raise URLError('invalid protocol for the apiurl: \'%s\'' % apiurl)
 
         self.buildtype = buildtype
+        self.binarytype = binarytype
         self.apiurl = apiurl
 
         # are we building .rpm or .deb?
         # XXX: shouldn't we deliver the type via the buildinfo?
         self.pacsuffix = 'rpm'
-        if self.buildtype == 'dsc' or self.buildtype == 'collax':
+        if self.buildtype == 'dsc' or self.buildtype == 'collax' or self.binarytype == 'deb':
             self.pacsuffix = 'deb'
         if self.buildtype == 'arch':
             self.pacsuffix = 'arch'
@@ -811,7 +812,15 @@ def main(apiurl, opts, argv):
         else:
             raise
 
-    bi = Buildinfo(bi_filename, apiurl, build_type, list(prefer_pkgs.keys()))
+    # Set default binary type if cannot be detected
+    binary_type = 'rpm'
+    if os.path.exists('/usr/lib/build/queryconfig'):
+        binary_type = return_external('/usr/lib/build/queryconfig', '--dist', bc_filename, 'binarytype').decode('utf-8').strip()
+    # If binary type is set to a useless value, reset to 'rpm'
+    if binary_type == 'UNDEFINED':
+        binary_type = 'rpm'
+
+    bi = Buildinfo(bi_filename, apiurl, build_type, list(prefer_pkgs.keys()), binary_type)
 
     if bi.debuginfo and not (opts.disable_debuginfo or '--debug' in buildargs):
         buildargs.append('--debug')


Conan-Kudo:
Example output from my computer:

```
$ osc build Debian_9.0 x86_64
...
Updating cache of required packages

0.7% cache miss. 152/153 dependencies cached.

(Debian:debbuild) debbuild_17.8.1-1.2_all.deb                                     100% |========================================|  38 kB    00:00
```

marcus-h:
On 2017-09-18 22:09:12 +0000, Neal Gompa (ニール・ゴンパ) wrote:
> As there's no property in buildinfo to indicate the binary type, and binarytype isn't always defined in buildconfig, use the Repotype as a hint to determine the correct pacsuffix. This should help make osc determine the right suffix for things like package caching and searching for packages built at the end.
> 
Your second version, which uses queryconfig + binarytype, looks good to
me. Minor remarks:

> +    if os.path.exists('/usr/lib/build/queryconfig'):
> +        binary_type = subprocess.check_output('/usr/lib/build/queryconfig --dist %s binarytype' % bc_filename, shell=True).decode('utf-8').strip()

- shell=True is not needed
- I don't see why you need to perform utf8 decoding
- instead of subprocess.check_output use core.return_external

Actually, we should also "cleanup" the Buildinfo class so that
it only relies on the binarytype... I'll take care of it, once
this is merged.

Thanks for the patch!

Conan-Kudo:
> - shell=True is not needed
> - I don't see why you need to perform utf8 decoding
> - instead of subprocess.check_output use core.return_external

I tried to get it to work with `shell=False`, but it kept saying "Missing distribution!". It wouldn't work until I switched to it being executed in the shell. :(

Conan-Kudo:
> Actually, we should also "cleanup" the Buildinfo class so that it only relies on the binarytype... I'll take care of it, once this is merged.

You might want to be careful about changing that. Test to be sure that things like Arch and snapcraft other targets will work correctly that way. I did not rewrite it because I observed weirdness with it, and only wanted to use it as a hint.

marcus-h:
On 2017-09-22 03:13:11 -0700, Neal Gompa (ニール・ゴンパ) wrote:
> > - shell=True is not needed
> > - I don't see why you need to perform utf8 decoding
> > - instead of subprocess.check_output use core.return_external
> 
> I tried to get it to work with `shell=False`, but it kept saying "Missing distribution!". It wouldn't work until I switched to it being executed in the shell. :(

Yes, you need to pass a list instead of a single string to
subprocess.check_output. If you use core.return_external
something like this should do the trick

return_external('/usr/lib/build/queryconfig', '--dist', bc_filename,
                'binarytype')

(Hmm hardcoding the path is not a good idea, though... but we can
have a look at it later (since it does not break anything))

Conan-Kudo:
@marcus-h I tried it with a list, it did not work. I only switched to a string and shell after spending a couple of hours trying to debug it.

marcus-h:
On 2017-09-22 09:00:41 -0700, Neal Gompa (ニール・ゴンパ) wrote:
> @marcus-h I tried it with a list, it did not work. I only switched to a string and shell after spending a couple of hours trying to debug it.
> 
Hmm works for me

subprocess.check_output(['/usr/lib/build/queryconfig', '--dist',
                        './_buildconfig-openSUSE_Factory-x86_64',
                        'binarytype'])

What error do you get?

Conan-Kudo:
@marcus-h Ah, I screwed up...

I did the following:
```python
subprocess.check_output(['/usr/lib/build/queryconfig', '--dist %s' %
                        './_buildconfig-openSUSE_Factory-x86_64',
                        'binarytype'])
```
which doesn't work.

Conan-Kudo:
@marcus-h Now it should work and I changed it accordingly.

marcus-h:
On 2017-09-23 12:05:36 +0000, Neal Gompa (ニール・ゴンパ) wrote:
> @marcus-h Now it should work and I changed it accordingly.
> 

> +    if os.path.exists('/usr/lib/build/queryconfig'):
> +        binary_type = subprocess.check_output(['/usr/lib/build/queryconfig', '--dist', bc_filename, 'binarytype']).decode('utf-8').strip()

Please use core.return_external instead of subprocess.check_output.
Also, I don't think the decode('utf-8') part is needed.

Conan-Kudo:
@marcus-h I did it for Python 3 compatibility. `return_external` doesn't convert it back to a string type (it remains a byte array).

marcus-h:
On 2017-09-25 04:07:08 -0700, Neal Gompa (ニール・ゴンパ) wrote:
> @marcus-h I did it for Python 3 compatibility. `return_external` doesn't convert it back to a string type (it remains a byte array).
> 
Ok (though there are several other places where we do not decode the
subprocess.check_output results...).
One last remark: can you please add some fallback code for the UNDEFINED
case? Something like

if binary_type == 'UNDEFINED':
    binary_type = 'rpm'

Afterwards, I'll happily merge it:)
Thanks again!

Conan-Kudo:
@marcus-h I figured fixing the Python 3 compatibility stuff is out of scope for this particular PR, but I may go through and make adjustments at a later time.

I added the necessary fallback code, though.

marcus-h:
On 2017-09-26 04:59:58 -0700, Neal Gompa (ニール・ゴンパ) wrote:
> @marcus-h I figured fixing the Python 3 compatibility stuff is out of scope for this particular PR, but I may go through and make adjustments at a later time.
> 
Yep of course (it was just meant as a general remark, but, of course,
patches are always welcome;) ).

> I added the necessary fallback code, though.
> 
Merged. Thanks a lot!

PR #337, opened 2017-09-22
check for incoming requests before commit

matejcik:
osc will now notify you that you have pending requests to a package that you are trying to commit. That way you won't accidentally mess up someone's submit request.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -4615,6 +4615,19 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             print("or use the --noservice option")
             return 1
 
+    def _has_pending_requests(self, api, prj, pkg, force=False):
+        if force:
+            return False
+
+        rqlist = get_request_list(api, prj, pkg)
+        if rqlist:
+            print("The following requests are pending for {}/{}:".format(prj, pkg))
+            for rq in rqlist:
+                print("\t{}\t by {}".format(rq.reqid, rq.creator))
+            print("Please resolve them before commiting. Use --force to commit anyway.")
+            return True
+        else:
+            return False
 
     def _commit(self, subcmd, opts, args):
         args = parseargs(args)
@@ -4643,8 +4656,15 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                     msg = edit_message()
 
                 # check any of the packages is a link, if so, as for branching
-                pacs = (Package(os.path.join(prj.dir, pac))
-                        for pac in prj.pacs_have if prj.get_state(pac) == ' ')
+                pacs = [Package(os.path.join(prj.dir, pac))
+                        for pac in prj.pacs_have if prj.get_state(pac) == ' ']
+
+                for p in pacs[:]:
+                    if self._has_pending_requests(p.apiurl, p.prjname, p.name, force=opts.force):
+                        pacs.remove(p)
+                if not pacs:
+                    continue
+
                 can_branch = False
                 if any(pac.is_link_to_different_project() for pac in pacs):
                     repl = raw_input('Some of the packages are links to a different project!\n' \
@@ -4697,6 +4717,11 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                     store_read_apiurl(pac, defaulturl=False)
             for prj_path, packages in prj_paths.items():
                 prj = Project(prj_path)
+                for p in packages[:]:
+                    if self._has_pending_requests(prj.apiurl, prj.name, p, force=opts.force):
+                        packages.remove(p)
+                if not packages:
+                    continue
                 if not msg and not opts.no_message:
                     msg = get_commit_msg(prj.absdir, pac_objs[prj_path])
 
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2017-09-22 06:34:52 -0700, matejcik wrote:
> osc will now notify you that you have pending requests to a package that you are trying to commit. That way you won't accidentally mess up someone's submit request.

Hmm this change is quite "intrusive", because it stops committing
packages in case of existing requests. If we want to have such a
feature, we should introduce a config option to enable it explicitly
(IMHO). Actually, we already have something that is close to your
feature:
If you set the check_for_request_on_action config option, for instance
via "osc config check_for_request_on_action 1", osc will print
pending requests after each update/commit etc. command. Would this
be sufficient for you?

One remark about the implementation:
- it will not work, if you issue the "osc commit" from within a
  project working copy (cd project; osc ci)

matejcik:
I'll let @scarabeusiv argue this one, but IIUC the intrusiveness is the point: we want to prevent people (incl. us) from overwriting incoming requests implicitly. That's why even if this is configurable, it should be enabled by default.

and also no, if the request doesn't interrupt the commit action (that breaks the request usually), then it is of no use

re "osc commit" from project, did i miss a branch? or is it in a different command?
(as written, i thought the code would skip all packages with pending requests and commit the rest)

marcus-h:
On 2017-09-26 06:33:42 -0700, matejcik wrote:
> I'll let @scarabeusiv argue this one, but IIUC the intrusiveness is the point: we want to prevent people (incl. us) from overwriting incoming requests implicitly. That's why even if this is configurable, it should be enabled by default.
> 
Ah ok. Who is "we"?:) Review team? Some other team? Or just you
two?:) (This question is not meant "offensive")

> and also no, if the request doesn't interrupt the commit action (that breaks the request usually), then it is of no use
> 
Ok. IMHO, if we need to change the default semantics in such a way,
we should at least bring this up on some mailinglist (e.g.,
buildservice, factory, or packaging) and ask for feedback/remarks
etc. What do you think?

Another question is, if this makes sense for all projects or is more
some kind of policy that applies only to, for instance, openSUSE:Factory?
Maybe this is something that can be implemented in the api (e.g. via some
attribute?)? Then, we (hopefully;) ) get a consistent behavior in the
webui and osc "for free".

> re "osc commit" from project, did i miss a branch? or is it in a different command?
> (as written, i thought the code would skip all packages with pending requests and commit the rest)
> 
See the first for-loop in _commit: if "prj.commit(...)" is reached, all
packages will be committed (regardless of the contents of "pacs"). But
this can be easily fixed... (Project.commit has a "pacs" parameter that
takes, e.g., a list of package names).

scarabeusiv:
By we in this instance I mean suse packaging team. With sle15 preps we would start commit more and thus step on contributor fingers, this is nice prevention of situations where someone provided something on his free time and us overriding them. It does apply for all develprj and factory at minimal, but generaly the default forcing you to review incoming queue first is not such a bad notion.

osc/commandline.py
```diff
@@ -4615,6 +4615,19 @@ def do_commit(self, subcmd, opts, *args):
             print("or use the --noservice option")
             return 1
 
+    def _has_pending_requests(self, api, prj, pkg, force=False):
+        if force:
+            return False
+
+        rqlist = get_request_list(api, prj, pkg)
```
andreasstieger: This should be limited to certain types of request, e.g. new, review.

osc/commandline.py
```diff
@@ -4615,6 +4615,19 @@ def do_commit(self, subcmd, opts, *args):
...
+        if rqlist:
+            print("The following requests are pending for {}/{}:".format(prj, pkg))
```
andreasstieger: Last time I did this this returned incoming as well as outgoing requests. In this case this may trigger on a request that is still pending for factory.

osc/commandline.py
matejcik: That's the default: new, review, declined. I suppose "declined" doesn't need to be there.

osc/commandline.py
matejcik: hm, this seems to depend on a _config value_.
uggghhhhhhhhh  why /bangs head against keyboard/

osc/commandline.py
andreasstieger: Yes, otherwise you will depend on a third party to revoke declined requests.

marcus-h:
On 2017-10-11 13:09:16 +0000, matejcik wrote:
> matejcik commented on this pull request.
> > @@ -4615,6 +4615,19 @@ def do_commit(self, subcmd, opts, *args):
>              print("or use the --noservice option")
>              return 1
>  
> +    def _has_pending_requests(self, api, prj, pkg, force=False):
> +        if force:
> +            return False
> +
> +        rqlist = get_request_list(api, prj, pkg)
> +        if rqlist:
> +            print("The following requests are pending for {}/{}:".format(prj, pkg))
> 
> hm, this seems to depend on a _config value_.
> uggghhhhhhhhh  why /bangs head against keyboard/
> 
Yes, quite inelegant, but we can easily fix it.

olafhering:
Having a lock on a pkg if there is pending work is probably not a bad idea.

marcus-h:
On 2017-10-12 12:55:36 +0000, matejcik wrote:
> @matejcik pushed 4 commits.
> 
> 5df98dd  get correct request list
> 2998246  better way to say "filter"
> f4c7265  actually use the filtered list
> f78b7e7  Merge branch 'prevent-commit'
> 
I just had a quick glance at your changes:

> index daed2f1b..a2e8d2ef 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -4670,7 +4670,7 @@ def _commit(self, subcmd, opts, args):
>                      if repl in('y', 'Y'):
>                          can_branch = True
>  
> -                prj.commit(msg=msg, skip_local_service_run=skip_local_service_run, verbose=opts.verbose, can_branch=can_branch)
> +                prj.commit(pacs, msg=msg, skip_local_service_run=skip_local_service_run, verbose=opts.verbose, can_branch=can_branch)
>                  args.remove(arg)

This (PATCH4) still "breaks" "osc ci" in a project wc, because
it only considers packages with state ' '. That is, it is does
not consider packages with state 'A' (new packages) or state 'D'
(packages that should be deleted). Probably, we should also
check for pending requests in case of state 'D'...

From a conceptual point of view, I still think that this
"no commit in case of requests/reviews" feature/policy should
be implemented in the API rather than osc.
@openSUSE/open-build-service any opinions?

mcepl:
I would vote for what @marcus-h wrote
> If you set the check_for_request_on_action config option, for instance via "osc config check_for_request_on_action 1", osc will print pending requests after each update/commit etc. command. Would this be sufficient for you?

and reject this completely.

PR #347, opened 2017-10-18
fixes broken osc results on project level

lethliel:
osc results on project level returns an empty line. 
This bug was introduced with https://github.com/openSUSE/osc/commit/2d327df4e78427ae9588490346c536ed21667ce0

Since then arch and repo are not strings anymore but lists. 
So `arch != None` and `repo != None` were always true even if the lists were empty.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -5725,9 +5725,9 @@ def get_prj_results(apiurl, prj, hide_legend=False, csv=False, status_filter=Non
     pacs = sorted(list(set(pacs)))
     for node in root.findall('result'):
         # filter architecture and repository
-        if arch != None and node.get('arch') not in arch:
+        if arch and node.get('arch') not in arch:
             continue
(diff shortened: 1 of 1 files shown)


lethliel:
@marcus-h: please review

marcus-h:
On 2017-10-18 06:24:40 -0700, Marco Strigl wrote:
> osc results on project level returns an empty line. 
> This bug was introduced with https://github.com/openSUSE/osc/commit/2d327df4e78427ae9588490346c536ed21667ce0
> 
> Since then arch and repo are not strings anymore but lists. 
> So `arch != None` and `repo != None` were always true even if the lists were empty.
> 
Merged. Thanks a lot!

(Btw., it would be nice to have such a detailed description in
the commit message as well:) )

lethliel:
> (Btw., it would be nice to have such a detailed description in
> the commit message as well:) )
> 

Yes. Will do :)

PR #348, opened 2017-10-23
debquery: support control.tar.xz

suihkulokki:
Similar to recent fixes in libsolv and obs-build. Since tarfile
on python2 doesn't do lzma, decompress the file into memory and
feed it as a fake file via StringIO to tarfile

### DIFF
--- a/osc/util/debquery.py
+++ b/osc/util/debquery.py
@@ -5,7 +5,14 @@ from . import ar
 import os.path
 import re
 import tarfile
+import StringIO
 from . import packagequery
+
+HAVE_LZMA = True
+try:
+    import lzma
+except ImportError:
+    HAVE_LZMA = False
 
 class DebError(packagequery.PackageError):
     pass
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2017-10-23 03:55:45 -0700, Riku Voipio wrote:
> Similar to recent fixes in libsolv and obs-build. Since tarfile
> on python2 doesn't do lzma, decompress the file into memory and
> feed it as a fake file via StringIO to tarfile
> 
The patch itself looks good to me! The only "problem" is that it
introduces a hard dependency to the non-standard lzma module.
Let's make it a soft/optional dependency, for instance, something
like

HAVE_LZMA = True
try:
    import lzma
except ImportError:
    HAVE_LZMA = False

...

On packaging level (spec, dsc etc.), we can recommend the lzma
module, though.

What do you think?

suihkulokki:
I can add the checks. But parsing debian.tar.xz is going to be mandatory for debian 10, and users can bump into them accidentally earlier (people using unstable/testing to build packages for older releases).

marcus-h:
On 2017-10-23 13:33:09 +0000, Riku Voipio wrote:
> I can add the checks. But parsing debian.tar.xz is going to be mandatory for debian 10, and users can bump into them accidentally earlier (people using unstable/testing to build packages for older releases).
> 
IMHO, this is ok, if we fail with an appropriate error. Adding
a hard dependency means that we need to make sure that the lzma
module is available on all currently supported distros (or we
have to manually provide an lzma package) (for instance,
SLE-12 does not ship the lzma module). Hence, for now, I would
prefer to make this a "soft" dependency.

suihkulokki:
Updated with the HAVE_LZMA change

marcus-h:
On 2017-10-30 12:37:05 +0000, Riku Voipio wrote:
> Updated with the HAVE_LZMA change
> 
Merged. Thanks a lot for the patch!

Issue #313, opened 2017-08-01
oscrc should be stored in $XDG_CONFIG_HOME on linux


Thaodan:
like the title says the config file shouldn't be stored to ~/. on linux.

PR #349, opened 2017-10-27
use XDG_CONFIG_HOME/osc/oscrc as default config

lethliel:
write oscrc to the default location for user-specific configuration.
If `XDG_CONFIG_HOME` is not set use `~/.config/osc/oscrc` which is basically the same.

If there is already a `~/.oscrc` use this one (for compat reasons). Existing user
installations should not get affected by this commit.

The order is the following:

Given config with `-c`
config defined in `OSC_CONFIG`
existing `~/.oscrc`
default `XDG_CONFIG_HOME/osc/oscrc`

is for https://github.com/openSUSE/osc/issues/313

@marcus-h: please review and have a nice weekend.

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -491,7 +491,7 @@ def check_trusted_projects(apiurl, projects):
             print("Note that malicious packages can compromise the build result or even your system.")
             r = raw_input(trustprompt % { 'project': prj })
             if r == '1':
-                print("adding '%s' to ~/.oscrc: ['%s']['trusted_prj']" % (prj, apiurl))
+                print("adding '%s' to oscrc: ['%s']['trusted_prj']" % (prj, apiurl))
                 trusted.append(prj)
             elif r != '2':
                 print("Well, good good bye then :-)")
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -5631,7 +5631,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
 
         The arguments REPOSITORY and ARCH are optional. They can be taken from
         the first two columns of the 'osc repos' output. If not specified,
-        REPOSITORY defaults to the 'build_repositoy' config entry in your '.oscrc'
+        REPOSITORY defaults to the 'build_repositoy' config entry in your 'oscrc'
         and ARCH defaults to your host architecture.
 
         usage:
@@ -6014,7 +6014,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
     @cmdln.option('--nochecks', '--no-checks', action='store_true',
                   help='Do not run build checks on the resulting packages.')
     @cmdln.option('--no-verify', '--noverify', action='store_true',
-                  help='Skip signature verification (via pgp keys) of packages used for build. (Global config in .oscrc: no_verify)')
+                  help='Skip signature verification (via pgp keys) of packages used for build. (Global config in oscrc: no_verify)')
     @cmdln.option('--noservice', '--no-service', action='store_true',
                   help='Skip run of local source services as specified in _service file.')
     @cmdln.option('-p', '--prefer-pkgs', metavar='DIR', action='append',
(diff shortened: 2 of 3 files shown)


lethliel:
I saw that the testsuite fails. Will have a look after lunch.

lethliel:
now the testsuite is working.

marcus-h:
On 2017-10-27 02:43:11 -0700, Marco Strigl wrote:
> @marcus-h: please review and have a nice weekend.

> index 9829f9bd..d43d9638 100644
> --- a/osc/conf.py
> +++ b/osc/conf.py
> @@ -7,12 +7,12 @@
>  
>  """Read osc configuration and store it in a dictionary
>  
> -This module reads and parses ~/.oscrc. The resulting configuration is stored
> +This module reads and parses ~/.config/osc/oscrc. The resulting configuration is stored

Maybe just "oscrc" (as you did before)?

> +        if 'OSC_CONFIG' in os.environ:
> +            conffile = os.environ.get('OSC_CONFIG')
> +        else:
> +            # needed for compat reasons(users may have their oscrc still in ~

Nitpick: missing whitespace + closing parenthesis

> +            if os.path.isfile(os.path.expanduser('~/.oscrc')):
> +                conffile = '~/.oscrc'
> +            else:
> +                conffile = os.environ.get('XDG_CONFIG_HOME', '~/.config') + '/osc/oscrc'

This code is repeated two times. Just move it into a new function:)


> @@ -654,6 +663,7 @@ def write_config(fname, cp):
>      if os.path.exists(fname) and not os.path.isfile(fname):
>          # only write to a regular file
>          return
> +    os.makedirs(os.path.dirname(fname))

This fails if the directory returned by os.path.dirname(fname) already
exists. Maybe we should also use a more restrictive mode instead of the
default 0777 (for instance, the basedir spec recommends 0700 (see [1])).

Actually, our implementation is not fully compliant with the basedir
spec [1], since we do not take $XDG_CONFIG_DIRS and an empty
$XDG_CONFIG_HOME into account... but I can live with that;)

[1] https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

lethliel:
@marcus-h: thanks for your review.

* corrected comment
* moved conffile determination in function
* the created directory has now mode 700
* if the directory already exists the code will not break anymore

lethliel:
@marcus-h: ping :-)

marcus-h:
On 2017-10-30 04:44:48 -0700, Marco Strigl wrote:
> @marcus-h: thanks for your review.
> 
> * corrected comment

Thanks!

> * moved conffile determination in function

Hmm can you also move the os.environ code into identify_conffile?

> * the created directory has now mode 700

Thanks!

> * if the directory already exists the code will not break anymore
> 

> +    try:
> +        os.makedirs(os.path.dirname(fname), mode=0o700)
> +    except OSError as e:
> +        if e.errno != os.errno.EEXIST:
> +            raise
> +        pass

The "pass" is not needed.

(Personally, I would simply do
if not os.path.exists(...):
    os.makedirs(...)
)

lethliel:
On 11/07/2017 05:39 PM, Marcus Hüwe  wrote:
> 
> Hmm can you also move the os.environ code into identify_conffile?
> 

done

> 
> The "pass" is not needed.
> 
> (Personally, I would simply do
> if not os.path.exists(...):
> os.makedirs(...)
> )
> 

Ok. Had it that way in the first place but changed it because I 
thought the exception handling is more python-ish :)

marcus-h:
On 2017-11-08 10:21:35 +0000, Marco Strigl wrote:
> On 11/07/2017 05:39 PM, Marcus Hüwe  wrote:
> > 
> > Hmm can you also move the os.environ code into identify_conffile?
> > 
> 
> done
> 
Great!

> > 
> > The "pass" is not needed.
> > 
> > (Personally, I would simply do
> > if not os.path.exists(...):
> > os.makedirs(...)
> > )
> > 
> 
> Ok. Had it that way in the first place but changed it because I 
> thought the exception handling is more python-ish :) 
> 
Hmm hard to give a generic answer here:) In this particular case,
os.makedirs should only be executed if the dir does not exist. This
condition can be easily checked.
The try-except approach always calls os.makedirs ("overhead") and,
in some sense, we misuse exceptions for control flow.
However, sometimes the latter is more convenient... for instance,
assume we have something like this:

data = raw_input('enter an integer: ')
# convert data to an int

control flow approach:

if re.match('^\s*[+-]?\s*\d+\s*$', data) is not None:
    i = int(data)
else:
    print('error')

vs.

try-except approach:

try:
    i = int(data)
except ValueError:
    print('error')

In the end, it is always common sense + personal matter of taste;)

Btw., I just merged the PR! It is a great improvement. Thanks!

PR #350, opened 2017-10-30
added code to get the sha256 hashes of files

lethliel:
This is needed for a new validation of the source server. 
It will 'ask' the client for the sha256 sums of files to check if someone is trying to use
md5 collisions to do something bad.

Related to commit https://github.com/openSUSE/open-build-service/commit/0ca1facc1845f3c99a9e154b2e04b76ee174fb2e

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -22,6 +22,7 @@ import re
 import socket
 import errno
 import shlex
+import hashlib
 
 try:
     from urllib.parse import urlsplit, urlunsplit, urlparse, quote_plus, urlencode, unquote
@@ -1393,7 +1394,7 @@ class Package:
             todo.append(n.get('name'))
         return todo
 
-    def __send_commitlog(self, msg, local_filelist):
+    def __send_commitlog(self, msg, local_filelist, validate=False):
         """send the commitlog and the local filelist to the server"""
         query = {}
         if self.islink() and self.isexpanded():
@@ -1405,6 +1406,8 @@ class Package:
                 query['linkrev'] = self.get_pulled_srcmd5()
         if self.islinkrepair():
             query['repairlink'] = '1'
+        if validate:
+            query['withvalidate'] = '1'
         return self.commit_filelist(self.apiurl, self.prjname, self.name,
                                     local_filelist, msg, **query)
 
@@ -1444,6 +1447,7 @@ class Package:
         todo_send = {}
         todo_delete = []
         real_send = []
+        sha256sums = {}
         for filename in self.filenamelist + [i for i in self.to_be_added if not i in self.filenamelist]:
             if filename.startswith('_service:') or filename.startswith('_service_'):
                 continue
@@ -1454,6 +1458,7 @@ class Package:
             elif filename in self.todo:
                 if st in ('A', 'R', 'M'):
                     todo_send[filename] = dgst(os.path.join(self.absdir, filename))
+                    sha256sums[filename] = sha256_dgst(os.path.join(self.absdir, filename))
                     real_send.append(filename)
                     print(statfrmt('Sending', os.path.join(pathn, filename)))
                 elif st in (' ', '!', 'S'):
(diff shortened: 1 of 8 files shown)


lethliel:
@marcus-h: please review

marcus-h:
On 2017-10-30 12:55:00 +0000, Marco Strigl wrote:
> @marcus-h: please review
> 
Will do, once [1] is clarified.


[1] https://lists.opensuse.org/obs-commits/2017-10/msg00159.html

marcus-h:
On 2017-10-30 12:55:00 +0000, Marco Strigl wrote:
> @marcus-h: please review

I just had a very, very quick glance:
- please use True/False instead of 1/0
- instead of reading the whole file into memory, let's just
  feed some "small" chunks to the sha256's update method
  (see, for instance, the core.dgst function (yes, this
  function deserves a cleanup:p)).

(sorry for the short feedback...)

lethliel:
On 11/07/2017 06:05 PM, Marcus Hüwe  wrote:
> On 2017-10-30 12:55:00 +0000, Marco Strigl wrote:
>> @marcus-h: please review
> 
> I just had a very, very quick glance:
> - please use True/False instead of 1/0

I don't know what you mean. Do you mean this:

sfilelist = self.__send_commitlog(msg, filelist, validate=0)

validate is a query argument for the api which indicates the backend to perform 
a file validation check or not. The api expects a 0 or 1 not true or false. 

> - instead of reading the whole file into memory, let's just
> feed some "small" chunks to the sha256's update method
> (see, for instance, the core.dgst function (yes, this
> function deserves a cleanup:p)).
>

I created a function sha256_dgst like the existing dgst function.
 
> (sorry for the short feedback...)
> 

No problem. Thanks for the review

lethliel:
Ok. In the new obs we have two states for a file entry: 

**`$entry->{'hash'} = 'new'`**
indicates that this is a complete new file that should be commited together 
with its sha256 sum. 

**`$entry->{'hash'} = 'missing'`**
indicates that this file exists already on the server and the server needs the sha256 sum to 
compare if it is the same file or if we have some md5 collision here.

If the returned filelist does only contain entries without the value 'hash' we assume that the server
is an old version which does not support sha256 validation. So we just figure out the missing files and send them to the server. 

I tried to visualize it: 


sha256_check

lethliel:
@marcus-h: Can you please take a look? Thanks. 
I will rebase in the end.

marcus-h:
On 2017-11-14 09:51:04 +0000, Marco Strigl wrote:
> @marcus-h: Can you please take a look? Thanks. 
> I will rebase in the end.
> 
Hmm the current code looks a bit fishy to me (the diff below is
the result of "git diff master..your_branch"):

> @@ -1487,6 +1511,16 @@ class Package:
>          print('Transmitting file data', end=' ')
>          filelist = self.__generate_commitlist(todo_send)
>          sfilelist = self.__send_commitlog(msg, filelist)

Old obs: could create a commit
New obs: does not create a commit (except if filelist is empty, but
let's ignore this case)

> +        send_with_hash, just_hash = self.commit_get_hashinfo(sfilelist)
> +        for hfile in just_hash:
> +            sha256hash = sha256_dgst(hfile)

This will fail, if we do something like "osc ci /path/to/wc" and
cwd != /path/to/wc. Also, using the file from the wc is in general
wrong (for instance, assume file "foo" has state 'M' and we just
commit file "bar" => this most-likely yields to a sha256 mismatch).

> +            filelist.find('.//entry[@name="%s"]' % hfile).set('hash', 'sha256:%s' % sha256hash)
> +        if just_hash:
> +            sfilelist = self.__send_commitlog(msg, filelist)

Hmm why is this needed? This is already covered by the subsequent
__send_commitlog call. Why do we differentiate between the two
lists at all?

New obs: could create a commit

> +        for sfile in send_with_hash:
> +            sha256hash = sha256_dgst(sfile)

See above.

> +            filelist.find('.//entry[@name="%s"]' % sfile).set('hash', 'sha256:%s' % sha256hash)
> +        sfilelist = self.__send_commitlog(msg, filelist)

Old obs: could create a commit
New obs: could create a commit

>          send = self.commit_get_missing(sfilelist)
>          real_send = [i for i in real_send if not i in send]
>          # abort after 3 tries
> @@ -1497,6 +1531,8 @@ class Package:
>              if os.path.isdir(tdir):
>                  shutil.rmtree(tdir)
>              os.mkdir(tdir)
> +            sfilelist = self.__send_commitlog(msg, filelist, validate=False)

Old obs: could create a commit
New obs: could create a commit

=> the current code might lead to _three_ commits (for old and new
obs instances).

> +            send = self.commit_get_missing(sfilelist)
>              while len(send) and tries:
>                  for filename in send[:]:
>                      sys.stdout.write('.')
> @@ -1504,7 +1540,7 @@ class Package:
>                      self.put_source_file(filename, tdir)
>                      send.remove(filename)
>                  tries -= 1
> -                sfilelist = self.__send_commitlog(msg, filelist)
> +                sfilelist = self.__send_commitlog(msg, filelist, validate=False)
>                  send = self.commit_get_missing(sfilelist)
>              if len(send):
>                  raise oscerr.PackageInternalError(self.prjname, self.name,
>

marcus-h:
Merged. Thanks!

PR #352, opened 2017-11-08
consider arch when checking the disabled repos

lethliel:
At the moment just repo.name is considered. So if the repo is disabled for s390 all other repo / arch combination are not shown in the repo list for 

`osc repos`
`osc getbinaries`

To be able to change this r is now a list of dicts containing the name and arch of the disabled repo.

`None` for repo if a complete arch gets disabled
`None` for arch if a complete repo gets disabled

This is a proposal to fix https://github.com/openSUSE/osc/issues/351

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -5846,13 +5846,17 @@ Please submit there instead, or use --nodevelproject to force direct submission.
 
         if subcmd == 'repos_only':
             for repo in get_repositories_of_project(apiurl, project):
-                if (disabled is None) or ((disabled is not None) and (repo not in disabled)):
+                if (disabled is None) or ((disabled is not None) and (repo not in [d['repo'] for d in disabled])):
                     print(repo)
         else:
             data = []
             for repo in get_repos_of_project(apiurl, project):
-                if (disabled is None) or ((disabled is not None) and (repo.name not in disabled)):
-                    data += [repo.name, repo.arch]
+                if disabled is not None:
+                    if ({'repo': repo.name, 'arch': repo.arch} in disabled
+                        or repo.name in [d['repo'] for d in disabled if d['arch'] is None]
+                        or repo.arch in [d['arch'] for d in disabled if d['repo'] is None]):
+                            continue
+                data += [repo.name, repo.arch]
 
             for row in build_table(2, data, width=2):
                 print(row)
--- a/osc/core.py
+++ b/osc/core.py
@@ -3554,7 +3554,12 @@ def show_package_disabled_repos(apiurl, prj, pac):
     try:
         root = ET.fromstring(''.join(m))
         elm = root.find('build')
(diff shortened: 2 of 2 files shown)


marcus-h:
On 2017-11-08 09:59:08 +0000, Marco Strigl wrote:
> index 03ad3915..04d67eb3 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -5846,13 +5846,14 @@ def do_repositories(self, subcmd, opts, *args):
>  
>          if subcmd == 'repos_only':
>              for repo in get_repositories_of_project(apiurl, project):
> -                if (disabled is None) or ((disabled is not None) and (repo not in disabled)):
> +                if (disabled is None) or ((disabled is not None) and (repo not in [d['repo'] for d in disabled])):
>                      print(repo)
>          else:
>              data = []
>              for repo in get_repos_of_project(apiurl, project):
> -                if (disabled is None) or ((disabled is not None) and (repo.name not in disabled)):
> -                    data += [repo.name, repo.arch]
> +                if (disabled is None) or (disabled is not None):

This is always True (hence, the subsequent line breaks if disabled is
None).

> +                    if (repo.name not in [d['repo'] for d in disabled]) or (repo.arch not in [d['arch'] for d in disabled]):
> +                        data += [repo.name, repo.arch]

Hmm if we do something like this a
<disable repository="openSUSE:Factory" arch="i586"/>
would result in:
- all openSUSE:Factory repos are disabled (regardless of their arch)
- all repos with an i586 arch are disabled (regardless of their name)

I think we should do something like this

if ({'repo': repo.name, 'arch': repo.arch} in disabled
    or repo.name in [d['repo'] for d in disabled if d['arch'] is None]
    or repo.arch in [d['arch'] for d in disabled if d['repo'] is None]):
    # skip repo

lethliel:
On 11/08/2017 12:10 PM, Marcus Hüwe  wrote:
>
>> + if (disabled is None) or (disabled is not None):
> 
> This is always True (hence, the subsequent line breaks if disabled is
> None).
>

Ups. Yes. Revised that part.
 
>> + if (repo.name not in [d['repo'] for d in disabled]) or (repo.arch not in [d['arch'] for d in disabled]):
>> + data += [repo.name, repo.arch]
> 
> Hmm if we do something like this a
> <disable repository="openSUSE:Factory" arch="i586"/>
> would result in:
> - all openSUSE:Factory repos are disabled (regardless of their arch)
> - all repos with an i586 arch are disabled (regardless of their name)
> 
> I think we should do something like this
> 
> if ({'repo': repo.name, 'arch': repo.arch} in disabled
> or repo.name in [d['repo'] for d in disabled if d['arch'] is None]
> or repo.arch in [d['arch'] for d in disabled if d['repo'] is None]):
> # skip repo

Done

marcus-h:
Merged. Thanks!

PR #353, opened 2017-11-08
Small documentation fix to clarify attribute usage.

abergmann:
The `osc meta attribute` usage help is not 100% clear how to create and set an `attribute`.

If an ATTRIBUTE needs a value during creation the `--set` option needs to be added. Right now the create/delete/set options are shown with an or operator.

The `value_list` for the `--set` operator is not optional, therefore `[value_list]` needs to be replaced with `<value_list>` to make this clear.


marcus-h:
On 2017-11-08 10:07:25 +0000, Alexander Bergmann wrote:
> The `osc meta attribute` usage help is not 100% clear how to create and set an `attribute`.
> 

Issue #355, opened 2017-11-21
subprocess.check_output() not in Python 2.6


birkenfeld:
Is osc supposed to be still Python 2.6/RHEL6 compatible? If yes, `check_output` used [here](https://github.com/openSUSE/osc/blob/master/osc/core.py#L7575) needs to be replaced, as it's new in 2.7.

lethliel:
I created a PR to make it 2.6 compatible.

birkenfeld:
Thanks, that was fast :)

PR #357, opened 2017-11-21
return_external python 2.6 compatibility

lethliel:
if subprocess has no method check_output us Popen instead.

should fix https://github.com/openSUSE/osc/issues/355

@marcus-h: please review

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -7567,6 +7567,16 @@ def return_external(filename, *args, **kwargs):
         cmd = filename
 
     try:
+        # backward compatibility for python 2.6
+        if 'check_output' not in dir(subprocess):
+            process = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+            output, errstr = process.communicate()
+            retcode = process.poll()
+            if retcode:
+                error = subprocess.CalledProcessError(retcode, cmd)
+                error.output = output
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2017-11-21 10:40:20 +0000, Marco Strigl wrote:
> if subprocess has no method check_output us Popen instead.
> 
> should fix https://github.com/openSUSE/osc/issues/355
> 
> @marcus-h: please review

> index 10f28ce0..8acf3f3e 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -7567,6 +7567,11 @@ def return_external(filename, *args, **kwargs):
>          cmd = filename
>  
>      try:
> +        # backward compatibility for python 2.6
> +        if 'check_output' not in dir(subprocess):

(An alternative would have been to use sys.version_info, but ok).

> +            process = subprocess.Popen(cmd, stdout=subprocess.PIPE)
> +            ret_val, errstr = process.communicate()

Misnamed variable: ret_val is actually stdout.
Maybe we should also take the retval into account to resemble the
subprocess.check_output behavior.

> +            return ret_val
>          return subprocess.check_output(cmd, **kwargs)
>      except OSError as e:
>          if e.errno != errno.ENOENT:

PR #366, opened 2018-01-11
fixes not working osc maintainer <prj>

### LINKED ISSUE #150: Getting project maintainers broken
I want to see who the maintainers are for a project:

`osc maintainer editors`
Empty search result, you may want to search with other or all roles via -r ''

Somewhere between versions 0.145 and 0.151 someone added BINARY to PRJ:
`osc maintainer -h`

>    osc maintainer BINARY <options>
>    osc maintainer PRJ <options>

It used to work in 0.145. The help output tells me it still ought to work.

To me it looks like someone overlayed BINARY, extending functionailty in a hurry, without checking too much whether it breaks existing functionality. And WTF is BINARY anyway? No mention of it. Not a user-friendly implementation of a new feature.

lethliel:
fixes #150 

The retrun at this point breaks the call, because in most
cases `<prj>` is not a binary. And the code always checks for the
binary first and then returns if no binary with the name `<prj>`
is found.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -7938,9 +7938,6 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                             repl = raw_input('\nUse this container? (y/n) ')
                             if repl.lower() != 'y':
                                 searchresult = None
-                    else:
-                       print("Empty search result, you may want to search with other or all roles via -r ''")
-                       return
             elif opts.user:
                 searchresult = owner(apiurl, opts.user, "user", usefilter=filterroles, devel=None)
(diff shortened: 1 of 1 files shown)


lethliel:
@marcus-h please review

marcus-h:
On 2018-01-11 02:57:38 -0800, Marco Strigl wrote:
> fixes #150 
> 
> The retrun at this point breaks the call, because in most
> cases `<prj>` is not a binary. And the code always checks for the
> binary first and then returns if no binary with the name `<prj>`
> is found.

Merged. Thanks!

PR #368, opened 2018-01-18
Export a real name for vc to consume

Conan-Kudo:
Required for https://github.com/openSUSE/obs-build/pull/214

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -8787,18 +8787,40 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         else:
             apiurl = self.get_api_url()
 
-        # set user's email if the mailaddr env variable is not set
-        if 'mailaddr' in os.environ:
+        # set user's full name and email if the realname and mailaddr env variables are not set
+
+        vc_id_missing = { 'realname': False, 'email': False }
+
+        if 'VC_REALNAME' in os.environ:
+            pass
+        elif 'realname' in conf.config['api_host_options'][apiurl]:
+            os.environ['VC_REALNAME'] = conf.config['api_host_options'][apiurl]['realname']
+        else:
+            vc_id_missing['realname'] = True
+
+        if 'VC_MAILADDR' in os.environ or 'mailaddr' in os.environ:
             pass
         elif 'email' in conf.config['api_host_options'][apiurl]:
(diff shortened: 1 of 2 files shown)


osc/commandline.py
```diff
@@ -8675,7 +8675,20 @@ def do_vc(self, subcmd, opts, *args):
         else:
             apiurl = self.get_api_url()
 
-        # set user's email if the mailaddr env variable is not set
+        # set user's full name and email if the realname and mailaddr env variables are not set
+
+        if 'realname' in os.environ:
+            pass
+        elif 'realname' in conf.config['api_host_options'][apiurl]:
+            os.environ['realname'] = conf.config['api_host_options'][apiurl]['realname']
```
adrianschroeter: "realname" is not a good idea for a global ENV variable, but use something like "OSC_VC_REALNAME" or alike.

Conan-Kudo: Do we also want to change the "mailaddr" one too? It's pretty generic as well...

Conan-Kudo:
@adrianschroeter This should be addressed now.

Conan-Kudo:
@marcus-h Could you please review/merge this?

jonsger:
Is there a way to disable this behaviour?

Conan-Kudo:
@jonsger `obs-build` has a flag to filter out the name if you want the old behavior, on a per-project basis. But I don't know of a reason why you'd intentionally want the ambiguity of email-address-only.

jonsger:
```
-------------------------------------------------------------------
Wed Nov 21 16:30:18 UTC 2018 - Jonathan Brielmaier <jbrielmaier@suse.de>
```
^ It looks a little broken, if you have a long name...

Conan-Kudo:
@jonsger I'm curious if extending the number of dashes would actually break the format...

lethliel:
The question is how long the line should be.

krig:
It looks like this PR broke the `tar_scm` service which uses `$VC_MAILADDR` to get the author name:

https://github.com/openSUSE/obs-service-tar_scm/issues/296

Conan-Kudo:
This PR sets `VC_MAILADDR`, so I don't know what the issue is?

krig:
> This PR sets `VC_MAILADDR`, so I don't know what the issue is?

Yeah, my mistake. There is something else causing the problem.

Issue #369, opened 2018-01-22
crash trying to view diff of a request


AdamMajer:
Hello,

    osc --version
    0.161.1git

using internal OBS,

    isc rq show -i 136448
    i (for diff)

results in crash `AttributeError: 'NoneType' object has no attribute 'strip'` with following backtrace

    File "/usr/bin/osc", line 41, in <module>
      r = babysitter.run(osccli)
    File "/usr/lib/python2.7/site-packages/osc/babysitter.py", line 61, in run
      return prg.main(argv)
    File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 344, in main
      return self.cmd(args)
    File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 367, in cmd
      retval = self.onecmd(argv)
    File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 501, in onecmd
      return self._dispatch_cmd(handler, argv)
    File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 1232, in _dispatch_cmd
      return handler(argv[0], opts, *args)
    File "/usr/lib/python2.7/site-packages/osc/commandline.py", line 2398, in do_request
      source_buildstatus=source_buildstatus)
    File "/usr/lib/python2.7/site-packages/osc/core.py", line 7312, in request_interactive_review
      issues = '\n\n' + get_formatted_issues(apiurl, request.reqid)
    File "/usr/lib/python2.7/site-packages/osc/core.py", line 7266, in get_formatted_issues
      issue_list = get_request_issues(apiurl, reqid)
    File "/usr/lib/python2.7/site-packages/osc/core.py", line 4835, in get_request_issues
      issue[issue_detail.tag] = issue_detail.text.strip()

lethliel:
will look at it.

PR #370, opened 2018-01-22
only assign issue detail if text for tag exists

lethliel:
check if issue_detail.text exists for issue_detail.tag before assigning it. 
This will fix the crashes reported in #369 

@marcus-h please review

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -4843,7 +4843,8 @@ def get_request_issues(apiurl, reqid):
         for cissue in collection:
             issue = {}
             for issue_detail in cissue.iter():
-                issue[issue_detail.tag] = issue_detail.text.strip()
+                if issue_detail.text:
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2018-01-22 10:11:22 +0000, Marco Strigl wrote:
> check if issue_detail.text exists for issue_detail.tag before assigning it. 
> This will fix the crashes reported in #369 
> 
For which tag does the crash happen? Can this break the code
that accesses the "label" tag in get_formatted_issues?

lethliel:
It happens for the realname tag. And this tag we do not query in get_formatted_issues anyway. 
So nothing breaks.

marcus-h:
On 2018-01-22 13:40:53 +0000, Marco Strigl wrote:
> It happens for the realname tag. And this tag we do not query in get_formatted_issues anyway. 
> So nothing breaks. 
> 
Ok. Merged. Thanks!

PR #374, opened 2018-01-28
osc chroot: also mount devpts

andreas-schwab:
This is needed for running expect, for example.


marcus-h:
Merged. Thanks a lot!. Also, thanks for the missing "sysfs" bits.
(We should eventually implement "osc chroot" via obs-build... *sigh*)

PR #377, opened 2018-02-06
[RFC] replace python-urlgrabber with own module

lethliel:
This is for discussing the substitution of python-urlgrabber


marcus-h:
On 2018-02-14 06:22:05 -0800, Marco Strigl wrote:
> @marcus-h: revised the urlgrabber replacement a little bit. If you want to take a look before friday. 
> 
Here's a very quick review:

> diff --git a/osc/babysitter.py b/osc/babysitter.py
> @@ -132,7 +132,7 @@ def run(prg, argv=None):
>          print(e, file=sys.stderr)
>      except URLError as e:
>          print('Failed to reach a server:\n', e.reason, file=sys.stderr)
> -    except URLGrabError as e:
> +    except MGError as e:
>          print('Failed to grab %s: %s' % (e.url, e.strerror), file=sys.stderr)

MGError has no "url" attribute.

> index c7ac5f6..0524957 100644
  
>      def get_cmd_help(self, cmdname):
> @@ -7702,8 +7702,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
>  
>          if '://' in srpm:
>              print('trying to fetch', srpm)
> -            import urlgrabber
> -            urlgrabber.urlgrab(srpm)
> +            from .mirror import OscMirrorGroup
> +            OscMirrorGroup.urlgrab(srpm, filename=srpm)

This will result in an TypeError (I'm not sure if we should
use the mirror group code here). (Also, filename "==" url
looks fishy to me.)

>              srpm = os.path.basename(srpm)
>  

> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -5988,8 +5988,18 @@ def streamfile(url, http_meth = http_GET, bufsize=8192, data=None, progress_obj=
>          cl = int(cl)
>  
>      if progress_obj:
> -        basename = os.path.basename(urlsplit(url)[2])
> -        progress_obj.start(basename=basename, text=text, size=cl)
> +        if not text:
> +            basename = os.path.basename(urlsplit(url)[2])
> +        else:
> +            basename = text
> +        if cl is not None:
> +            widgets = [basename + ': ', progress_obj.Percentage(), progress_obj.Bar(), ' ', progress_obj.ETA()]
> +            bar = progress_obj.ProgressBar(widgets=widgets, maxval=cl)
> +        else:
> +            widgets = [basename + ': ', progress_obj.AnimatedMarker(), ' ' , progress_obj.Timer()]
> +            bar = progress_obj.ProgressBar(widgets=widgets, maxval=progress_obj.UnknownLength)
> +        print() #print pkt_act/pkt_todo
> +        bar.start()
>  
The way we use "progress_obj" is really counterintuitive (it is rather
a progress_mod than an object). We should simply encapsulate the logic
from above in an own osc progress bar class. The interface could
look like this:
class OscProgressBar(object):
    def __init__(self, text, total=None):
        ...

    def start(self):
        ...

    def update(self, length):
        ...

    def end(self, length??):
        ...

(we can also rename the methods, but this interface guarantees backwards
compatibility (I'm not sure if anyone is using our progress object code
path, though)).

>      if bufsize == "line":
>          bufsize = 8192
> @@ -6004,11 +6014,11 @@ def streamfile(url, http_meth = http_GET, bufsize=8192, data=None, progress_obj=
>              break
>          read += len(data)
>          if progress_obj:
> -            progress_obj.update(read)
> +            bar.update(read)
>          yield data
>  
>      if progress_obj:
> -        progress_obj.end(read)
> +        bar.finish()
>      f.close()
>  
>      if not cl is None and read != cl:

> @@ -37,11 +37,8 @@ def join_url(self, base_url, rel_url):
>      return base_url
>  
>  
> -class OscFileGrabber(URLGrabber):
> +class OscFileGrabber():

We should probably inherit from object (in order to have a new-style
class in python2).

> @@ -50,20 +47,20 @@ class OscFileGrabber(URLGrabber):
>              if os.path.isfile(f):
>                  return f
>              else:
> -                raise URLGrabError(2, 'Local file \'%s\' does not exist' % f)
> -        with file(filename, 'wb') as f:
> +                raise grabError(2, 'Local file \'%s\' does not exist' % f)

grabError is not defined.

> +        with open(filename, 'wb') as f:
>              try:
>                  for i in streamfile(url, progress_obj=self.progress_obj,
>                                      text=text):
>                      f.write(i)
>              except HTTPError as e:
> -                exc = URLGrabError(14, str(e))
> +                exc = str(e)

Hmm this breaks the code below:

>                  exc.url = url
>                  exc.exception = e
>                  exc.code = e.code
>                  raise exc


> @@ -156,7 +145,7 @@ class Fetcher:
>                          raise oscerr.APIError('failed to fetch file \'%s\': '
>                                                'missing in CPIO archive' %
>                                                pac.repofilename)
> -        except URLGrabError as e:
> +        except MGError as e:
>              if e.errno != 14 or e.code != 414:

Hrm this is currently flawed:
- the current code never sets the code attribute (since the
  OscFileGrabber is not used)
- the old code is also flawed (in case the MGError "represents" an
  IOError)

(Maybe we should get rid of URLGrabError/MGError and its strange
error codes).

>                  raise
>              # query str was too large
> @@ -181,8 +170,8 @@ class Fetcher:
>          # for use by the failure callback
>          self.curpac = pac
>  
> -        MirrorGroup._join_url = join_url
> -        mg = MirrorGroup(self.gr, pac.urllist, failure_callback=(self.failureReport, (), {}))
> +        #MirrorGroup._join_url = join_url
> +        mg = OscMirrorGroup(self.gr, pac.urllist)
>  
>          if self.http_debug:
>              print('\nURLs to try for package \'%s\':' % pac, file=sys.stderr)
> @@ -195,12 +184,14 @@ class Fetcher:
>                  mg.urlgrab(pac.filename, filename=tmpfile.name,
>                             text='%s(%s) %s' % (prefix, pac.project, pac.filename))
>                  self.move_package(tmpfile.name, pac.localdir, pac)
> -        except URLGrabError as e:
> +        except Exception as e:

That's way too general. In general, an instance of the class Exception
does not have an errno attribute.


> diff --git a/osc/mirror.py b/osc/mirror.py

> +class MGError(IOError):
> +    def __init__(self, *args):
> +        IOError.__init__(self, *args)
> +
> +class OscMirrorGroup:
> +    def __init__(self, grabber, mirrors, **kwargs):
> +        self.grabber = grabber
> +        self.mirrors = mirrors
> +
kwargs is unused.

> +    def urlgrab(self, url, filename=None, **kwargs):
> +        max_m = len(self.mirrors)
> +        tries = 0
> +        for mirror in self.mirrors:
> +            if mirror.startswith('file'):
> +                path = mirror.replace('file:/', '')
> +                if not os.path.exists(path):
> +                    tries += 1
> +                    continue
> +                else:
> +                    copyfile(path,filename)
> +                    break
> +            try:
> +                u = urlopen(mirror)
> +            except HTTPError as e:
> +                if e.code == 414:
> +                    raise MGError
> +                tries += 1
> +                continue

Why the urlopen code path?

> +            f = open(filename, 'wb')
> +            for i in streamfile(mirror, progress_obj=pb):
> +                f.write(i)
> +            f.flush
> +            f.close

Some remarks:
- This neither flushes nor closes the file
- Why don't we simply use the grabber object?
- Missing exception handling (for instance, "tries" is not
  incremented in case of an error)?

> +            break
> +
Not needed.

> +        if max_m == tries:
> +            raise MGError(256, 'No mirrors left')

marcus-h:
On 2018-02-21 16:07:01 +0000, Marco Strigl wrote:
> @lethliel pushed 1 commit.
> 
I think we are almost done!:)
Here's a quick review:


> diff --git a/osc/__init__.py b/osc/__init__.py
> index 7872a4e..5de8f26 100644
> --- a/osc/__init__.py
> +++ b/osc/__init__.py
> @@ -1,3 +1,3 @@
> -__all__ = ['babysitter', 'core', 'commandline', 'oscerr', 'othermethods', 'build', 'fetch', 'meter']
> +__all__ = ['babysitter', 'core', 'commandline', 'oscerr', 'othermethods', 'build', 'fetch']
>  
We can keep "meter" here (you could also add the new grabber module, but
that's up to you:) ).
The "othermethods" module can be removed - it is really prehistoric...

> diff --git a/osc/babysitter.py b/osc/babysitter.py
> index 284a3d1..5e1aa54 100644
> --- a/osc/babysitter.py
> +++ b/osc/babysitter.py
> @@ -132,8 +132,8 @@ def run(prg, argv=None):
>          print(e, file=sys.stderr)
>      except URLError as e:
>          print('Failed to reach a server:\n', e.reason, file=sys.stderr)
> -    except URLGrabError as e:
> -        print('Failed to grab %s: %s' % (e.url, e.strerror), file=sys.stderr)
> +    except MGError as e:
> +        print('Failed to grab %s: %s' % (e.errno, e.strerror), file=sys.stderr)

Hmm I'm not sure if the errno is really helpful here, but I think we
can adapt the code so that we do not have to fool around with MGError
here anymore (see below).

> index 3536242..1b0f55e 100644
> --- a/osc/build.py
> +++ b/osc/build.py
> @@ -35,6 +35,7 @@ from .conf import config, cookiejar
>  
>  try:
>      from .meter import TextMeter
> +    Textmeter = TextMeter()

Hmm why do we need to make an instance here?

>  except:
>      TextMeter = None
>  
> @@ -258,7 +259,7 @@ class Pac:
>          self.urllist = []
>  
>          # build up local URL
> -        # by using the urlgrabber with local urls, we basically build up a cache.
> +        # by using the OscMirrorGroup with local urls, we basically build up a cache.
>          # the cache has no validation, since the package servers don't support etags,
>          # or if-modified-since, so the caching is simply name-based (on the assumption
>          # that the filename is suitable as identifier)
> @@ -319,13 +320,13 @@ def get_preinstall_image(apiurl, arch, cache_dir, img_info):
>                  print(e, file=sys.stderr)
>                  sys.exit(1)
>          if sys.stdout.isatty() and TextMeter:
> -            progress_obj = TextMeter(fo=sys.stdout)
> +            progress_obj = TextMeter

Just instantiate TextMeter here.

>          else:
>              progress_obj = None
>          gr = OscFileGrabber(progress_obj=progress_obj)
>          try:
>              gr.urlgrab(url, filename=ifile_path_part, text='fetching image')
> -        except URLGrabError as e:
> +        except MGError as e:

Hmm we need to catch an urllib2.HTTPError here. A MGError cannot occur,
because url does not start with "file:/".

>              print("Failed to download! ecode:%i errno:%i" % (e.code, e.errno))
>              return ('', '', [])
>          # download ok, rename partial file to final file name

> index a2de16f..fbe17b4 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -166,8 +166,8 @@ class Osc(cmdln.Cmdln):
>          self.options.verbose = conf.config['verbose']
>          self.download_progress = None
>          if conf.config.get('show_download_progress', False):
> -            from .meter import TextMeter
> -            self.download_progress = TextMeter(hide_finished=True)
> +            import progressbar as pb
> +            self.download_progress = pb
>  
Just keep the old code.

> @@ -7702,8 +7702,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
>  
>          if '://' in srpm:
>              print('trying to fetch', srpm)
> -            import urlgrabber
> -            urlgrabber.urlgrab(srpm)
> +            from .grabber import OscFileGrabber
> +            OscFileGrabber().urlgrab(srpm)
>              srpm = os.path.basename(srpm)
>  
>          srpm = os.path.abspath(srpm)
> index e8b3c53..14e8798 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -5988,8 +5988,12 @@ def streamfile(url, http_meth = http_GET, bufsize=8192, data=None, progress_obj=
>          cl = int(cl)
>  
>      if progress_obj:
> -        basename = os.path.basename(urlsplit(url)[2])
> -        progress_obj.start(basename=basename, text=text, size=cl)
> +        if not text:
> +            basename = os.path.basename(urlsplit(url)[2])
> +        else:
> +            basename = text
> +        pb = progress_obj
> +        pb.start(basename, cl)
>  
There is no need to introduce a "pb" variable (actually, it makes the
subsequent code "harder" to understand (we check for "progress_obj"
and use "pb":) )).
I have to admit that I didn't have a look at the progressbar module
so far, but do we get in both cases the "expected" output? (I'm just
asking, because "basename = text" looks a bit suspicious)

>      if bufsize == "line":
>          bufsize = 8192
> @@ -6004,11 +6008,11 @@ def streamfile(url, http_meth = http_GET, bufsize=8192, data=None, progress_obj=
>              break
>          read += len(data)
>          if progress_obj:
> -            progress_obj.update(read)
> +            pb.update(read)
>          yield data
>  
Use progress_object (see above).

>      if progress_obj:
> -        progress_obj.end(read)
> +        pb.end()

Use progress_object (see above).

> index b96d1ae..34d470e 100644
> --- a/osc/fetch.py
> +++ b/osc/fetch.py
> @@ -14,19 +14,19 @@ except ImportError:
>      #python 2.x
>      from urllib import quote_plus
>      from urllib2 import HTTPBasicAuthHandler, HTTPCookieProcessor, HTTPPasswordMgrWithDefaultRealm, HTTPError
> -
> -from urlgrabber.grabber import URLGrabber, URLGrabError
> -from urlgrabber.mirror import MirrorGroup
>  from .core import makeurl, streamfile, dgst
> +from .grabber import OscFileGrabber, OscMirrorGroup, MGError
>  from .util import packagequery, cpio
>  from . import conf
>  from . import oscerr
>  import tempfile
>  import re
> +
>  try:
>      from .meter import TextMeter
> -except:
> -    TextMeter = None
> +    useProgressBar = True
> +except ImportError:
> +    useProgressBar = None
>  
+1 for just catching ImportError!:)
Hmm why not simply keeping "TextMeter = None" in the exception case?
(If you desperately want to keep the "useProgressBar" variable, set
it to "False" in the exception case)

>  def join_url(self, base_url, rel_url):
> @@ -36,43 +36,12 @@ def join_url(self, base_url, rel_url):
>      """
>      return base_url
>  
Let's get rid of join_url, because it is not used anymore.

> @@ -195,12 +155,14 @@ class Fetcher:
>                  mg.urlgrab(pac.filename, filename=tmpfile.name,
>                             text='%s(%s) %s' % (prefix, pac.project, pac.filename))
>                  self.move_package(tmpfile.name, pac.localdir, pac)
> -        except URLGrabError as e:
> +        except MGError as e:
>              if self.enable_cpio and e.errno == 256:
> +                print('%s/%s: attempting download from api, since not found'
> +                      % (self.curpac.project, pac))

We can get rid of the self.curpac attribute. Just use pac.project and
pac.name. (Hmm this slightly changes the error message compared to
the old code, because we do not print the hostname anymore)

>                  self.__add_cpio(pac)
>                  return
>              print()
> -            print('Error:', e.strerror, file=sys.stderr)
> +            print('Error:', e, file=sys.stderr)

Hmm why not keeping the old code? Printing the errno is probably not
too helpful.

>              print('Failed to retrieve %s from the following locations '
>                    '(in order):' % pac.filename, file=sys.stderr)
>              print('\n'.join(pac.urllist), file=sys.stderr)
> @@ -311,7 +273,7 @@ class Fetcher:
>              try:
>                  if self.offline and not os.path.exists(dest):
>                      # may need to try parent
> -                    raise URLGrabError(2)
> +                    raise MGError(2)
>                  elif not self.offline:
>                      OscFileGrabber().urlgrab(url, dest)
>                  # not that many keys usually
> @@ -324,7 +286,7 @@ class Fetcher:
>                  if os.path.exists(dest):
>                      os.unlink(dest)
>                  sys.exit(0)
> -            except URLGrabError as e:
> +            except MGError as e:
>                  # Not found is okay, let's go to the next project
>                  if e.errno == 14 and e.code != 404:

The only MGError exception we can get here is errno == 2 (without a
code attribute). (Maybe we can get rid of the MGError here as well?)

We definitely need to handle an urllib2.HTTPError here (that's what
the OscFileGrabber may raise).

> diff --git a/osc/grabber.py b/osc/grabber.py
> new file mode 100644
> index 0000000..68512f2
> --- /dev/null
> +++ b/osc/grabber.py
> @@ -0,0 +1,72 @@
> +# Copyright (C) 2018 SUSE Linux.  All rights reserved.
> +# This program is free software; it may be used, copied, modified
> +# and distributed under the terms of the GNU General Public Licence,
> +# either version 2, or (at your option) any later version.
> +
> +import sys
> +import os.path
> +from shutil import copyfile
> +from .core import streamfile
> +
> +try:
> +    from urllib.parse import unquote
> +    from urllib.request import urlopen, HTTPError, url2pathname
> +except ImportError:
> +    from urllib2 import urlopen, HTTPError, url2pathname
> +    from urllib import unquote
> +
urlopen and url2pathname are never used.

> +class MGError(IOError):
> +    def __init__(self, *args):
> +        IOError.__init__(self, *args)
> +
Hmm do we really need to inherit from IOError? If so, you could
write it like this:

class MGError(IOError):
    pass

> +class OscFileGrabber(object):
> +    def __init__(self, progress_obj=None):
> +        self.progress_obj = progress_obj
> +
> +    def urlgrab(self, url, filename=None, text=None, **kwargs):

kwargs is unused.

> +        if filename is None:
> +            filename = os.path.basename(unquote(path))
> +            if not filename:
> +                # This is better than nothing.
> +                filename = 'osc_urlgrab_download'

Hmm how about making the filename required? Then it is the caller's
responsability to supply a correct filename. (We would just need to
adapt the call in do_importsrcpkg.)

> +        if url.startswith('file://'):
> +            f = url.replace('file://', '', 1)
> +            if os.path.isfile(f):
> +                return f
> +            else:
> +                raise MGError(2, 'Local file \'%s\' does not exist' % f)

We do not need to support a file:// url, because this code is only called
if the file url does not exist (see Fetcher.run). Even worse, this code
path is racy, because if url is a file:// url and the file was created in
the meantime (for instance, a second "osc build..." was executed) then
Fetcher.move_package will fail...
So let's simply get rid of it (sorry for not noticing this earlier:/)


> +        with open(filename, 'wb') as f:
> +            for i in streamfile(url, progress_obj=self.progress_obj,
> +                                text=text):
> +                f.write(i)
> +            return filename
> +
We can remove the return (I know the old code did this as well, but
it is never used...).

> +
> +class OscMirrorGroup(object):
> +    def __init__(self, grabber, mirrors):
> +        self.grabber = grabber
> +        self.mirrors = mirrors
> +
Minor: we could "mark" them as private by prefixing the attributes
with a "_" (this is just a convention)

> +    def urlgrab(self, url, filename=None, **kwargs):
> +        max_m = len(self.mirrors)
> +        tries = 0
> +        for mirror in self.mirrors:
> +            if mirror.startswith('file'):
> +                path = mirror.replace('file:/', '')
> +                if not os.path.exists(path):
> +                    tries += 1
> +                    continue
> +                else:
> +                    copyfile(path,filename)
> +                    break

Let's get rid of the file:// url code here as well. (We also need
to adapt Pac.makeurls (in build.py) (just remove the self.url_local
parts)).

> +            try:
> +                self.grabber.urlgrab(mirror, filename) 
> +            except HTTPError as e:
> +                print('Error %s' % e.code)
> +                if e.code == 414:
> +                    raise MGError

Hmm why do you raise a MGError?

> +                tries += 1
> +                continue
> +
The "continue" is not needed.

> +        if max_m == tries:

Minor: len(self.mirrors) == tries (then we do not need the max_m
local variable).

> +            raise MGError(256, 'No mirrors left')
> diff --git a/osc/meter.py b/osc/meter.py
> index f24b18d..85204f4 100644
> --- a/osc/meter.py
> +++ b/osc/meter.py
> @@ -21,83 +21,26 @@
>  

We should also adjust the comments (+ license?) at the beginning
of this file.

>  from __future__ import print_function
>  
Unused.

> -from urlgrabber.progress import BaseMeter, format_time, format_number
> +import progressbar as pb
>  import sys, os
>  
"sys" and "os" are unused.

> -def getScreenWidth():
> -    import termios, struct, fcntl
> -    s = struct.pack('HHHH', 0, 0, 0, 0)
> -    try:
> -        x = fcntl.ioctl(1, termios.TIOCGWINSZ, s)
> -    except IOError:
> -        return 80
> -    return struct.unpack('HHHH', x)[1]
> +class TextMeter(object):
> +    def __init__(self):
> +        pass
>  
Just omit the constructor.

lethliel:
@marcus-h 

* deleted the unused vars and imports
* got rid of most MGErrors
* simplified TextMeter / progress_obj calls
* pep8 check of meter.py and grabber.py

lethliel:
This will be in the python3 branch as well: https://github.com/openSUSE/osc/tree/python3

PR #378, opened 2018-02-06
[RFC / Discussion] python3 compat mode

lethliel:
This is for discussing the python3 compatibility changes.

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -1143,7 +1143,10 @@ def main(apiurl, opts, argv):
         if bi.installonly_list:
             rpmlist.append('installonly: ' + ' '.join(bi.installonly_list) + '\n')
 
-    rpmlist_file = NamedTemporaryFile(mode='w+t', prefix='rpmlist.')
+    if sys.version_info >= (3, 0):
+      rpmlist_file = NamedTemporaryFile(mode='w+t', prefix='rpmlist.')
+    else:
+      rpmlist_file = NamedTemporaryFile(prefix='rpmlist.')
     rpmlist_filename = rpmlist_file.name
     rpmlist_file.writelines(rpmlist)
     rpmlist_file.flush()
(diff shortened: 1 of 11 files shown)


osc/build.py
```diff
@@ -1143,7 +1143,10 @@ def __str__(self):
         if bi.installonly_list:
             rpmlist.append('installonly: ' + ' '.join(bi.installonly_list) + '\n')
 
-    rpmlist_file = NamedTemporaryFile(prefix='rpmlist.')
+    if sys.version_info >= (3, 0):
+      rpmlist_file = NamedTemporaryFile(mode='w+t', prefix='rpmlist.')
```
Conan-Kudo: This variant works in both Python 2 and Python 3, so we can simplify to that.

osc/util/packagequery.py
```diff
@@ -56,7 +56,10 @@ def read(self, all_tags = False, *extra_tags):
     @staticmethod
     def query(filename, all_tags=False, extra_rpmtags=(), extra_debtags=(), self_provides=True):
         f = open(filename, 'rb')
-        magic = f.read(7)
+        if sys.version_info >= (3, 0):
+            magic = f.read(7).decode('latin-1')
```
Conan-Kudo: Why are we forcing `latin-1` here instead of `utf-8`?

osc/build.py
lethliel: yes. you are right.

osc/util/packagequery.py
lethliel: This will be handled completely different. I do not decode here at all. I will work with the bytes

lethliel:
This pull request is obsolete. It will be handled in an own branch until everything works and will be merged in the end. 

https://github.com/openSUSE/osc/tree/python3

PR #381, opened 2018-02-12
Fix osc build -M flavor

andreas-schwab:
Evaluate opts.multibuild_package only after pacname has been determined

### DIFF
--- a/osc/util/rpmquery.py
+++ b/osc/util/rpmquery.py
@@ -166,7 +166,7 @@ class RpmQuery(packagequery.PackageQuery, packagequery.PackageQueryResult):
     def __reqprov(self, tag, flags, version, strong=None):
         pnames = self.header.gettag(tag)
         if not pnames:
-	    return []
+            return []
         pnames = pnames.data
         pflags = self.header.gettag(flags).data
         pvers = self.header.gettag(version).data


marcus-h:
On 2018-02-12 15:49:16 +0000, Andreas Schwab wrote:
> index 35362422..b1421352 100644
> --- a/osc/build.py
> +++ b/osc/build.py
> @@ -618,9 +618,6 @@ def main(apiurl, opts, argv):
>              pac = '_repository'
>          else:
>              pac = store_read_package(os.curdir)
> -    if opts.multibuild_package:
> -        buildargs.append('--buildflavor=%s' % opts.multibuild_package)
> -        pac = pac + ":" + opts.multibuild_package
>      if opts.shell:
>          buildargs.append("--shell")
>  
> @@ -644,6 +641,9 @@ def main(apiurl, opts, argv):
>                  opts.local_package = True
>          if opts.local_package:
>              pacname = os.path.splitext(os.path.basename(build_descr))[0]
> +    if opts.multibuild_package:
> +        buildargs.append('--buildflavor=%s' % opts.multibuild_package)
> +        pacname = pacname + ":" + opts.multibuild_package

In this case, the build flavor is not considered when calculating
the buildinfo. For instance, something like

BuildRequires: foo-@BUILD_FLAVOR@

etc. won't work. Admittedly, even with [1], the @OBS_PACKAGE@
substitution is "broken" in case of the --local-package option,
but at least @BUILD_FLAVOR@ works.

[1] https://github.com/openSUSE/open-build-service/pull/4461

PR #392, opened 2018-03-10
Fix Bash completion if PWD contains whitespace

theMarix:
The script used `${PWD}` without proper quoting causing Bash completion to not work properly if there was whitespace in the path of the current working directory.


marcus-h:
On 2018-03-10 19:46:47 +0000, Matthias Bach wrote:
> The script used `${PWD}` without proper quoting causing Bash completion to not work properly if there was whitespace in the path of the current working directory.

Great! Merged. Thanks for the patch!

PR #393, opened 2018-03-13
fix getbinaries with DOD binaries

lethliel:
getbinaries of dod binaries do not have a size or mtime.
This will break. So just set to None and print unkown instead.

Example output:

` unknown unknown noarch/python3-websocket-client-0.44.0-1.5.noarch.rpm
`

Example: 

```
# osc ls -lb SUSE:SLE-15:GA

Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python2.7/site-packages/osc/babysitter.py", line 61, in run
    return prg.main(argv)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python2.7/site-packages/osc/commandline.py", line 377, in do_list
    results.append((repo, get_binarylist(apiurl, project, repo.name, repo.arch, package=package, verbose=opts.verbose)))
  File "/usr/lib/python2.7/site-packages/osc/core.py", line 5588, in get_binarylist
    int(node.get('size')),
ValueError: invalid literal for int() with base 10: ''
```

This PR fixes this

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -384,7 +384,10 @@ class Osc(cmdln.Cmdln):
 
                 if opts.verbose:
                     for f in result[1]:
-                        print("%9d %s %-40s" % (f.size, shorttime(f.mtime), f.name))
+                        if f.size is None and f.mtime is None:
+                            print("%9s %12s %-40s" % ('unknown', 'unknown', f.name))
+                        else:
+                            print("%9d %s %-40s" % (f.size, shorttime(f.mtime), f.name))
                 else:
                     for f in result[1]:
                         print(indent+f)
--- a/osc/core.py
+++ b/osc/core.py
@@ -5587,8 +5587,8 @@ def get_binarylist(apiurl, prj, repo, arch, package=None, verbose=False):
         for node in tree.findall('binary'):
             f = File(node.get('filename'),
                      None,
-                     int(node.get('size')),
-                     int(node.get('mtime')))
+                     int(node.get('size') or 0) or None,
+                     int(node.get('mtime') or 0) or None)
             l.append(f)
         return l
 


marcus-h:
On 2018-03-13 13:59:38 +0000, Marco Strigl wrote:
> getbinaries of dod binaries do not have a size or mtime.
> This will break. So just set 0. 
> 
Hmm what about setting it to None instead of 0? A zero size
binary might "confuse" users.
(If we set it to None, we have to adjust the callers accordingly.)

lethliel:
On 03/13/2018 03:21 PM, Marcus Hüwe  wrote:
> On 2018-03-13 13:59:38 +0000, Marco Strigl wrote:
>> getbinaries of dod binaries do not have a size or mtime.
>> This will break. So just set 0.
>>
> Hmm what about setting it to None instead of 0? A zero size
> binary might "confuse" users.
> (If we set it to None, we have to adjust the callers accordingly.)

Yes. That was my first approach. Michael said 0 is sufficient.
But if you prefer the "None" Version I can implement that. No problem.

marcus-h:
On 2018-03-13 14:25:28 +0000, Marco Strigl wrote:
> On 03/13/2018 03:21 PM, Marcus Hüwe  wrote:
> > On 2018-03-13 13:59:38 +0000, Marco Strigl wrote:
> >> getbinaries of dod binaries do not have a size or mtime.
> >> This will break. So just set 0.
> >>
> > Hmm what about setting it to None instead of 0? A zero size
> > binary might "confuse" users.
> > (If we set it to None, we have to adjust the callers accordingly.)
> 
> Yes. That was my first approach. Michael said 0 is sufficient.
> But if you prefer the "None" Version I can implement that. No problem. 
> 
Well it is up to you - I have no strong opinion here... (almost
everything is better than an exception;) )

If "None" is translated to "unknown" during "osc ls -v -b...", then
the output is bit more "informative" (but it is also a bit more
work and I'm not sure if it is worth the effort...).

Issue #400, opened 2018-04-10
remote osc_hotshot.py


sleep-walker:
It seems that osc_hotshot.py is long time dead and noone cared. Let's just remove it from GIT and setup.py.

marcus-h:
On 2018-04-10 04:11:02 +0000, Tomáš Čech wrote:
> It seems that osc_hotshot.py is long time dead and noone cared. Let's just remove it from GIT and setup.py.
> 
Let it RIP...
Thanks for the issue + PR!

lethliel:
removed.

PR #401, opened 2018-04-10
remove osc_hotshot.py

sleep-walker:
osc_hotshot.py is long time untouched and broken thus I assume it is
safe to be removed

fixes issue #400


marcus-h:
Merged. Thanks!

PR #405, opened 2018-04-18
fix order of options valdiation. 

### LINKED ISSUE #404: "osc rq list -P" doesn't really filter by project
Example:

```
osc rq -P KDE:Qt5 -s new -t submit
```
I expect to find only requests that involve KDE:Qt5. However this doesn't happen, I get, for example (taken from the massive output that comes out):

```
submit:          home:glaubitz:branches:devel:languages:python/python-azure-batch@2 -> devel:languages:python
```
This request doesn't belong to KDE:Qt5 at all, so either the option does nothing, or behaves in a different way than expected in the documentation (or there's user error ;).

lethliel:
This fixes #404 

If `-P` or `-p` is given osc will now honor these options instead of rewriting
them to None.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -2198,16 +2198,16 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             project = None
             if len(args) > 0:
                 project = args[0]
+            elif opts.project:
+                project = opts.project
+                if opts.package:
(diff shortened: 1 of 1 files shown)


marcus-h:
Sorry for the delay. Merged. Thanks a lot!

PR #412, opened 2018-04-27
always run build using --statistics. I could have made this

adrianschroeter:
yet another option, but
* only very old build scripts don't know it, we should just require a recent one
* build script is ignoring it for chroot case

so why bother with another option?

But doing this via pull request for a second opinion

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,7 @@
 0.163
   - add sendsysrq command (requires OBS 2.10)
   - add addcontainers command (requires OBS 2.10)
+  - enable statistics for local builds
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2018-04-27 13:02:54 +0000, Adrian Schröter wrote:
> yet another option, but
> * only very old build scripts don't know it, we should just require a recent one
> * build script is ignoring it for chroot case
> 
> so why bother with another option?
> 
+1

> But doing this via pull request for a second opinion

No one complained. So here we go... :)

PR #414, opened 2018-05-03
This adds a -D to osc co (--deleted)

### LINKED ISSUE #413: checkout deleted package (osc co -D <project> <package>)
It is not possible to checkout a server side deleted package. See discussion on ML https://lists.opensuse.org/opensuse-buildservice/2018-04/msg00032.html

There should be a osc co -D for deleted packages.

lethliel:
This checks out the sources of a deleted package of a project.
meta files are not considered here, because we do not want to
work ob deleted packages, just show the sources.

fixes #413

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -4241,6 +4241,9 @@ Please submit there instead, or use --nodevelproject to force direct submission.
     @cmdln.option('-e', '--expand-link', action='store_true',
                         help='if a package is a link, check out the expanded '
                              'sources (no-op, since this became the default)')
+    @cmdln.option('-D', '--deleted', action='store_true',
+                        help='checkout a already deleted package. No meta information '
+                       'gets checked out. Just the plain files')
     @cmdln.option('-u', '--unexpand-link', action='store_true',
                         help='if a package is a link, check out the _link file ' \
                              'instead of the expanded sources')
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2018-05-03 13:00:39 +0000, Marco Strigl wrote:
> This checks out the sources of a deleted package of a project.
> meta files are not considered here, because we do not want to
> work ob deleted packages, just show the sources.
> 
Exactly, this directly raises the question why we track deleted
packages at all:) ("package tracking"). IMHO, it does not make
too much sense to track+checkout a deleted package into a (potentially)
existing project. What about just checking out the files into a
plain directory (without creating+tracking a package wc)?

About the patch itself:

> @ -3823,8 +3823,8 @@ def show_files_meta(apiurl, prj, pac, revision=None, expand=False, linkrev=None,
>      f = http_GET(makeurl(apiurl, ['source', prj, pac], query=query))
>      return f.read()
>  
> -def show_upstream_srcmd5(apiurl, prj, pac, expand=False, revision=None, meta=False, include_service_files=False, deleted=False):
> -    m = show_files_meta(apiurl, prj, pac, expand=expand, revision=revision, meta=meta, deleted=deleted)
> +def show_upstream_srcmd5(apiurl, prj, pac, expand=False, revision=None, meta=False, include_service_files=False):
> +    m = show_files_meta(apiurl, prj, pac, expand=expand, revision=revision, meta=meta)

This breaks the existing code.

lethliel:
I will close this as I will do it in a complete different way. New pull request will follow this week.

Issue #406, opened 2018-04-19
Why save password in plain-text? Set gnome-keyring by default.


miphilomath:
Issue/Feature description
====================
Gnome-keyring should be enabled by default while asking for user name and to store password.
So I suggest the following features to be added: 
* To enable the keyring setting by default
* To ask the user about his/her preference during the first configuration setup.

Expected result
============
It would provide user a choice on whether to save his/her password in plain text or using keyring.

Further information
===============
* My operating System: output of `cat /etc/*-release`

> DISTRIB_ID=LinuxMint
> DISTRIB_RELEASE=18.3
> DISTRIB_CODENAME=sylvia
> DISTRIB_DESCRIPTION="Linux Mint 18.3 Sylvia"
> NAME="Linux Mint"
> VERSION="18.3 (Sylvia)"
> ID=linuxmint
> ID_LIKE=ubuntu
> PRETTY_NAME="Linux Mint 18.3"
> VERSION_ID="18.3"
> HOME_URL="http://www.linuxmint.com/"
> SUPPORT_URL="http://forums.linuxmint.com/"
> BUG_REPORT_URL="http://bugs.launchpad.net/linuxmint/"
> VERSION_CODENAME=sylvia
> UBUNTU_CODENAME=xenial

* I installed osc using `sudo apt install osc` (Though I need to install python-keyring later, which should be installed with the osc package as a dependency)

#233 issue is somewhat similar or not but IMO this issue should be given higher priority.
(This issue was first reported at https://github.com/openSUSE/open-build-service/issues/4875)

philo-math:
I would like to know what community thinks about this issue? If it seems okay, I would like to work on it too.

lethliel:
On 04/23/2018 10:04 PM, Mohit Bansal wrote:
> I would like to know what community thinks about this issue? If it seems okay, I would like to work on it too.
> 

I like the idea to ask during the initial call of osc what the user prefers. I also 
would be happy to see some kind of 'server mode' in which the password does not get 
stored at all and queried each time you call an osc command.

marcus-h:
On 2018-04-24 02:07:31 -0700, Marco Strigl wrote:
> On 04/23/2018 10:04 PM, Mohit Bansal wrote:
> > I would like to know what community thinks about this issue? If it seems okay, I would like to work on it too.
> > 
> 
> I like the idea to ask during the initial call of osc what the user prefers. I also 
> would be happy to see some kind of 'server mode' in which the password does not get 
[...]

PR #416, opened 2018-05-05
Adding User prompt to enable keyring or not.

miphilomath:
This patch is referenced to Issue: #406. A user prompt is added to commandline.py  file. 
@lethliel @marcus-h 
Please review this patch. I have few doubts regarding the patch, if anyone can help: 
* Since config file is created using a template, so if the user choose to enable keyring, same config file template is used with the use_keyring commented which might be confusing. So, should we create another template for keyring enabled scenario?
* An error occurred when the keyring is enabled by the user, which is resolved by installing keyrings.alt package. If there is any better way, can anyone help me?
> Traceback (most recent call last):
  File "osc-wrapper.py", line 41, in <module>
    r = babysitter.run(osccli)
  File "/home/matrix/projects/opensuse/osc_test/osc/osc/babysitter.py", line 61, in run
    return prg.main(argv)
  File "/home/matrix/projects/opensuse/osc_test/osc/osc/cmdln.py", line 336, in main
    self.postoptparse()
  File "/home/matrix/projects/opensuse/osc_test/osc/osc/commandline.py", line 177, in postoptparse
    conf.write_initial_config(e.file, config)
  File "/home/matrix/projects/opensuse/osc_test/osc/osc/conf.py", line 747, in write_initial_config
    keyring.set_password(host, config['user'], config['pass'])
  File "/home/matrix/projects/opensuse/osc_test/osc/requirements/local/lib/python2.7/site-packages/keyring/core.py", line 47, in set_password
    _keyring_backend.set_password(service_name, username, password)
  File "/home/matrix/projects/opensuse/osc_test/osc/requirements/local/lib/python2.7/site-packages/keyring/backends/fail.py", line 23, in get_password
    raise RuntimeError(msg)
RuntimeError: No recommended backend was available. Install the keyrings.alt package if you want to use the non-recommended backends. See README.rst for details.

If I am missing something in the patch, please guide me on improving it.
Thanks

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -139,14 +139,35 @@ class Osc(cmdln.Cmdln):
         except oscerr.NoConfigfile as e:
             print(e.msg, file=sys.stderr)
             print('Creating osc configuration file %s ...' % e.file, file=sys.stderr)
+
             import getpass
+
             config = {}
             config['user'] = raw_input('Username: ')
+            keyring_choice = str(raw_input('Enable Keyring? (Default: No): '))
+
+            if keyring_choice in ['yes', 'y', 'Yes', 'Y']:
+                try:
+                    import keyring
+                    self.options.no_keyring = False
+                except:
+                    try:
+                        import gnomekeyring
+                        self.options.no_gnome_keyring = False
+                    except: 
+                        print("No keyring available!")
+            #else:
+            #    print("Are you sure you don't want to enable keyring?")
+
             config['pass'] = getpass.getpass()
             if self.options.no_keyring:
                 config['use_keyring'] = '0'
+            else:
+                config['use_keyring'] = '1'
             if self.options.no_gnome_keyring:
                 config['gnome_keyring'] = '0'
+            else:
+                config['gnome_keyring'] = '1'
             if self.options.apiurl:
                 config['apiurl'] = self.options.apiurl
 
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -144,30 +144,27 @@ class Osc(cmdln.Cmdln):
 
             config = {}
             config['user'] = raw_input('Username: ')
-            keyring_choice = str(raw_input('Enable Keyring? (Default: No): '))
 
-            if keyring_choice in ['yes', 'y', 'Yes', 'Y']:
-                try:
-                    import keyring
-                    self.options.no_keyring = False
-                except:
-                    try:
-                        import gnomekeyring
-                        self.options.no_gnome_keyring = False
-                    except: 
-                        print("No keyring available!")
-            #else:
-            #    print("Are you sure you don't want to enable keyring?")
+            # Check if the keyrings available on the system or not using conf.GENERIC_KEYRING and conf.GNOME_KEYRING 
+            # If yes, Ask the user to enable the keyring or not
+            # Set the 'use_keyring' or 'use_gnome_keyring' variable to 1 if user choose Yes else pass to use Defaults value
+            if conf.GENERIC_KEYRING or conf.GNOME_KEYRING:
+                keyring_choice = str(raw_input("Keyring utility is found on your system. Would you like to Enable keyring? (Default: No): "))
+                if keyring_choice.lower() in ['y', 'yes']:
+                    # Set the use_keyring variable
+                    if conf.GENERIC_KEYRING:
+                        config['use_keyring'] = '1'
+                    elif conf.GNOME_KEYRING:
+                        config['gnome_keyring'] = '1'
+            else:
+                print("\nWarning: No keyring available! Password will be saved in plain text.")
+                print(" To use keyring, please install python-keyring or gnome-keyring on your system and try again.\n")
 
             config['pass'] = getpass.getpass()
             if self.options.no_keyring:
                 config['use_keyring'] = '0'
-            else:
-                config['use_keyring'] = '1'
             if self.options.no_gnome_keyring:
                 config['gnome_keyring'] = '0'
-            else:
-                config['gnome_keyring'] = '1'
             if self.options.apiurl:
                 config['apiurl'] = self.options.apiurl
 
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -328,7 +328,7 @@ apiurl = %(apiurl)s
 #traceback = 1
 
 # use KDE/Gnome/MacOS/Windows keyring for credentials if available
-#use_keyring = 1
+use_keyring = %(use_keyring)s
 
 # check for unversioned/removed files before commit
 #check_filelist = 1


marcus-h:
First of all, thanks for the patch!

On 2018-05-05 19:00:39 +0000, Mohit Bansal wrote:
> Please review this patch. I have few doubts regarding the patch, if anyone can help: 
> * Since config file is created using a template, so if the user choose to enable keyring, same config file template is used with the use_keyring commented which might be confusing. So, should we create another template for keyring enabled scenario?

No, we could make this configurable in the template. Currently, we have
something like this in the template:

# use KDE/Gnome/MacOS/Windows keyring for credentials if available
#use_keyring = 1

which could be changed to

# use KDE/Gnome/MacOS/Windows keyring for credentials if available
use_keyring = %(use_keyring)s

(probably, we also need something similar for the gnome_keyring).

> * An error occurred when the keyring is enabled by the user, `Traceback (most recent call last):
>   File "osc-wrapper.py", line 41, in <module>
>     r = babysitter.run(osccli)
>   File "/home/matrix/projects/opensuse/osc_test/osc/osc/babysitter.py", line 61, in run
>     return prg.main(argv)
>   File "/home/matrix/projects/opensuse/osc_test/osc/osc/cmdln.py", line 336, in main
>     self.postoptparse()
>   File "/home/matrix/projects/opensuse/osc_test/osc/osc/commandline.py", line 177, in postoptparse
>     conf.write_initial_config(e.file, config)
>   File "/home/matrix/projects/opensuse/osc_test/osc/osc/conf.py", line 747, in write_initial_config
>     keyring.set_password(host, config['user'], config['pass'])
>   File "/home/matrix/projects/opensuse/osc_test/osc/requirements/local/lib/python2.7/site-packages/keyring/core.py", line 47, in set_password
>     _keyring_backend.set_password(service_name, username, password)
>   File "/home/matrix/projects/opensuse/osc_test/osc/requirements/local/lib/python2.7/site-packages/keyring/backends/fail.py", line 23, in get_password
>     raise RuntimeError(msg)
> RuntimeError: No recommended backend was available. Install the keyrings.alt package if you want to use the non-recommended backends. See README.rst for details.
> ` which is resolved by installing keyrings.alt package. If there is any better way, can anyone help me?
> 
Hmm it seems that you have no keyring backend installed. Maybe we should
only set conf.GENERIC_KEYRING to True, if there is at least one backend. If
I understand it correctly, installing a backend/backends fixed it?

Let's have a look at the patch itself:

> @@ -139,14 +139,35 @@ def postoptparse(self, try_again = True):
>          except oscerr.NoConfigfile as e:
>              print(e.msg, file=sys.stderr)
>              print('Creating osc configuration file %s ...' % e.file, file=sys.stderr)
> +
>              import getpass
> +
>              config = {}
>              config['user'] = raw_input('Username: ')
> +            keyring_choice = str(raw_input('Enable Keyring? (Default: No): '))
> +
> +            if keyring_choice in ['yes', 'y', 'Yes', 'Y']:
> +                try:
> +                    import keyring
> +                    self.options.no_keyring = False

Hmm it would be strange if self.options.no_keyring was True prio this
assignment (if it was True, then the user explicitly called
"osc --no-keyring...", but now decided to enable the keyring... maybe
we should error out in this case).

> +                except:
> +                    try:
> +                        import gnomekeyring
> +                        self.options.no_gnome_keyring = False

See above

> +                    except: 
> +                        print("No keyring available!")

In the general, the code above basically repeats some parts of the code we
already have in conf.py. If you want to check the presence of
keyring/gnome_keyring just use conf.GENERIC_KEYRING or conf.GNOME_KEYRING,
respectively.

> +            #else:
> +            #    print("Are you sure you don't want to enable keyring?")
> +

Hmm there is no needed to introduce a commented "else" branch:)

>              config['pass'] = getpass.getpass()
>              if self.options.no_keyring:
>                  config['use_keyring'] = '0'
> +            else:
> +                config['use_keyring'] = '1'

Hmm this is wrong:
- the keyring will be enabled, even if the user entered "no" during the
  "Enable Keyring?..." prompt
- the keyring will be enabled, even if no keyring package is installed

>              if self.options.no_gnome_keyring:
>                  config['gnome_keyring'] = '0'
> +            else:
> +                config['gnome_keyring'] = '1'

See above.

IMHO, the code should be roughly like this:
- check if a keyring is available
- if so, ask the user if she wants to use a keyring and _which_ one

philo-math:
Thanks @marcus-h for the detailed explanation.

philo-math:
> - if so, ask the user if she wants to use a keyring and _which_ one

Should I ask for the preferred keyring too? @marcus-h

marcus-h:
On 2018-05-06 10:49:44 +0000, Mohit Bansal wrote:
> > - if so, ask the user if she wants to use a keyring and _which_ one
> 
> Should I ask for the preferred keyring too? @marcus-h 
> 
Yes, that would be the most reasonable approach (IMHO).

philo-math:
@marcus-h I have a little doubt regarding that.
* How to check all available keyring other than GENERIC and GNOME KEYRING? I mean the user choice should be GENERIC and GNOME KEYRING or are there any other keyring that can be used? (Sorry for being so naive, I am recently introduced with the concept of keyring)
* Is the above updated patch correct other than the user choice for the keyring?

adrianschroeter:
On Samstag, 5. Mai 2018, 22:30:50 CEST wrote Marcus Hüwe:
...
> # use KDE/Gnome/MacOS/Windows keyring for credentials if available
> use_keyring = %(use_keyring)s
> 
> (probably, we also need something similar for the gnome_keyring).

A bit out of topic, but can't we drop gnome_keyring and rely on the gnome
implementation of python-keyring?

I can remember gnome_keyring only as problem child, causing all kinds
of wired bugreports esp. when it is not used on the right gnome version.
Or you had shared home directory but were loged-in into some other server
which did not have the gnome stuff installed.

any reason to keep it?


-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

marcus-h:
On 2018-05-07 14:52:54 +0000, Adrian Schröter wrote:
> On Samstag, 5. Mai 2018, 22:30:50 CEST wrote Marcus Hüwe:
> ...
> > # use KDE/Gnome/MacOS/Windows keyring for credentials if available
> > use_keyring = %(use_keyring)s
> > 
> > (probably, we also need something similar for the gnome_keyring).
> 
> A bit out of topic, but can't we drop gnome_keyring and rely on the gnome
> implementation of python-keyring?
> 
Hmm does something like that exist? (At least I did not find such a thing
in the tumbleweed repo (or I missed it:) )

> I can remember gnome_keyring only as problem child, causing all kinds
> of wired bugreports esp. when it is not used on the right gnome version.

Yes... the whole keyring stuff is (sometimes) quite "fragile":/

> Or you had shared home directory but were loged-in into some other server
> which did not have the gnome stuff installed.
> 
> any reason to keep it?
> 
If there's a gnome_keyring keyring backend, I'm happy to drop the
gnome_keyring code (Hmm we could also provide our own gnome_keyring
keyring backend (it shouldn't be too hard to implement...))

marcus-h:
On 2018-05-07 14:52:15 +0000, Mohit Bansal wrote:
> @marcus-h I have a little doubt regarding that.
> * How to check all available keyring other than GENERIC and GNOME KEYRING? I mean the user choice should be GENERIC and GNOME KEYRING or are there any other keyring that can be used? (Sorry for being so naive, I am recently introduced with the concept of keyring)

Yes, that's what I meant: the user could choose between the generic
keyring or gnome keyring. (Btw., you can also retrieve all available
keyring backends via keyring.backend.get_all_keyring().)

> * Is the above updated patch correct other than the user choice for the keyring?
> 
Yes, it should work in principle (the gnome_keyring config option
is missing, though - but see also @adrianS' comment). Btw., you could
also squash the commits into one commit (it makes the review a bit
easier).

marcus-h:
On 2018-05-07 22:41:53 +0200, Marcus Hüwe wrote:
> Yes, it should work in principle (the gnome_keyring config option
> is missing, though - but see also @adrianS' comment).

Gnaa... I meant @adrianschroeter.

lethliel:
@philo-math Thanks for this PR. 

This works like a charm and out of the box. Regarding to my tests we could drop all gnome_keyring specific code. 

This PR just makes it better

@adrianschroeter @marcus-h: Let us go with this. The gnome_keyring related code can be dropped later, but at the moment this PR would help a lot of people, because the keyring support is not working out of the box.

frispete:
> I can implement the user choice of keyring but I don't understand why do we need that?

There're a couple of reasons: the keyring package tries to prioritize the backend, e.g. preferring kwallet, when it detects a KDE. OTOH, kwallets depend on a viable dbus session, that might not be available within a SSH session!

BTW, I've created a [keyrings.cryptfile](https://pypi.org/project/keyrings.cryptfile) package, that implements a file based backend with strong cryptography (e.g. Argon2 KDF, and AES AEAD).

marcus-h:
On 2018-08-07 07:05:28 -0700, Marco Strigl wrote:
> @philo-math Thanks for this PR. 
> 
> This works like a charm and out of the box. Regarding to my tests we could drop all gnome_keyring specific code. 
> 
Hmm which package provides a keyring backend that uses the gnome_keyring?

> This PR just makes it better
> 
> @adrianschroeter @marcus-h: Let us go with this. The gnome_keyring related code can be dropped later, but at the moment this PR would help a lot of people, because the keyring support is not working out of the box. 
> 
I'm really hesitant here... I would prefer to do it "right", that is
listing all keyring backends and let the user make a conscious decision
which backend she wants to use (if she wants to use the keyring at all).

Note that this requires an additional change: we probably have to store
the chosen keyring backend in the config's apiurl section (alternatively,
we could set/modify the "default_backend" setting in the user's
keyringrc.cfg, but it could be changed by the user/other apps again and
its probably not nice to simply "change" the option's value...)

What about the following plan:

1) Phase out the gnome_keyring specific osc code:
For this, we need to wrap the gnome_keyring code into a keyring-like
interface/class (for instance a OscGnomeKeyringBackend or
OscGnomeKeyringKeyringBackend class). It is important that this class
does _not_ inherit from the KeyringBackend class, because it must also
work in the absense of python-keyring package.
Next, we introduce a get_keyring_backend() function that returns the
appropriate keyring (or keyring-like) backend according to our existing
rules:

if config['use_keyring'] and GENERIC_KEYRING:
    return <the default keyring backend>
elif config['gnome_keyring'] and GNOME_KEYRING:
    return OscGnomeKeyringBackend()
return None

Now, whenever we want to communicate with a keyring, we do something
like:

keyring_backend = get_keyring_backend()
if keyring_backend is not None:
    keyring_backend.set_password(url, user, pass)

etc.

This way we can get rid of the gnome_keyring specific codepaths and
neither loose functionality nor break the users' setups.


2) Introduce a CredentialsManager abstraction (could be optional)
In essence, what we want to do is to manage credentials. After step
1), we still distinguish codewise between "creds are stored in the
keyring" and "creds are stored in the config file". There is no reason
for this distinction. Hence, let's introduce a credentials manager
abstraction, which looks roughly like this:

def get_creds_manager(conffile):
    keyring_backend = get_keyring_backend():
    if keyring_backend is not None:
        return keyring_backend
    return ConfigFileBasedCredentialsManager(conffile)

To store the password, we simply do something like this:

creds_mgr = get_creds_manager(conffile)
creds_mgr.set_password(url, user, pass)

This would simplify the conf module a lot!
(Of course, we would need to work out the concrete details of
the CredentialsManager interface).


3) Prompt the user for a credentials manager
That's what this PR is actually supposed to do for the keyring backends.
Now, the goal is that the user can select the preferred credentials manager
when entering the credentials for a specific apiurl for the first time (of
course, we need to store the chosen credentials manager in the config's
apiurl section).


Each of these steps can be individual PR.


Neat side effect: we can support @frispete's keyrings.cryptfile backend
out of the box.


What do you guys think?

lethliel:
On 08/08/2018 02:37 PM, Marcus Hüwe  wrote:
> On 2018-08-07 07:05:28 -0700, Marco Strigl wrote:
>> @philo-math Thanks for this PR.
>>
>> This works like a charm and out of the box. Regarding to my tests we could drop all gnome_keyring specific code.
>>
> Hmm which package provides a keyring backend that uses the gnome_keyring?

SecretStorage provides the backend for gnome-keyring. 

> 
>> This PR just makes it better
>>
>> @adrianschroeter @marcus-h: Let us go with this. The gnome_keyring related code can be dropped later, but at the moment this PR would help a lot of people, because the keyring support is not working out of the box.
>>
> I'm really hesitant here... I would prefer to do it "right", that is
> listing all keyring backends and let the user make a conscious decision
> which backend she wants to use (if she wants to use the keyring at all).
> 
> Note that this requires an additional change: we probably have to store
> the chosen keyring backend in the config's apiurl section (alternatively,
> we could set/modify the "default_backend" setting in the user's
> keyringrc.cfg, but it could be changed by the user/other apps again and
> its probably not nice to simply "change" the option's value...)
> 
> What about the following plan:
> 
> 1) Phase out the gnome_keyring specific osc code:
> For this, we need to wrap the gnome_keyring code into a keyring-like
> interface/class (for instance a OscGnomeKeyringBackend or
> OscGnomeKeyringKeyringBackend class). It is important that this class
> does _not_ inherit from the KeyringBackend class, because it must also
> work in the absense of python-keyring package.
> Next, we introduce a get_keyring_backend() function that returns the
> appropriate keyring (or keyring-like) backend according to our existing
> rules:
> 
> if config['use_keyring'] and GENERIC_KEYRING:
> return <the default keyring backend>
> elif config['gnome_keyring'] and GNOME_KEYRING:
> return OscGnomeKeyringBackend()
> return None
> 
> Now, whenever we want to communicate with a keyring, we do something
> like:
> 
> keyring_backend = get_keyring_backend()
> if keyring_backend is not None:
> keyring_backend.set_password(url, user, pass)
> 
> etc.
> 
> This way we can get rid of the gnome_keyring specific codepaths and
> neither loose functionality nor break the users' setups.
> 

You can just set the backend with: 

gnome-keyring:	keyring.set_keyring(keyring.backends.kwallet.DBusKeyring())
kwallet:	keyring.set_keyring(keyring.backends.SecretService.Keyring())

So I do not see the urge to create classes. 

> 
> 2) Introduce a CredentialsManager abstraction (could be optional)
> In essence, what we want to do is to manage credentials. After step
> 1), we still distinguish codewise between "creds are stored in the
> keyring" and "creds are stored in the config file". There is no reason
> for this distinction. Hence, let's introduce a credentials manager
> abstraction, which looks roughly like this:
> 
> def get_creds_manager(conffile):
> keyring_backend = get_keyring_backend():
> if keyring_backend is not None:
> return keyring_backend
> return ConfigFileBasedCredentialsManager(conffile)
> 
> To store the password, we simply do something like this:
> 
> creds_mgr = get_creds_manager(conffile)
> creds_mgr.set_password(url, user, pass)
> 
> This would simplify the conf module a lot!
> (Of course, we would need to work out the concrete details of
> the CredentialsManager interfac
This would be nice. So we would have:

1. keyring
2. pain text pass in config file 
3. encoded pass in config file

In addition to that there are currently discussions:

4. not storing the password at all
5. storing the password in /var/run/user/$id/oscpw (tmpfs)

> 
> 
> 3) Prompt the user for a credentials manager
> That's what this PR is actually supposed to do for the keyring backends.
> Now, the goal is that the user can select the preferred credentials manager
> when entering the credentials for a specific apiurl for the first time (of
> course, we need to store the chosen credentials manager in the config's
> apiurl section).

Yes.

> 
> 
> Each of these steps can be individual PR.
> 
> 
> Neat side effect: we can support @frispete's keyrings.cryptfile backend
> out of the box.
> 
> 
> What do you guys think?
> 
>

I would say:

1. remove gnome-keyring code and map gnome_keyring cases to
keyring functions. Through that we will not break the configs of the users
using gnome_keyring.

2. and then the rest ;)

marcus-h:
On 2018-08-08 13:03:58 +0000, Marco Strigl wrote:
> On 08/08/2018 02:37 PM, Marcus Hüwe  wrote:
> > On 2018-08-07 07:05:28 -0700, Marco Strigl wrote:
> >> @philo-math Thanks for this PR.
> >>
> >> This works like a charm and out of the box. Regarding to my tests we could drop all gnome_keyring specific code.
> >>
> > Hmm which package provides a keyring backend that uses the gnome_keyring?
> 
> SecretStorage provides the backend for gnome-keyring. 
> 
Ah ok, thanks for the pointer!

<SNIP>

> > if config['use_keyring'] and GENERIC_KEYRING:
> > return <the default keyring backend>
> > elif config['gnome_keyring'] and GNOME_KEYRING:
> > return OscGnomeKeyringBackend()
> > return None
> > 
> > Now, whenever we want to communicate with a keyring, we do something
> > like:
> > 
> > keyring_backend = get_keyring_backend()
> > if keyring_backend is not None:
> > keyring_backend.set_password(url, user, pass)
> > 
> > etc.
> > 
> > This way we can get rid of the gnome_keyring specific codepaths and
> > neither loose functionality nor break the users' setups.
> > 
> 
> You can just set the backend with: 
> 
> gnome-keyring:	keyring.set_keyring(keyring.backends.kwallet.DBusKeyring())
> kwallet:	keyring.set_keyring(keyring.backends.SecretService.Keyring())
> 
> So I do not see the urge to create classes. 
> 
We would need the OscGnomeKeyring keyring-like backend for backward
compatibility: an existing osc installation could already be using
gnome_keyring without having python2-SecretStorage installed.

> > 2) Introduce a CredentialsManager abstraction (could be optional)
> > In essence, what we want to do is to manage credentials. After step
> > 1), we still distinguish codewise between "creds are stored in the
> > keyring" and "creds are stored in the config file". There is no reason
> > for this distinction. Hence, let's introduce a credentials manager
> > abstraction, which looks roughly like this:
> > 
> > def get_creds_manager(conffile):
> > keyring_backend = get_keyring_backend():
> > if keyring_backend is not None:
> > return keyring_backend
> > return ConfigFileBasedCredentialsManager(conffile)
> > 
> > To store the password, we simply do something like this:
> > 
> > creds_mgr = get_creds_manager(conffile)
> > creds_mgr.set_password(url, user, pass)
> > 
> > This would simplify the conf module a lot!
> > (Of course, we would need to work out the concrete details of
> > the CredentialsManager interfac
> This would be nice. So we would have:
> 
> 1. keyring
> 2. pain text pass in config file 
> 3. encoded pass in config file
> 
> In addition to that there are currently discussions:
> 
> 4. not storing the password at all
> 5. storing the password in /var/run/user/$id/oscpw (tmpfs)
> 
They could be easily added via the new framework. In a later step,
we could also use a plugin mechanism so that "external" credential
managers could be registered.

lethliel:
On 08/08/2018 11:37 PM, Marcus Hüwe  wrote:
> On 2018-08-08 13:03:58 +0000, Marco Strigl wrote:
>> On 08/08/2018 02:37 PM, Marcus Hüwe wrote:

[SNIP]

>>
>> So I do not see the urge to create classes.
>>
> We would need the OscGnomeKeyring keyring-like backend for backward
> compatibility: an existing osc installation could already be using
> gnome_keyring without having python2-SecretStorage installed.
> 

If you have python-keyring you (should) have python-SecretStorage.
python-keyring requires python-SecretStorage. 


This would be the hard cut (similar as we do it with M2Crypto):

try:
    import keyring
    GENERIC_KEYRING = True
except ImportError as e:
    raise ImportError('python-keyring is not installed. Please install python-keyring.')

But I agree that we should make the switch as smooth as possible for the users. 


I would still print out a warning that gnomekeyring is deprecated and we suggest to 
use keyring instead. 


>> > 2) Introduce a CredentialsManager abstraction (could be optional)
>> > In essence, what we want to do is to manage credentials. After step
>> > 1), we still distinguish codewise between "creds are stored in the
>> > keyring" and "creds are stored in the config file". There is no reason
>> > for this distinction. Hence, let's introduce a credentials manager
>> > abstraction, which looks roughly like this:
>> >
>> > def get_creds_manager(conffile):
>> > keyring_backend = get_keyring_backend():
>> > if keyring_backend is not None:
>> > return keyring_backend
>> > return ConfigFileBasedCredentialsManager(conffile)
>> >
>> > To store the password, we simply do something like this:
>> >
>> > creds_mgr = get_creds_manager(conffile)
>> > creds_mgr.set_password(url, user, pass)
>> >
>> > This would simplify the conf module a lot!
>> > (Of course, we would need to work out the concrete details of
>> > the CredentialsManager interfac
>> This would be nice. So we would have:
>>
>> 1. keyring
>> 2. pain text pass in config file
>> 3. encoded pass in config file
>>
>> In addition to that there are currently discussions:
>>
>> 4. not storing the password at all
>> 5. storing the password in /var/run/user/$id/oscpw (tmpfs)
>>
> They could be easily added via the new framework. In a later step,
> we could also use a plugin mechanism so that "external" credential
> managers could be registered.
> 

Yes :) Then I see the need of classes.

OscGnomeKeyringBackend()
OscWalletKeyringBackend()
OscFilePWBackend(encrypted=true|false)
OscFilePWTmpfs()
[ OscYouNameIt() ]


@marcus-h, @philo-math: 
Should I start implementing this? Or does anyone else insist to work on this?

marcus-h:
On 2018-08-08 23:51:04 -0700, Marco Strigl wrote:
> On 08/08/2018 11:37 PM, Marcus Hüwe  wrote:
> > On 2018-08-08 13:03:58 +0000, Marco Strigl wrote:
> >> On 08/08/2018 02:37 PM, Marcus Hüwe wrote:
> >>
> >> So I do not see the urge to create classes.
> >>
> > We would need the OscGnomeKeyring keyring-like backend for backward
> > compatibility: an existing osc installation could already be using
> > gnome_keyring without having python2-SecretStorage installed.
> > 
> 
> If you have python-keyring you (should) have python-SecretStorage.
> python-keyring requires python-SecretStorage. 
> 
Yes, but currently our gnome keyring support does not depend on
python-keyring - that's why we still need to OscGnomeKeyring
keyring-like backend (Or maybe in our new parlance, a gnome keyring
credentials manager that uses the gnome keyring python bindings).

<SNIP>

> >> > This would simplify the conf module a lot!
> >> > (Of course, we would need to work out the concrete details of
> >> > the CredentialsManager interfac
> >> This would be nice. So we would have:
> >>
> >> 1. keyring
> >> 2. pain text pass in config file
> >> 3. encoded pass in config file
> >>
> >> In addition to that there are currently discussions:
> >>
> >> 4. not storing the password at all
> >> 5. storing the password in /var/run/user/$id/oscpw (tmpfs)
> >>
> > They could be easily added via the new framework. In a later step,
> > we could also use a plugin mechanism so that "external" credential
> > managers could be registered.
> > 
> 
> Yes :) Then I see the need of classes.
> 
> OscGnomeKeyringBackend()
> OscWalletKeyringBackend()

Hmm we do not need an "explicit" class for each keyring backend. I think
we could use some class hierarchy that looks like this:

class AbstractCredentialsManager(object): pass

class KeyringBasedCredentialsManager(AbstractCredentialsManager)

(The methods are omitted)

Then, we could do something like

creds_mgrs = []
for backend in keyring.backend.get_all_keyring():
    creds_mgrs.append(KeyringBasedCredentialsManager(backend))

That is, there is one KeyringBasedCredentialsManager class that works with
all keyring backends.

> @marcus-h, @philo-math: 
> Should I start implementing this? Or does anyone else insist to work on this? 
> 
If you feel like it, go ahead:)
@philo-math: If/Once we arrive at step 3, you could update this PR (if you
are still interested in it:) ).

philo-math:
Thanks @marcus-h @lethliel to review the PR. 

I will definitely update this PR as per the needs later :)
Thanks again @lethliel for the update :)

marcus-h:
github ignored my mail again...
```----- Forwarded message from Marcus Hüwe <suse-tux@gmx.de> -----
  
Date: Tue, 3 Sep 2019 12:20:49 +0200
To: reply+00104a4c7a8f2064b7de1687b9de11912a52f0fb6d5f867a92cf000000011781775c92a169ce131af334@reply.github.com
Subject: Re: [openSUSE/osc] Adding User prompt to enable keyring or not. (#416)
User-Agent: NeoMutt/20180716

On 2018-08-07 15:06:04 +0000, Hans-Peter Jansen wrote:
<SNIP>
> BTW, I've created a [keyrings.cryptfile](https://pypi.org/project/keyrings.cryptfile) package, that implements a file based backend with strong cryptography (e.g. Argon2 KDF, and AES AEAD). 
> 
Last week, we rewrote osc's password handling code (for instance,
see commit eb3a3ef0ec20de7015ad5ba35af090ca15518569 ("Introduction of
new credential management")). If your keyrings.cryptfile backend is
installed (and available via python-keyring), osc will automatically
offer it.
Hence, we can close this PR.

```

PR #419, opened 2018-05-16
attempt to handle missing/empty keyring passwords

frispete:
- improve user/password query
- catch keyring.get_password exceptions
- retry on empty/missing passwords

improve keyring migration scenarios

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -170,13 +170,7 @@ class Osc(cmdln.Cmdln):
     def query_user_and_password(self, url):
         import getpass
         user = raw_input('Username for %s [^C to exit]: ' % url)
-        # empty user/password have a special meaning,
-        # we cannot do these checks here :-(
-        #if not user:
-        #    sys.exit('No user. Exiting...')
         passwd = getpass.getpass()
-        #if not passwd:
-        #    sys.exit('No password. Exiting...')
         return user, passwd
 
 
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -139,10 +139,9 @@ class Osc(cmdln.Cmdln):
         except oscerr.NoConfigfile as e:
             print(e.msg, file=sys.stderr)
             print('Creating osc configuration file %s ...' % e.file, file=sys.stderr)
-            import getpass
             config = {}
-            config['user'] = raw_input('Username: ')
-            config['pass'] = getpass.getpass()
+            _, host, _ = conf.parse_apisrv_url(None, self.options.apiurl)
+            config['user'], config['pass'] = self.query_user_and_password(host)
             if self.options.no_keyring:
                 config['use_keyring'] = '0'
             if self.options.no_gnome_keyring:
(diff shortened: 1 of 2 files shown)


marcus-h:
First of all thanks for the PR! I just had a very quick glance
at the patch. Some comments below.

On 2018-05-16 03:32:13 -0700, Hans-Peter Jansen wrote:
> index 87f8a4fd..ee4310a9 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py

> @@ -170,6 +167,19 @@ def postoptparse(self, try_again = True):
>              self.download_progress = TextMeter(hide_finished=True)
>  
>  
> +    def query_user_and_password(self, url):
> +        import getpass
> +        user = raw_input('Username for %s [^C to exit]: ' % url)
> +        # empty user/password have a special meaning,
> +        # we cannot do these checks here :-(
> +        #if not user:
> +        #    sys.exit('No user. Exiting...')
> +        passwd = getpass.getpass()
> +        #if not passwd:
> +        #    sys.exit('No password. Exiting...')
> +        return user, passwd

Hmm why do you think empty user/password have a special meaning? You
are probably referring to the code in conf.get_config, right? We
should probably treat an empty user as an error and an empty password
is debatable (IMHO). The best is to check the api code to see if an
empty user/password is allowed (and if not, when this changed).
Alternatively, keep the patch as it is, but remove the commented
code:)

> index abff0194..efbce6e0 100644
> --- a/osc/conf.py
> +++ b/osc/conf.py
> @@ -892,10 +892,18 @@ def get_config(override_conffile=None,
>              try:
>                  # Read from keyring lib if available
>                  user = cp.get(url, 'user', raw=True)
> -                password = str(keyring.get_password(host, user))
> -            except:
> +                password = keyring.get_password(host, user)
> +            except Exception as e:
> +                print('get_password failed for \'%s@%s\': %s' % (user, host, e), file=sys.stderr)
>                  # Fallback to file based auth.
> -                pass
> +            else:
> +                if password is None:
> +                    # password missing for this user on this host, offer user/password query
> +                    raise oscerr.ConfigMissingApiurl('missing credentials for apiurl: \'%s\'' % apiurl,
> +                                                     conffile, apiurl)
> +                else:
> +                    # isn't the password always a str already?
> +                    password = str(password)

Hmm we should probably use the same logic in the gnome-keyring case
as well. The str conversion is done to avoid some unicode issues
(see commit 9561a99d6ba62d3fcfce262e359ba31ce78e2fa7).

frispete:
> Aternatively, keep the patch as it is, but remove the commented code:)

Will take that route (later today). :-)

> Hmm we should probably use the same logic in the gnome-keyring case as well. The str conversion is done to avoid some unicode issues (see commit 9561a99).

Yes, the idea was to get feedback on the approach and test it. So far, it allowed me to migrate from plain passwords to kwallet easily. I still don't like the changed pw case, and noticed some warnings, if osc was used from within ssh connections. 

In fact, shouldn't we think about a gnome-keyring deprecation, given that keyring should handle the prefered Gnome keyring SecretService just well? The code would profit, of course, but I don't know, why/if users prefer the gnome-keyring for some reason (and the least thing I want is nagging users...).

marcus-h:
```----- Forwarded message from Marcus Hüwe <suse-tux@gmx.de> -----
  
Date: Tue, 3 Sep 2019 12:31:35 +0200
To: reply+00104a4c5ec029abb03cf87b82eb7292688b81e997296eec92cf000000011713caad92a169ce13492d49@reply.github.com
Subject: Re: [openSUSE/osc] attempt to handle missing/empty keyring passwords (#419)
User-Agent: NeoMutt/20180716

Since commit cba4b58bbe5e46326bb96eef571f0758c74eb537 ("Improve
password handling"), osc handles missing credentials more gracefully
(it will ask for the credentials). So I think we can close this PR.

----- End forwarded message -----```

PR #422, opened 2018-05-25
Setuptools

mcepl:
With this I get some additional subcommands to ``setup.py``, namely ``develop``, which is the same as ``install``, except it just links to the repository, not copies. It is therefore possible to run the version from the checkout as normal command.

Why there is ``osc-py3-wrapper.py`` and not just ``osc`` command?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,11 +2,13 @@
 
 from distutils.core import setup
 import distutils.core
-import distutils.command.build
-import distutils.command.install_data
+from distutils.command import build, install_data
 import os.path
 import osc.core
 import sys
+
+import setuptools
+
 from osc import commandline
 from osc import babysitter
 # optional support for py2exe
@@ -17,7 +19,7 @@ except:
     HAVE_PY2EXE = False
 
 
-class build_osc(distutils.command.build.build, object):
+class build_osc(build.build, object):
     """
     Custom build command which generates man page.
     """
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2018-05-25 07:49:19 -0700, Matěj Cepl wrote:
> With this I get some additional subcommands to ``setup.py``, namely ``develop``, which is the same as ``install``, except it just links to the repository, not copies. It is therefore possible to run the version from the checkout as normal command.
> 
Hmm I don't see a new "develop" command (when looking at the diff [1]).
Also, if I try to merge to it it to the master branch I get several
conflicts. Is this based on @lethliel's python3 branch? (Just a guess
from the commit summary).

Btw., hi @mcepl and thanks for the PR:)

[1] https://patch-diff.githubusercontent.com/raw/openSUSE/osc/pull/422.patch

mcepl:
Run
```
python setup.py --help-commands
```

mcepl:
> Is this based on @lethliel's python3 branch? (Just a guess from the commit summary).

See above, it is based on opensuse/python3.

setup.py
```diff
@@ -2,11 +2,13 @@
 
 from distutils.core import setup
```
Conan-Kudo: Replace this with `from setuptools import setup`

setup.py
```diff
@@ -2,11 +2,13 @@
...
 import distutils.core
-import distutils.command.build
-import distutils.command.install_data
+from distutils.command import build, install_data
```
Conan-Kudo: IIRC, these are all in `setuptools, too. Switch to it completely rather than using both.

marcus-h:
On 2018-05-26 09:57:02 -0700, Matěj Cepl wrote:
> > Is this based on @lethliel's python3 branch? (Just a guess from the commit summary).
> 
> See above, it is based on opensuse/python3.
> 
Ah, you changed it afterwards. Since github does not send a
notification for it, I won't notice it (the same applies to edited
comments:/ ).

mcepl:
So?

marcus-h:
On 2018-07-02 08:50:36 -0700, Matěj Cepl wrote:
> So?
> 
The python3 branch is @lethliel's baby:)
So it is up to him.

Issue #398, opened 2018-04-02
SIGSEGV on "osc commit"


z3ntu:
with fish:
```
 /m/h/h/razergenie  osc commit
Sending    _service
Sending    razergenie.dsc
Sending    razergenie.spec
Deleting    razergenie_0.6-1.debian.tar.xz
Deleting    razergenie_0.6.orig.tar.xz
Sending    razergenie_0.7-1.debian.tar.xz
Sending    razergenie_0.7.orig.tar.xz
fish: “osc commit” terminated by signal SIGSEGV (Address boundary error)
```
with bash:
```
[luca@g550jk razergenie]$ osc commit
Log message not specified
a)bort, c)ontinue, e)dit: c
Sending    _service
Sending    razergenie.dsc
Sending    razergenie.spec
Deleting    razergenie_0.6-1.debian.tar.xz
Deleting    razergenie_0.6.orig.tar.xz
Sending    razergenie_0.7-1.debian.tar.xz
Sending    razergenie_0.7.orig.tar.xz
Transmitting file data .....
Committed revision 16.
Waiting for server side source service run
..............double free or corruption (!prev)
Aborted (core dumped)
```
osc version is 0.162.1 and python2 2.7.14-2 (Arch Linux)

sometimes the upload succeeds and the build runs serverside

z3ntu:
gdb backtrace for `/usr/bin/python2 /usr/bin/osc build openSUSE_Leap_42.3 x86_64`: https://pastebin.com/raw/NyLW7icn

z3ntu:
After running this multiple times, I get different results but all end in a crash...
```
 /m/h/h/razergenie  osc commit                                                                                                                                                      Mon 02 Apr 2018 16:04:21 CEST
Sending    razergenie.spec
Transmitting file data .fish: “osc commit” terminated by signal SIGSEGV (Address boundary error)
 !  /m/h/h/razergenie  osc commit                                                                                                                                                  Mon 02 Apr 2018 16:04:31 CEST
Log message not specified
a)bort, c)ontinue, e)dit: c
corrupted double-linked list
fish: “osc commit” terminated by signal SIGABRT (Abort)
 !  /m/h/h/razergenie  osc commit                                                                                                                                                  Mon 02 Apr 2018 16:04:36 CEST
Log message not specified
a)bort, c)ontinue, e)dit: c
Sending    razergenie.spec
fish: “osc commit” terminated by signal SIGSEGV (Address boundary error)
[...]

PR #424, opened 2018-05-27
Disable ssl session resumption

gojpdchx:
As this patch is currently not present in the `python3` branch, those who are using M2Crypto 0.30 may encounter problems similar to issue #398.


lethliel:
I already cheery-picked it from master. Thank you for pointing this out.

Issue #427, opened 2018-06-08
M2Crypto excepts hostname as bytes not string 


mimi1vx:
```
e "/usr/lib/python2.7/site-packages/osc/core.py", line 3355, in http_GET
    def http_GET(*args, **kwargs):    return http_request('GET', *args, **kwargs)
  File "/usr/lib/python2.7/site-packages/osc/core.py", line 3344, in http_request
    fd = urlopen(req, data=data)
  File "/usr/lib64/python2.7/urllib2.py", line 154, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python2.7/urllib2.py", line 429, in open
    response = self._open(req, data)
  File "/usr/lib64/python2.7/urllib2.py", line 447, in _open
    '_open', req)
  File "/usr/lib64/python2.7/urllib2.py", line 407, in _call_chain
    result = func(*args)
  File "/usr/lib/python2.7/site-packages/osc/oscssl.py", line 220, in https_open
    h.request(req.get_method(), selector, req.data, headers)
  File "/usr/lib64/python2.7/httplib.py", line 1042, in request
    self._send_request(method, url, body, headers)
  File "/usr/lib64/python2.7/httplib.py", line 1082, in _send_request
    self.endheaders(body)
  File "/usr/lib64/python2.7/httplib.py", line 1038, in endheaders
    self._send_output(message_body)
  File "/usr/lib64/python2.7/httplib.py", line 882, in _send_output
    self.send(msg)
  File "/usr/lib64/python2.7/httplib.py", line 844, in send
    self.connect()
  File "/usr/lib/python2.7/site-packages/osc/oscssl.py", line 281, in connect
    connected = self._connect(addrinfo[0])
  File "/usr/lib/python2.7/site-packages/osc/oscssl.py", line 268, in _connect
    self.sock.set_tlsext_host_name(self.host)
  File "/usr/lib64/python2.7/site-packages/M2Crypto/SSL/Connection.py", line 674, in set_tlsext_host_name
    m2.ssl_set_tlsext_host_name(self.ssl, name)
TypeError: in method 'ssl_set_tlsext_host_name', argument 2 of type 'char const *'
```

mimi1vx:
```
rpm -qiv osc
Name        : osc
Version     : 0.162.1
Release     : 1.1
Architecture: noarch
Install Date: Wed 13 Dec 2017 02:03:33 PM CET
Group       : Development/Tools/Other
Size        : 1910316
[...]

PR #428, opened 2018-06-08
M2Crypto requires bytes for sock.set_tlsext_host_name

mimi1vx:
Fixes problem when osc passes string/ unicode to `sock.set_tlsext_host_name` requires `bytes`

### DIFF
--- a/osc/oscssl.py
+++ b/osc/oscssl.py
@@ -254,13 +254,18 @@ class myHTTPSConnection(M2Crypto.httpslib.HTTPSConnection):
         elif family != socket.AF_INET:
             # old SSL.Connection classes implicitly use socket.AF_INET
             return False
-
+        
+        if not isinstance(self.host, bytes):
+            host = self.host.encode()
+        else:
+            host = self.host
+        
         self.sock = SSL.Connection(self.ssl_ctx, **kwargs)
         if self.session:
             self.sock.set_session(self.session)
         if hasattr(self.sock, 'set_tlsext_host_name'):
-            self.sock.set_tlsext_host_name(self.host)
-        self.sock.connect((self.host, self.port))
+            self.sock.set_tlsext_host_name(host)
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2018-06-08 03:16:06 -0700, Ondřej Súkup wrote:
> Fixes problem when osc passes string/ unicode to `sock.set_tlsext_host_name` requires `bytes`

Hmm the str type works for me using M2Crypto 0.29. Which M2Crypto
version are you using? What's the type of self.host? (it should be
str)

marcus-h:
On 2018-06-08 05:26:10 -0700, Ondřej Súkup wrote:
> not `str` but `bytes` , on Python2 str == bytes but with python3 this will fail.
> 
Yes, but osc is inherently not py3 ready. @lethliel is working on
py3 support, though.

> my problem in issue #427 comes from this:
> 
> ```
> >>> from  osc.core import makeurl
> >>> makeurl("https://api.suse.de", ["comments", "project", u"QA:Head"], query=None)
> makeurl: https://api.suse.de ['comments', 'project', u'QA:Head'] None
> u'https://api.suse.de/comments/project/QA:Head'
> ```
> 
Ah it is about a custom script (that was not obvious from the
incomplete stacktrace in #427).

> so url is parsed as `unicode` type on python2 and passed also to this M2Crypto function... 
> 
Ok. We should do the unicode to bytes conversion earlier (at least
not in the oscssl.py) or it is the responsibility of the caller....

Why do you explicitly pass a unicode object in the first place?

mimi1vx:
This is pretty bad: 
>>>Yes, but osc is inherently not py3 ready. 


>> Why do you explicitly pass a unicode object in the first place?

because I read data passed to this from third source , which sends part of data as unicode strings ..
( and , my shame I forgot check and sanitize this data)

dmach:

PR #430, opened 2018-06-25
Support larger uids and gids

seankelly:


marcus-h:
On 2018-06-25 13:18:38 -0700, Sean Kelly wrote:
> -- Commit Summary --
> 
>   * Support larger uids and gids
> 
Hmm what about completely dropping the upper bound? At least
I see no reason why we should limit it (since uid_t is system
dependent). Maybe something like

buildidre = re.compile('^[0-9]+:[0-9]+$')

What do you think?

seankelly:
Sounds good to me. Pushed a new commit with no upper limit.

marcus-h:
On 2018-06-25 13:56:53 -0700, Sean Kelly wrote:
> Sounds good to me. Pushed a new commit with no upper limit.
> 
Great. Merged. Thanks a lot for the PR!

PR #432, opened 2018-07-19
checkout / get deleted package files

lethliel:
* new command getdeleted (restore)
  requires arguments project / package / destionation

  Examples:
  ```
  osc restore home:mstrigl foo deleted_foo
  osc restore -b home:mstrigl foo
  ```

* altered command do_checkout
  new option -D | --deleted. 
  Can only be used with PACKAGE and needs -o to work

* what happens:
  core function get_deleted_sources fetches the file list of the deleted
  package and saves it to given destdir. If destdir is not existent it
  creates the destination directory. Otherwise the files will be written
  in the existing directory.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -4280,6 +4280,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
     @cmdln.option('-e', '--expand-link', action='store_true',
                         help='if a package is a link, check out the expanded '
                              'sources (no-op, since this became the default)')
+    @cmdln.option('-D', '--deleted', action='store_true',
+                        help='checkout an already deleted package. No meta information ')
     @cmdln.option('-u', '--unexpand-link', action='store_true',
                         help='if a package is a link, check out the _link file ' \
                              'instead of the expanded sources')
@@ -4364,6 +4366,12 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                 project_dir = os.curdir
                 package = args[0]
 
+        if opts.deleted and package:
+            if not opts.output_dir:
+                raise oscerr.WrongOptions('-o | --output-dir is needed to get deleted sources')
+        elif opts.deleted and not package:
+            raise oscerr.WrongOptions('-D | --deleted can only be used with a package')
+
         rev, dummy = parseRevisionOption(opts.revision)
         if rev == None:
             rev = "latest"
@@ -4379,14 +4387,17 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             get_source_file(apiurl, project, package, filename, revision=rev, progress_obj=self.download_progress)
 
         elif package:
-            if opts.current_dir:
-                project_dir = None
-            checkout_package(apiurl, project, package, rev, expand_link=expand_link, \
-                             prj_dir=project_dir, service_files = opts.source_service_files, \
-                             server_service_files=opts.server_side_source_service_files, \
-                             progress_obj=self.download_progress, size_limit=opts.limit_size, \
-                             meta=opts.meta, outdir=opts.output_dir)
-            print_request_list(apiurl, project, package)
+            if opts.deleted:
+                checkout_deleted_package(apiurl, project, package, opts.output_dir)
+            else:
+                if opts.current_dir:
+                    project_dir = None
+                checkout_package(apiurl, project, package, rev, expand_link=expand_link, \
+                                 prj_dir=project_dir, service_files = opts.source_service_files, \
+                                 server_service_files=opts.server_side_source_service_files, \
+                                 progress_obj=self.download_progress, size_limit=opts.limit_size, \
+                                 meta=opts.meta, outdir=opts.output_dir)
+                print_request_list(apiurl, project, package)
 
         elif project:
             prj_dir = opts.output_dir if opts.output_dir else project
--- a/osc/core.py
+++ b/osc/core.py
@@ -7763,4 +7763,24 @@ def get_rpmlint_log(apiurl, proj, pkg, repo, arch):
     f = http_GET(u)
     return f.read()
 
+def checkout_deleted_package(apiurl, proj, pkg, dst):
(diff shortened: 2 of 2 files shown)


lethliel:
@marcus-h please review

lethliel:
On 07/19/2018 02:27 PM, Adrian Schröter  wrote:
> On Donnerstag, 19. Juli 2018, 13:42:12 CEST wrote Marco Strigl:
>> * new command getdeleted (restore)
>> requires arguments project / package / destionation
>>
>> Examples:
>> ```
>> osc restore home:mstrigl foo deleted_foo
>> osc restore -b home:mstrigl foo
>> ```
> 
> I am not a friend of yet another command ...
> 
> checkout, cat and less can already retrieve single files.
> Can't we just support
> 
> osc co -D home:mstrigl foo [deleted_foo]
> 
> ?

Of course. Just thought that an extra command can be enriched in the future. 
Perhaps to check out with meta information (if still available). But the 
core function can be enriched too. 

So:

I have no strong opinion here. If Marcus sees it the same way I am outvoted ;)

marcus-h:
For this PR my email didn't make it through (again, for whatever reason...)

```

Date: Tue, 14 Aug 2018 01:16:16 +0200
To: reply+00104a4c47f272c8d077e845e3a836a70fbcf64f6802dbcb92cf0000000117683b1392a169ce146d1565@reply.github.com
Subject: Re: [openSUSE/osc] checkout / get deleted package files (#432)

On 2018-07-19 11:42:12 +0000, Marco Strigl wrote:
> index 87f8a4fd..b84f56a1 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -4280,6 +4280,8 @@ def do_repourls(self, subcmd, opts, *args):
>      @cmdln.option('-e', '--expand-link', action='store_true',
>                          help='if a package is a link, check out the expanded '
>                               'sources (no-op, since this became the default)')
> +    @cmdln.option('-D', '--deleted', action='store_true',
> +                        help='checkout a already deleted package. No meta information ')

"an already..."

>      @cmdln.option('-u', '--unexpand-link', action='store_true',
>                          help='if a package is a link, check out the _link file ' \
>                               'instead of the expanded sources')
> @@ -4364,6 +4366,12 @@ def do_checkout(self, subcmd, opts, *args):
>                  project_dir = os.curdir
>                  package = args[0]
>  
> +        if opts.deleted and package:
> +            if not opts.output_dir:
> +                raise oscerr.WrongOptions('-o | --output_dir is needed to get deleted sources')

s/output_dir/output-dir/

What about --expand-link/--unexpand-link in combination with -D?

> @@ -4443,6 +4454,51 @@ def do_checkout(self, subcmd, opts, *args):
>              raise oscerr.WrongArgs('Missing argument.\n\n' \
>                    + self.get_cmd_help('checkout'))
>  
> +    @cmdln.alias('restore')
> +    @cmdln.option('-d', '--destdir', metavar='DIR',
> +                  help='destination directory')
> +    def do_getdeleted(self, subcmd, opts, *args):
> +        """${cmd_name}: get sources of a deleted project
> +
> +        Restores the sources of a deleted package into <destdir>.
> +        Only the sources are restored. There will be no meta information
> +
> +        Usage:
> +            osc getdeleted [-d <destdir> PROJECT PACKAGE [destdir]
> +
> +        Examples:
> +            osc getdeleted my:project foo /home/bar/deleted_foo
> +            osc getdeleted -d /home/bar/deleted_foo my:project foo
> +

Hmm the UI looks really "strange", IMHO.
I would say let's stick with osc co -D... for now.

> index 34a625fa..d87182e0 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -7763,4 +7763,25 @@ def get_rpmlint_log(apiurl, proj, pkg, repo, arch):
>      f = http_GET(u)
>      return f.read()
>  
> +def get_deleted_sources(apiurl, proj, pkg, dst):

Hmm what about calling it checkout_package_deleted or
checkout_deleted_package?

(Optional: we could add a rev parameter in the future)

> +    pl = meta_get_filelist(apiurl, proj, pkg, deleted=True)
> +    query = {}
> +    query['deleted'] = 1
> +
> +    if os.path.isdir(dst):
> +        print('Restoring in existing directory %s' % dst)
> +    else:
> +        print('Creating %s' % dst)
> +        os.mkdir(dst)
> +
> +    for filename in pl:
> +        print('Restoring %s to %s' % (filename, dst))
> +        full_file_path = dst + '/' + filename

Hmm just use os.path.join(dst, filename)

> +        u = makeurl(apiurl, ['source', proj, pkg, filename], query=query)
> +        o = open(full_file_path, 'wb')
> +        for data in streamfile(u):
> +            o.write(data)
> +        o.close()

Hmm what about a "with" statement here:

with open(full_file_path, 'wb') as f:
    ...

```

lethliel:
@marcus-h: revised based on your comments. 
We don't need to restore sources from a link.

marcus-h:
On 2018-10-17 06:01:57 -0700, Marco Strigl wrote:
> @marcus-h: revised based on your comments. 
> We don't need to restore sources from a link. 
> 
From a quick glance: LGTM!

> +    else:
> +        print('Creating %s' % dst)
> +        os.mkdir(dst)

Maybe os.makedirs(dst)? For instance, if someone passes
"--output-dir does/not/exist".

lethliel:
> + else: + print('Creating %s' % dst) + os.mkdir(dst)
> Maybe os.makedirs(dst)? For instance, if someone passes "--output-dir does/not/exist".

You are right. That is better. Switched to os.makedirs(dst) now. 
@marcus-h: Please review again ;)

marcus-h:
On 2018-10-18 07:01:07 +0000, Marco Strigl wrote:
> > + else: + print('Creating %s' % dst) + os.mkdir(dst)
> > Maybe os.makedirs(dst)? For instance, if someone passes "--output-dir does/not/exist".
> 
> You are right. That is better. Switched to os.makedirs(dst) now. 
> @marcus-h: Please review again ;) 
> 
Merged. Thanks!

Issue #435, opened 2018-08-06
osc meta attribute openSUSE:Factory -a ATTR -e -> no edit option


DimStar77:
According to osc --help, one should be able to edit an attribute using meta attribute -e;

This though results in a NO-OP, no editor showing up.

Calling the command ```osc -d meta attribute openSUSE:Factory -a OSRT:Config -e``` does not show a single API call being done

lethliel:
This is not that easy. 

-e results in a PUT that uploads the resulting XML to the server. 

But for attributes there is no PUT route, because every attribute can have different 
permissions (and thus is completely different to other meta files like project meta)

What I can do is add a edit mode for single attributes. 

Like:
`osc meta attribute home:mstrigl -e -a OBS:RejectRequests`

And then a editor shows up for just this attribute. 

And from the help: 

```
Usage:

    [SNIP]

    osc meta attribute PRJ [PKG [SUBPACKAGE]] [--attribute ATTRIBUTE] [--create [--set <value_list>]|--delete|--set <value_list>]

```

DimStar77:
> What I can do is add a edit mode for single attributes.
>
>Like:
>osc meta attribute home:mstrigl -e -a OBS:RejectRequests

Single attribute editing would be helpful in this case, indeed

SchoolGuy:
I got hit by this today since the Python code for this is directly in the CLI and not accessible through the Python API. Thus I need to duplicate this code now into openSUSE-release-tools. It would really be nice if [this](https://github.com/openSUSE/open-build-service/issues/7839) will be implemented soon. I can't write Ruby, so I sadly can't contribute this, however I could fix `osc` if the server side code is existing.

PR #436, opened 2018-08-06
edit mode for single attributes

lethliel:
This will ease the creation of multiline values for attributes.

--edit and --set or --create can not be used together

Approach to issue https://github.com/openSUSE/osc/issues/435

@marcus-h: please review

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -860,7 +860,7 @@ class Osc(cmdln.Cmdln):
             osc meta <prj|pkg|prjconf|user|group|pattern> [-m|--message TEXT] -e|--edit ARGS...
             osc meta <prj|pkg|prjconf|user|group|pattern> [-m|--message TEXT] -F|--file ARGS...
             osc meta pattern --delete PRJ PATTERN
-            osc meta attribute PRJ [PKG [SUBPACKAGE]] [--attribute ATTRIBUTE] [--create [--set <value_list>]|--delete|--set <value_list>]
+            osc meta attribute PRJ [PKG [SUBPACKAGE]] [--attribute ATTRIBUTE] [--create [--set <value_list>]|--delete|--set <value_list>|--edit]
         ${cmd_option_list}
         """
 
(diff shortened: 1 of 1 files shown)


lethliel:
I rebased this PR onto master. 

@marcus-h: Can you please have a look. Otherwise this will get merged on June, 14th

marcus-h:
github ignored my mail again...

```
  
Date: Thu, 13 Jun 2019 21:57:54 +0200
To: reply+00104a4c1f15feaa8a96307df4338f384d4c085b5022464892cf000000011780145c92a169ce14bd0851@reply.github.com
Subject: Re: [openSUSE/osc] edit mode for single attributes (#436)
User-Agent: NeoMutt/20180716

On 2018-08-06 06:51:24 -0700, Marco Strigl wrote:
> index 17d8fe1d..2b370d55 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -1043,6 +1045,26 @@ def do_meta(self, subcmd, opts, *args):
>              url = makeurl(apiurl, attributepath)
>              for data in streamfile(url, http_POST, data=d):
>                  sys.stdout.write(decode_it(data))
> +        elif opts.edit and cmd == 'attribute':
> +            if not opts.attribute:
> +                raise oscerr.WrongOptions('no attribute given')
> +            aname = opts.attribute.split(":")
> +            if len(aname) != 2:
> +                raise oscerr.WrongOptions('Given attribute is not in "NAMESPACE:NAME" style')
> +            old_value = ''.join(show_attribute_meta(apiurl, project, package, subpackage, opts.attribute, opts.attribute_defaults, opts.attribute_project))
> +            root = ET.fromstring(old_value)
> +            attribs = root.find('attribute')
> +            text = '<value></value>'
> +            if attribs:
> +                text = ''
> +                for ovalue in attribs.getiterator():
> +                    if ovalue.tag == 'value':
> +                        text += '<value>%s</value>' % ovalue.text
> +            values = edit_message(template=text)

Hmm two remarks:
- why don't we present the complete xml to the user?
- using edit_message is a bit misleading if the user did not modify
  the file (it will talk about a missing log message etc.). Maybe it
  is easier if we add support for the attribute stuff to the edit_meta
  code? (I thought about this a while ago and IIRC there was something
  special about attributes (need to think about it again).)

> +            d = '<attributes><attribute namespace=\'%s\' name=\'%s\' >%s</attribute></attributes>' % (aname[0], aname[1], values)
> +            url = makeurl(apiurl, attributepath)
> +            for data in streamfile(url, http_POST, data=d):
> +                sys.stdout.write(data)
> 

```

lethliel:
@marcus-h 
The API will be changed so that attributes are handled the same way as prj or pkg meta. 
So this PR is obsolete and can be closed. 

handled in https://github.com/openSUSE/open-build-service/issues/7839

PR #437, opened 2018-08-07
cleanup xpath. submit/* is ignored by the API since 2012

lethliel:
remove submit/* parts from xpath. They are ignored anyway by the API since commit https://github.com/openSUSE/open-build-service/commit/b502316e28c256d4f3213c1bea09d7ed52f9a4c4

as discussed in https://github.com/openSUSE/open-build-service/issues/5571

So the xpath becomes a little bit cleaner and easier ti understand / debug. 

@marcus-h: Please review

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -4309,12 +4309,10 @@ def get_review_list(apiurl, project='', package='', byuser='', bygroup='', bypro
     if package:
         todo['package'] = package
     for kind, val in todo.items():
-        xpath_base = 'action/target/@%(kind)s=\'%(val)s\' or ' \
-                     'submit/target/@%(kind)s=\'%(val)s\''
+        xpath_base = 'action/target/@%(kind)s=\'%(val)s\''
 
         if conf.config['include_request_from_project']:
-            xpath_base = xpath_join(xpath_base, 'action/source/@%(kind)s=\'%(val)s\' or ' \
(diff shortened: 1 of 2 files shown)


adrianschroeter:
the submit/* is still supported in the xpath api, because we are compatible;)

however, I think it is fair to define a minimum required OBS version and that code can be removed then most likely. But I would like to define that somewhere so that can reapply the rule in other code places.

My current proposal would be to define OBS 2.4 (2013, 5 years is somewhat how long a SP can live) as minimum requirement.... but we should check what that means regarding cleanup.

marcus-h:
Hmm github ignored my message again. Forwarding...
```
  
Date: Thu, 30 Aug 2018 13:46:36 +0200
To: reply+00104a4c86bf0b3868af337aac6cd6041bd514578d8357e692cf0000000117812bac92a169ce14c1baec@reply.github.com
Subject: Re: [openSUSE/osc] cleanup xpath. submit/* is ignored by the API since 2012 (#437)

On 2018-08-07 02:43:08 -0700, Adrian Schröter wrote:
> however, I think it is fair to define a minimum required OBS version and that code can be removed then most likely. But I would like to define that somewhere so that can reapply the rule in other code places.
> 
> My current proposal would be to define OBS 2.4 (2013, 5 years is somewhat how long a SP can live) as minimum requirement.... but we should check what that means regarding cleanup.
> 
2.4 sounds fine...
@adrianschroeter: do you have any objections against merging this PR?

```

PR #438, opened 2018-08-12
Clarify use of osc token

andreas-schwab:
Clarify the fact that the parameters of osc token -t and osc token -d are
of different types.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -720,7 +720,7 @@ class Osc(cmdln.Cmdln):
                         help='Create a new token')
     @cmdln.option('-d', '--delete', metavar='TOKENID',
                         help='Create a new token')
-    @cmdln.option('-t', '--trigger', metavar='TOKENID',
+    @cmdln.option('-t', '--trigger', metavar='TOKENSTRING',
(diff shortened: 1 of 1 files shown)


lethliel:
LTGM. Thanks for the patch.

marcus-h:
Thanks github for eating my message again...

```
  
Date: Tue, 14 Aug 2018 00:26:44 +0200
To: reply+00104a4cadec890b4c385eb0b61ee5bd956c607ff1f7f2e892cf000000011787deeb92a169ce14d99d3c@reply.github.com
Subject: Re: [openSUSE/osc] Clarify use of osc token (#438)

On 2018-08-12 04:41:31 -0700, Andreas Schwab wrote:
> Clarify the fact that the parameters of osc token -t and osc token -d are
> of different types.

Hmm I'm not sure if I understand this correctly, but you are talking about
"osc trigger -t 'TOKENID?project=foo&package=bar'" vs.
"osc trigger -d TOKENID", right?

If so, yes you are right (and we could/should fix the description). The
question is if this kind of usage was intended at all... :/

```

@andreas-schwab: was that ^^ your intention?

PR #410, opened 2018-04-27
remove UrlGrabber to enable python3 

lethliel:
This pull request removes urlgrabber from osc. 
The mirror and grabber functions were moved to grabber.py

The progressbar function from Urlgrabber was replaced by the module progressbar.

@marcus-h: we discussed the changes already. 

This is the first of many PR to enable python3 in osc.

PR #442, opened 2018-09-03
Fix Python 3.7 support

z3ntu:
See PEP 479

### DIFF
--- a/osc/OscConfigParser.py
+++ b/osc/OscConfigParser.py
@@ -87,7 +87,6 @@ class ConfigLineOrder:
         for line in self._lines:
             if line.type == 'section':
                 yield line.name
-        raise StopIteration()
 
 class Line:
     """Base class for all line objects"""
@@ -160,7 +159,6 @@ class SectionLine(Line):
     def __iter__(self):
         for line in self._lines:
             yield line.name
-        raise StopIteration()
 
 
 class CommentLine(Line):


adrianschroeter:
why the return? just removing the exception would have the same effect, no?

z3ntu:
It's what the PEP says (`If raise StopIteration occurs directly in a generator, simply replace it with return.`)

marcus-h:
On 2018-09-03 22:35:28 -0700, Adrian Schröter wrote:
> why the return? just removing the exception would have the same effect, no?
> 
Yep, that's the preferred solution:)

@z3ntu: since the python3 branch is @lethliel's baby, let's wait
until he is back so that he can merge/review/comment on your PR.

nickbroon:
I would be great to see the python3 branch merged into master soon. What's the state of the python3 port?

lethliel:
@nickbroon I am currently on vacation in Italy. I will be back next week and then there will be progress again..

mcepl:
@lethliel Welcome back from vacations ;) I have been using ``python3`` branch for couple of months without any problems. If not merging outright (which I would prefer, of course), could we at least get the branch rebased on the top the ``master``, please?

lethliel:
@mcepl @nickbroon I am back from vacation and I am working on the python3 port again. 
Before anything else can be merged into the master this PR has to be merged: https://github.com/openSUSE/osc/pull/410

After that I will create the pull requests for each affected module and discuss the changes with Marcus. 

@mcepl: I will merge master into this branch during this week.

lethliel:
And for this PR I would say just remove the returns. (even if PEP suggests them)

z3ntu:
Found the part you mean:

> The return is necessary for a strictly-equivalent translation, though in this particular file, there is no further code, and the return can be omitted.

And updated the PR.

lethliel:
LGTM. Thanks for the contribution.

PR #443, opened 2018-09-07
make 'osc copypac --keep-link' description easier to understand

johanneskastl:
Make the --keep-link documentation for `osc copypac` easier to understand.

See [this mail thread](https://lists.opensuse.org/opensuse-buildservice/2018-09/msg00004.html) for my initial question and Marcus very helpful explanation .

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -2880,7 +2880,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
     @cmdln.option('-k', '--keep-maintainers', action='store_true',
                         help='keep original maintainers. Default is remove all and replace with the one calling the script.')
     @cmdln.option('-K', '--keep-link', action='store_true',
-                        help='keep the source link in target, this also expands the source')
+                        help='If the target package is a link, the link is kept, but may be updated. If the source package is a link, its expanded version is considered.')
     @cmdln.option('-d', '--keep-develproject', action='store_true',
                         help='keep develproject tag in the package metadata')
     @cmdln.option('-r', '--revision', metavar='rev',


marcus-h:
On 2018-09-07 14:56:09 +0000, Johannes Kastl wrote:
> index 0bd1a939..9778061c 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -2880,7 +2880,7 @@ def do_aggregatepac(self, subcmd, opts, *args):
>      @cmdln.option('-k', '--keep-maintainers', action='store_true',
>                          help='keep original maintainers. Default is remove all and replace with the one calling the script.')
>      @cmdln.option('-K', '--keep-link', action='store_true',
> -                        help='keep the source link in target, this also expands the source')
> +                        help='Only useful if the target already exists. If the target is a link or branch, keep the source link in target. If not, the target will have expanded sources and will not be a link.')

Hmm the new description is a bit misleading. What about

"If the target package is a link, the link is kept. The source package is
always expanded."

?

johanneskastl:
Hi Marcus,

> If the target package is a link, the link is kept. The source package is always expanded.

I like the first part. But the second is difficult IMHO. This reads like the **source package** is being touched and modified to no longer be a link or something along those lines.

How about:
"Only useful if the target is already existing, otherwise it behaves like the -e option.
If the target is a link, the link is kept, but the link revision may be updated. The sources in the source package are being copied over in their expanded form."

Or:
"Only useful if the target is already existing, otherwise it behaves like the -e option.
If the target is a link, the link is kept, but the link revision may be updated. The expanded sources will be copied from the source package to the target package."

Doesn't it copy over the diff with all its files, so the target package ends up with the exact same modifications as the source package?

marcus-h:
On 2018-09-10 12:35:18 -0700, Johannes Kastl wrote:
> > If the target package is a link, the link is kept. The source package is always expanded.
> 
> I like the first part. But the second is difficult IMHO. This reads like the **source package** is being touched and modified to no longer be a link or something along those lines.
> 
Ok, agreed:)

> How about:
> "Only useful if the target is already existing, otherwise it behaves like the -e option.
> If the target is a link, the link is kept, but the link revision may be updated. The sources in the source package are being copied over in their expanded form."
> 
> Or:
> "Only useful if the target is already existing, otherwise it behaves like the -e option.
> If the target is a link, the link is kept, but the link revision may be updated. The expanded sources will be copied from the source package to the target package."
> 
Hmm both variants are too long (IMHO) and "the link revision may be updated"
part is too implemention specific: in case of a branch, the baserev might
be updated (baserev != link revision (the "rev" attribute)). In case of
a non-branch link, the resulting _link could be "significantly" different.

Let's try to find a precise and concise description:)

What about

"If the target package is a link, the link is kept. If the source package
is a link, its expanded version is considered."

?

> Doesn't it copy over the diff with all its files, so the target package ends up with the exact same modifications as the source package?
> 
Hmm "which" diff?

johanneskastl:
After discussing with Marcus, I reworked the description:

> "If the target package is a link, the link is kept but may be updated.
> If the source package is a link, its expanded version is considered."

marcus-h:
On 2018-09-17 11:51:35 -0700, Johannes Kastl wrote:
> After discussing with Marcus, I reworked the description:
> 
> > "If the target package is a link, the link is kept but may be updated.
> > If the source package is a link, its expanded version is considered."
> 
Merged. Thanks a lot!

PR #444, opened 2018-09-21
chroot: mount /dev/shm

bmwiedemann:
otherwise programs like 'gyp' fail to work
that need it for interprocess communication


adrianschroeter:
The build script is supporting to jump into shell now (works at least for chroot and kvm). IMHO we should replace the code from this command entirely with that (and fix stuff there in case)

adrianschroeter:
we use now the build script for jumping into chroot

Issue #445, opened 2018-09-25
osc sr ignores commandline options


olafhering:
Creating a submit request in some project, targeted for another. The intention on the commandline is clear, the target API host is not the one in `.osc/_apiurl`:

```
$ ibs sr  SUSE:SLE-12-SP4:GA
created request id 637890
```
Actual result: [https://build.opensuse.org/request/show/637890](url)

Expected result: some error that current pkg can not directly operate with `-A $url`
```
$  cat $(type -P ibs)
#/bin/sh
export OSC_CONFIG=$HOME/.osc/oscrc
exec /usr/bin/osc --config=$OSC_CONFIG -A https://api.suse.de "$@"
```

marcus-h:
On 2018-09-25 09:12:13 +0000, Olaf Hering wrote:
> Creating a submit request in some project, targeted for another. The intention on the commandline is clear, the target API host is not the one in `.osc/_apiurl`:
> 
> ```
> $ ibs sr  SUSE:SLE-12-SP4:GA
> created request id 637890
> ```
> Actual result: [https://build.opensuse.org/request/show/637890](url)
> 
> Expected result: some error that current pkg can not directly operate with `-A $url`

Hmm no that's the "expected" behavior because the "sr" command is
(partially) context sensitive. That is, if cwd is a project or
package working copy, it will try to retrieve the information
(including the apiurl) from the working copy.
The same holds for other context sensitive commands. For instance,
"cd pkg_wc; osc -A xyz ci" will commit to the api that is specified
in pkg_wc.

PR #446, opened 2018-09-25
print error and exit if p.apiurl != -A <apiurl> using osc sr

lethliel:
Print error and exit if `osc -A <apiurl> sr` is issued on a
package with a different apiurl in its working copy. 

This leads to unexpected behavior, because the SR is created for the package
apiurl, not the apiurl given by the -A command line option.

Fixes https://github.com/openSUSE/osc/issues/445

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -1311,6 +1311,10 @@ class Osc(cmdln.Cmdln):
             p = findpacs(os.curdir)[0]
             src_project = p.prjname
             src_package = p.name
+            if self.options.apiurl and self.options.apiurl != p.apiurl:
+                print('The apiurl for the working copy of this package is %s' % p.apiurl)
+                print('You cannot use this command with the -A %s option.' % self.options.apiurl)
+                sys.exit(1)
             apiurl = p.apiurl
             if len(args) == 0 and p.islink():
                 dst_project = p.linkinfo.project


marcus-h:
On 2018-09-25 04:54:21 -0700, Marco Strigl wrote:
> Print error and exit if `osc -A <apiurl> sr` is issued on a
> package with a different apiurl in its working copy. 
> 
> This leads to unexpected behavior, because the SR is created for the package
> apiurl, not the apiurl given by the -A command line option.
> 
I'm not really sure if we should add such a check. If we do it, we probably
have to "fix" a lot other commands as well. Passing the -A option does not
imply that osc only talks to this apiurl - it just overrides the config's
default apiurl, which is used if no apiurl can be retrieved, for instance,
from the package wc (or if the command is not context sensitive).

lethliel:
On 9/25/18 2:51 PM, Marcus Hüwe  wrote:
> On 2018-09-25 04:54:21 -0700, Marco Strigl wrote:
>> Print error and exit if `osc -A <apiurl> sr` is issued on a
>> package with a different apiurl in its working copy.
>>
>> This leads to unexpected behavior, because the SR is created for the package
>> apiurl, not the apiurl given by the -A command line option.
>>
> I'm not really sure if we should add such a check. If we do it, we probably
> have to "fix" a lot other commands as well. Passing the -A option does not
> imply that osc only talks to this apiurl - it just overrides the config's
> default apiurl, which is used if no apiurl can be retrieved, for instance,
> from the package wc (or if the command is not context sensitive).
> 

It is confusing. I would expect osc to *always* use the apiurl specified with -A.
In the help it says: 

"-A URL/alias, --apiurl=URL/alias
 specify URL to access API server at or an alias"

which gives no hint that in some cases the apiurl of the working copy is used
instead of the one given with -A <apiurl>.

So I would assume that if a user specifies the -A option he wants to work on 
another apiurl. And thus checking and throwing an error is completely ok. (IMHO)

marcus-h:
On 2018-09-25 05:58:42 -0700, Marco Strigl wrote:
> On 9/25/18 2:51 PM, Marcus Hüwe  wrote:
> > On 2018-09-25 04:54:21 -0700, Marco Strigl wrote:
> >> Print error and exit if `osc -A <apiurl> sr` is issued on a
> >> package with a different apiurl in its working copy.
> >>
> >> This leads to unexpected behavior, because the SR is created for the package
> >> apiurl, not the apiurl given by the -A command line option.
> >>
> > I'm not really sure if we should add such a check. If we do it, we probably
> > have to "fix" a lot other commands as well. Passing the -A option does not
> > imply that osc only talks to this apiurl - it just overrides the config's
> > default apiurl, which is used if no apiurl can be retrieved, for instance,
> > from the package wc (or if the command is not context sensitive).
> > 
> 
> It is confusing. I would expect osc to *always* use the apiurl specified with -A.
> In the help it says: 
> 
> "-A URL/alias, --apiurl=URL/alias
>  specify URL to access API server at or an alias"
> 
> which gives no hint that in some cases the apiurl of the working copy is used
> instead of the one given with -A <apiurl>.
> 
Hmm maybe we should improve the help text.

> So I would assume that if a user specifies the -A option he wants to work on 
> another apiurl.

That's not the semantics of the -A option.

lethliel:
On 9/25/18 6:04 PM, Marcus Hüwe  wrote:
> On 2018-09-25 05:58:42 -0700, Marco Strigl wrote:
>> On 9/25/18 2:51 PM, Marcus Hüwe wrote:
>> > On 2018-09-25 04:54:21 -0700, Marco Strigl wrote:
>> >> Print error and exit if `osc -A <apiurl> sr` is issued on a
>> >> package with a different apiurl in its working copy.
>> >>
>> >> This leads to unexpected behavior, because the SR is created for
> the package
>> >> apiurl, not the apiurl given by the -A command line option.
>> >>
>> > I'm not really sure if we should add such a check. If we do it, we
> probably
>> > have to "fix" a lot other commands as well. Passing the -A option
> does not
>> > imply that osc only talks to this apiurl - it just overrides the
> config's
>> > default apiurl, which is used if no apiurl can be retrieved, for
> instance,
>> > from the package wc (or if the command is not context sensitive).
>> >
>>
>> It is confusing. I would expect osc to *always* use the apiurl
> specified with -A.
>> In the help it says:
>>
>> "-A URL/alias, --apiurl=URL/alias
>> specify URL to access API server at or an alias"
>>
>> which gives no hint that in some cases the apiurl of the working copy
> is used
>> instead of the one given with -A <apiurl>.
>>
> Hmm maybe we should improve the help text.
> 
>> So I would assume that if a user specifies the -A option he wants to
> work on
>> another apiurl.
> 
> That's not the semantics of the -A option.
> 

huh?
What it is then?

marcus-h:
On 2018-09-25 13:02:24 -0700, Marco Strigl wrote:
> On 9/25/18 6:04 PM, Marcus Hüwe  wrote:
> > On 2018-09-25 05:58:42 -0700, Marco Strigl wrote:
> >> On 9/25/18 2:51 PM, Marcus Hüwe wrote:
> >> > On 2018-09-25 04:54:21 -0700, Marco Strigl wrote:
> >> >> Print error and exit if `osc -A <apiurl> sr` is issued on a
> >> >> package with a different apiurl in its working copy.
> >> >>
> >> >> This leads to unexpected behavior, because the SR is created for
> > the package
> >> >> apiurl, not the apiurl given by the -A command line option.
> >> >>
> >> > I'm not really sure if we should add such a check. If we do it, we
> > probably
> >> > have to "fix" a lot other commands as well. Passing the -A option
> > does not
> >> > imply that osc only talks to this apiurl - it just overrides the
> > config's
> >> > default apiurl, which is used if no apiurl can be retrieved, for
> > instance,
> >> > from the package wc (or if the command is not context sensitive).
> >> >
> >>
> >> It is confusing. I would expect osc to *always* use the apiurl
> > specified with -A.
> >> In the help it says:
> >>
> >> "-A URL/alias, --apiurl=URL/alias
> >> specify URL to access API server at or an alias"
> >>
> >> which gives no hint that in some cases the apiurl of the working copy
> > is used
> >> instead of the one given with -A <apiurl>.
> >>
> > Hmm maybe we should improve the help text.
> > 
> >> So I would assume that if a user specifies the -A option he wants to
> > work on
> >> another apiurl.
> > 
> > That's not the semantics of the -A option.
> > 
> 
> huh?
> What it is then?
> 
As I wrote above: "-A xyz" just overrides the config's default apiurl. That
is, conf.config['apiurl'] is set to xyz/the corresponding url. This does
not mean that osc uses this url for all requests - it is only used if the
osc code accesses conf.config['apiurl'].

lethliel:
On 9/25/18 10:57 PM, Marcus Hüwe  wrote:
> On 2018-09-25 13:02:24 -0700, Marco Strigl wrote:

[SNIP]

>> >
>> >> So I would assume that if a user specifies the -A option he wants to
>> > work on
>> >> another apiurl.
>> >
>> > That's not the semantics of the -A option.
>> >
>>
>> huh?
>> What it is then?
>>
> As I wrote above: "-A xyz" just overrides the config's default apiurl. That
> is, conf.config['apiurl'] is set to xyz/the corresponding url. This does
> not mean that osc uses this url for all requests - it is only used if the
> osc code accesses conf.config['apiurl'].
> 

Yes. The problem is that the user cannot know which command accesses conf.config['apiurl'].
This can lead to a quite confusing behavior. 

For example: 

Assuming we are in a package working copy 'helloworld' and we have two APIs (api_1 and api_2).
The working copy of 'helloworld' has api_1 in its _apiurl file. 

osc -A api_2 buildlog tries to fetch the information from api_2. If the package exists 
on api_2 the information is retrieved. If not a 404 is thrown.

osc -A api_2 sr TARGET:PROJECT creates a request on api_1.

I think there are some commands where -A makes absolutely no sense and we should the user
inform about it. (preventing actions the user did not want to do)

marcus-h:
On 2018-09-26 02:27:09 -0700, Marco Strigl wrote:
> On 9/25/18 10:57 PM, Marcus Hüwe  wrote:
> > On 2018-09-25 13:02:24 -0700, Marco Strigl wrote:
> >> >
> >> >> So I would assume that if a user specifies the -A option he wants to
> >> > work on
> >> >> another apiurl.
> >> >
> >> > That's not the semantics of the -A option.
> >> >
> >>
> >> huh?
> >> What it is then?
> >>
> > As I wrote above: "-A xyz" just overrides the config's default apiurl. That
> > is, conf.config['apiurl'] is set to xyz/the corresponding url. This does
> > not mean that osc uses this url for all requests - it is only used if the
> > osc code accesses conf.config['apiurl'].
> > 
> 
(or, in case of commandline.py, if Osc.get_api_url() is used)

> Yes. The problem is that the user cannot know which command accesses conf.config['apiurl'].
> This can lead to a quite confusing behavior. 
> 
Yep.

> For example: 
> 
> Assuming we are in a package working copy 'helloworld' and we have two APIs (api_1 and api_2).
> The working copy of 'helloworld' has api_1 in its _apiurl file. 
> 
> osc -A api_2 buildlog tries to fetch the information from api_2. If the package exists 
> on api_2 the information is retrieved. If not a 404 is thrown.
> 
> osc -A api_2 sr TARGET:PROJECT creates a request on api_1.
> 
> I think there are some commands where -A makes absolutely no sense and we should the user
> inform about it. (preventing actions the user did not want to do)
> 
This nicely demonstrates the "inconsistencies" of the current UI.
Now, the question is which command does the "wrong" thing?:)

IMHO, both implementations are reasonable. The fundamental issue is
that they are inconsistent...

adrianschroeter:
On Mittwoch, 26. September 2018, 13:17:40 CEST wrote Marcus Hüwe:
> On 2018-09-26 02:27:09 -0700, Marco Strigl wrote:
..
> > For example: 
> > 
> > Assuming we are in a package working copy 'helloworld' and we have two APIs (api_1 and api_2).
> > The working copy of 'helloworld' has api_1 in its _apiurl file. 
> > 
> > osc -A api_2 buildlog tries to fetch the information from api_2. If the package exists 
> > on api_2 the information is retrieved. If not a 404 is thrown.
> > 
> > osc -A api_2 sr TARGET:PROJECT creates a request on api_1.
> > 
> > I think there are some commands where -A makes absolutely no sense and we should the user
> > inform about it. (preventing actions the user did not want to do)
> > 
> This nicely demonstrates the "inconsistencies" of the current UI.
> Now, the question is which command does the "wrong" thing?:)
> 
> IMHO, both implementations are reasonable. The fundamental issue is
> that they are inconsistent...

My understanding is that an explicit -A is enforced. This makes sense
to me under the perspective of security and privacy, when I want to
be sure that the modifications happen on specific server.

I wouldn't add it to CLI if I would not mean to talk only to that server ...

-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

marcus-h:
On 2018-09-26 05:57:12 -0700, Adrian Schröter wrote:
> On Mittwoch, 26. September 2018, 13:17:40 CEST wrote Marcus Hüwe:
> > On 2018-09-26 02:27:09 -0700, Marco Strigl wrote:
> ..
> > > For example: 
> > > 
> > > Assuming we are in a package working copy 'helloworld' and we have two APIs (api_1 and api_2).
> > > The working copy of 'helloworld' has api_1 in its _apiurl file. 
> > > 
> > > osc -A api_2 buildlog tries to fetch the information from api_2. If the package exists 
> > > on api_2 the information is retrieved. If not a 404 is thrown.
> > > 
> > > osc -A api_2 sr TARGET:PROJECT creates a request on api_1.
> > > 
> > > I think there are some commands where -A makes absolutely no sense and we should the user
> > > inform about it. (preventing actions the user did not want to do)
> > > 
> > This nicely demonstrates the "inconsistencies" of the current UI.
> > Now, the question is which command does the "wrong" thing?:)
> > 
> > IMHO, both implementations are reasonable. The fundamental issue is
> > that they are inconsistent...
> 
> My understanding is that an explicit -A is enforced.

Currently, this varies from command to command.

> This makes sense
> to me under the perspective of security and privacy, when I want to
> be sure that the modifications happen on specific server.
> 

I'm a bit worried about WCs - for example

osc -A url1 co prj/pkg1
osc -A url2 co prj/pkg2
cd prj/pkg2
...
osc -A url1 ci

Technically, it can be implemented without effort, but from a user's
POV it is a quite "drastic" change IMHO (I'm not sure if there are
people out there who rely on the current behavior in the above
scenario (ignoring the -A option))

So the question is: should we change/break existing semantics (and
get a consistent ui) or live with the inconsistencies?

lethliel:
On 9/27/18 3:38 AM, Marcus Hüwe  wrote:
> On 2018-09-26 05:57:12 -0700, Adrian Schröter wrote:
>> On Mittwoch, 26. September 2018, 13:17:40 CEST wrote Marcus Hüwe:
>> > On 2018-09-26 02:27:09 -0700, Marco Strigl wrote:
>> ..
>> > > For example:
>> > >
>> > > Assuming we are in a package working copy 'helloworld' and we have two APIs (api_1 and api_2).
>> > > The working copy of 'helloworld' has api_1 in its _apiurl file.
>> > >
>> > > osc -A api_2 buildlog tries to fetch the information from api_2. If the package exists
>> > > on api_2 the information is retrieved. If not a 404 is thrown.
>> > >
>> > > osc -A api_2 sr TARGET:PROJECT creates a request on api_1.
>> > >
>> > > I think there are some commands where -A makes absolutely no sense and we should the user
>> > > inform about it. (preventing actions the user did not want to do)
>> > >
>> > This nicely demonstrates the "inconsistencies" of the current UI.
>> > Now, the question is which command does the "wrong" thing?:)
>> >
>> > IMHO, both implementations are reasonable. The fundamental issue is
>> > that they are inconsistent...
>>
>> My understanding is that an explicit -A is enforced.
> 
> Currently, this varies from command to command.
> 
>> This makes sense
>> to me under the perspective of security and privacy, when I want to
>> be sure that the modifications happen on specific server.
>>
> 
> I'm a bit worried about WCs - for example
> 
> osc -A url1 co prj/pkg1
> osc -A url2 co prj/pkg2
> cd prj/pkg2
> ...
> osc -A url1 ci
> 
> Technically, it can be implemented without effort, but from a user's
> POV it is a quite "drastic" change IMHO (I'm not sure if there are
> people out there who rely on the current behavior in the above
> scenario (ignoring the -A option))
> 
> So the question is: should we change/break existing semantics (and
> get a consistent ui) or live with the inconsistencies?
> 

I would go with an consistent UI. As a user, if you have no clue about the internal 
functions of osc or/and the historical development of osc, it is hard to guess that 
-A is not honored always and some commands will ignore this option. 

Perhaps we can go with a warning first and ask the opensuse buildervice mailinglist
about opinions?

olafhering:
The code does not work as documented:
````
     -A URL/alias, --apiurl=URL/alias
                        specify URL to access API server at or an alias
````
If the implementation is different, fix the implementation.
Perhaps just error out if the value for -A differs from the value in .osc/_apiurl. Or just ignore local values obtained from .osc/ if both differ. That would allow things like `ibs rq list -s new` in a pkg that is supposed to be used with `obs $cmd`.

lethliel:
> The code does not work as documented:
> 
> ```
>      -A URL/alias, --apiurl=URL/alias
>                         specify URL to access API server at or an alias
> ```
> If the implementation is different, fix the implementation.
> Perhaps just error out if the value for -A differs from the value in .osc/_apiurl. Or just ignore local values obtained from .osc/ if both differ. That would allow things like `ibs rq list -s new` in a pkg that is supposed to be used with `obs $cmd`.

This is what the code will do after merging this Pull Request. 
I am the same opinion as @olafhering that if the given -A <apiurl> differs from the one osc would work with we should error out.

lethliel:
any opinions? 

@adrianschroeter, @marcus-h, @darix

lethliel:
as discussed with Marcus this will be merged.

PR #454, opened 2018-10-17
add support for approved requests (requires OBS 2.10)

adrianschroeter:


adrianschroeter:
fixed your first two comments.

regarding the space, I did not added one on purpose, since you see that it belongs together and we don't with the result special states like "(outdated)" neither IIRC.

PR #456, opened 2018-10-23
add --alternative-project option to buildinfo

lethliel:
to be consistent with osc build

@marcus-h: please review

The option is ignored intentional when calling:

osc buildinfo <project> <package> <repo> <arch>


marcus-h:
Merged. Thanks a lot!

PR #457, opened 2018-10-25
Fix help for osc token

andreas-schwab:


lethliel:
Thanks for fixing this

PR #459, opened 2018-10-31
Add --no-delay for osc release

JanZerebecki:
This is useful so one can wait with a following osc prjresult until
everything done. Without this one can not know if a release job hasn't
started yet or is already finished.


marcus-h:
On 2018-10-30 19:38:43 -0700, Jan Zerebecki wrote:
> This is useful so one can wait with a following osc prjresult until
> everything done. Without this one can not know if a release job hasn't
> started yet or is already finished.

Issue #458, opened 2018-10-26
Where to find / Add link to release management chapter in the OBS book


JanZerebecki:
`osc release --help` refers to a release management chapter in the OBS book https://github.com/openSUSE/osc/blob/1fab61c875fb48de0447501b170c1512abe3bae1/osc/commandline.py#L2997
but I can't find it.
Where is it?
Does it exist?
It would be helpful to add a link there.

PR #460, opened 2018-10-31
Remove reference to non existing documentation

JanZerebecki:
The release management chapter in the OBS book is no where to be found.

Resolves: https://github.com/openSUSE/osc/issues/458

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -2993,8 +2993,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         """${cmd_name}: Release sources and binaries
 
         This command is used to transfer sources and binaries without rebuilding them.
-        It requires defined release targets set to trigger="manual". Please refer the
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2018-10-31 13:28:55 +0000, Jan Zerebecki wrote:
> The release management chapter in the OBS book is no where to be found.
> 
Hmm the help text probably refers to the unpublished project building
chapter [1].
@adrianschroeter: are you going to publish this chapter in the near future?

[1] https://github.com/openSUSE/obs-docu/blob/master/xml/obs_product_building.xml

tomschr:
> From the commit message [...] that was maybe because it looked like there was no useful information besides placeholders in it.

Exactly! When I worked on this chapter, there was no information about this. So I've marked this chapter as "to be definied" (tbd). :wink:  Basically, they are just structures that I think would be useful to describe.

PR #463, opened 2018-11-06
fix broken URLError handling in OscMirrorGroup.urlgrab()

lethliel:
If urlgrab returns a URLError (for example if the Network is unreachable)
the for loop did not continue and the osc build aborts.

Now we also catch the URLError and try the next mirror and return False
correctly if no mirror could be reached. And then try to download it from
api.

removed the tries counter. Not needed anymore.

### DIFF
--- a/osc/grabber.py
+++ b/osc/grabber.py
@@ -11,10 +11,12 @@ try:
     from urllib.request import HTTPError
     from urllib.parse import urlparse
     from urllib.parse import unquote
+    from urllib.error import URLError
 except ImportError:
     from urllib2 import HTTPError
     from urlparse import urlparse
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2018-11-06 04:34:25 -0800, Marco Strigl wrote:
> If urlgrab returns a URLError (for example if the Network is unreachable)
> the for loop did not continue and the osc build aborts.
> 
> Now we also catch the URLError and try the next mirror and return False
> correctly if no mirror could be reached. And then try to download it from
> api.
> 
Nice. Thanks for the PR! Merged.

> removed the tries counter. Not needed anymore.

Good catch:)

PR #464, opened 2018-11-07
[python3] add helper functions for python3 support

lethliel:
This functions are used in the whole code and are
mandatory for the python3 support to work. In python2
case nothing is touched.

* cmp_to_key:
  converts a cmp= into a key= function

* decode_list:
  decodes each element of a list. This is needed if
  we have a mixed list with strings and bytes.

* decode_it:
  Takes the input and checks if it is not a string.
  Then it uses chardet to get the encoding.

### DIFF
--- a/osc/util/helper.py
+++ b/osc/util/helper.py
@@ -1,0 +1,65 @@
+# Copyright (C) 2018 SUSE Linux.  All rights reserved.
+# This program is free software; it may be used, copied, modified
+# and distributed under the terms of the GNU General Public Licence,
+# either version 2, or (at your option) any later version.
+
+
+def cmp_to_key(mycmp):
+    """ Converts a cmp= function into a key= function.
+    """
+
+    class K(object):
+        def __init__(self, obj, *args):
+            self.obj = obj
+
+        def __lt__(self, other):
+            return mycmp(self.obj, other.obj) < 0
+
+        def __gt__(self, other):
+            return mycmp(self.obj, other.obj) > 0
+
+        def __eq__(self, other):
+            return mycmp(self.obj, other.obj) == 0
+
+        def __le__(self, other):
+            return mycmp(self.obj, other.obj) <= 0
+
+        def __ge__(self, other):
+            return mycmp(self.obj, other.obj) >= 0
+
+        def __ne__(self, other):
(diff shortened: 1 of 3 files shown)


lethliel:
@marcus-h 

an example of the use of decode_it can be seen here: 
https://github.com/openSUSE/osc/blob/6db194aee0b1ad8beee4e33c10f679f1875bd56f/osc/commandline.py#L2987

and for decode_list:
https://github.com/openSUSE/osc/blob/6db194aee0b1ad8beee4e33c10f679f1875bd56f/osc/core.py#L3818

and for cmp_to_key:
https://github.com/openSUSE/osc/blob/6db194aee0b1ad8beee4e33c10f679f1875bd56f/osc/commandline.py#L4545

There is a cmp_to _key function in the functools module which is available starting at python 2.7.15. And as we want to support python 2.6 too I had to implement the function on my own.

marcus-h:
On 2018-11-07 14:43:07 +0000, Marco Strigl wrote:
> This functions are used in the whole code and are
> mandatory for the python3 support to work. In python2
> case nothing is touched.
> 
I just had a quick glance at it... in general: LGTM.
(From my side it is OK to merge it)
(I haven't checked any callers.)

> * cmp_to_key:
>   converts a cmp= into a key= function
> 
> * decode_list:
>   decodes each element of a list. This is needed if
>   we have a mixed list with strings and bytes.
> 
Hmm sounds scary;)

> * decode_it:
>   Takes the input and checks if it is not a string.
>   Then it uses chardet to get the encoding.

Hmm guessing the encoding is also a bit worrying... but well...

marcus-h:
On 2018-11-12 21:45:00 +0100, Marcus Hüwe wrote:
> On 2018-11-07 14:43:07 +0000, Marco Strigl wrote:
> > This functions are used in the whole code and are
> > mandatory for the python3 support to work. In python2
> > case nothing is touched.
> > 
> I just had a quick glance at it... in general: LGTM.
> (From my side it is OK to merge it)
> (I haven't checked any callers.)
> 
One minor remark: can we be a bit more explicit about the
exceptions in decode_it? We should probably except
ImportError + <chardet specific stuff>.

lethliel:

PR #465, opened 2018-11-08
 Exit 3 if osc prjresults has not successful 

JanZerebecki:
Exit with value 3 if the osc prjresults has repositories / architectures
that are not finished and succesful.

Test osc prjresults

Split formating from api call in get_prj_results
This will be needed to support osc prjresults --watch without --xml.

Support osc prjresults --watch without --xml 

osc result: check all packages for wait states
The first package could be in a state that doesn't need waiting. Before
5e03ffc this worked, as only one
package was asked of the server. I.e. osc prjresults didn't use this
function, only osc results.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -5240,7 +5240,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
     @cmdln.option('-q', '--hide-legend', action='store_true',
                         help='hide the legend')
     @cmdln.option('-w', '--watch', action='store_true',
-                        help='watch the results until all finished building, only supported with --xml')
+                        help='watch the results until all finished building')
     @cmdln.option('-c', '--csv', action='store_true',
                         help='csv output')
     @cmdln.option('', '--xml', action='store_true', default=False,
@@ -5278,26 +5278,22 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             wd = os.curdir
             project = store_read_project(wd)
 
-        if opts.xml:
-            kwargs = {}
-            if opts.repo:
-                kwargs['repository'] = opts.repo
-            if opts.arch:
-                kwargs['arch'] = opts.arch
-            kwargs['wait'] = opts.watch
-            for results in get_package_results(apiurl, project, **kwargs):
+        kwargs = {}
+        if opts.repo:
+            kwargs['repository'] = opts.repo
+        if opts.arch:
+            kwargs['arch'] = opts.arch
+        kwargs['wait'] = opts.watch
+
+        for results in get_package_results(apiurl, project, **kwargs):
+            if opts.xml:
                 print(results)
-            return
-
-        if opts.watch:
-            print('Please implement support for osc prjresults --watch without --xml.')
-            return 2
-
-        print('\n'.join(get_prj_results(apiurl, project, hide_legend=opts.hide_legend, \
+            else:
+                print('\n'.join(format_prj_results(results, hide_legend=opts.hide_legend, \
                                         csv=opts.csv, status_filter=opts.status_filter, \
                                         name_filter=opts.name_filter, repo=opts.repo, \
                                         arch=opts.arch, vertical=opts.vertical, \
-                                        show_excluded=opts.show_excluded)))
+                                        show_excluded=opts.show_excluded))+'\n')
 
     @cmdln.option('-q', '--hide-legend', action='store_true',
                         help='hide the legend')
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -5261,6 +5261,9 @@ Please submit there instead, or use --nodevelproject to force direct submission.
     def do_prjresults(self, subcmd, opts, *args):
         """${cmd_name}: Shows project-wide build results
 
+        Exit value will be 3 if the repositories / architectures are not
+        finshed and successful.
+
         Usage:
             osc prjresults (inside working copy)
             osc prjresults PROJECT
@@ -5285,7 +5288,9 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             kwargs['arch'] = opts.arch
         kwargs['wait'] = opts.watch
 
+        last = None
         for results in get_package_results(apiurl, project, **kwargs):
+            last = results
             if opts.xml:
                 print(results)
             else:
@@ -5294,6 +5299,9 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                                         name_filter=opts.name_filter, repo=opts.repo, \
                                         arch=opts.arch, vertical=opts.vertical, \
                                         show_excluded=opts.show_excluded))+'\n')
+        if last and is_package_results_success(last):
+            return
+        return 3
 
     @cmdln.option('-q', '--hide-legend', action='store_true',
                         help='hide the legend')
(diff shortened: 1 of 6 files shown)


lethliel:
@JanZerebecki: Is this still work in progress or can this be reviewed? 
I want this feature in the next release if possible...

JanZerebecki:
This is ready for review.

marcus-h:
On 2018-11-20 16:58:25 -0800, Jan Zerebecki wrote:
> This is ready for review.
> 
I'll have a look at it within the next few days. Sorry
for the long delay - stay tuned:)

marcus-h:
Hmm github does not seem to like my mail(s) again... Here's the forwarded mail
```
----- Forwarded message from Marcus HÃ¼we <suse-tux@gmx.de> -----

Date: Mon, 10 Dec 2018 21:14:14 +0100
From: Marcus HÃ¼we <suse-tux@gmx.de>
To: reply+00104a4c9106cb2b0aa18dae1502557338df4149ffb5970292cf0000000117fb548f92a169ce16901547@reply.github.com
Subject: Re: [openSUSE/osc]  Exit 3 if osc prjresults has not successful  (#465)
User-Agent: NeoMutt/20180716

Thanks a lot for the PR! Apart from PATCH 4, it looks really good!
For the comments, see below.


On 2018-11-07 17:34:07 -0800, Jan Zerebecki wrote:
> Subject: [PATCH 1/5] Test osc prjresults
> 
LGTM.

> Subject: [PATCH 2/5] Split formating from api call in get_prj_results
> 
> This will be needed to support osc prjresults --watch without --xml.
> ---
>  osc/core.py | 8 ++++++--
>  1 file changed, 6 insertions(+), 2 deletions(-)
> 
> index f5a5ef0f..870a9502 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -5778,12 +5778,16 @@ def get_package_results(apiurl, project, package=None, wait=False, *args, **kwar
>  
>  
>  def get_prj_results(apiurl, prj, hide_legend=False, csv=False, status_filter=None, name_filter=None, arch=None, repo=None, vertical=None, show_excluded=None):
> -    #print '----------------------------------------'
> +    """ this function is only needed for backward/api compatibility; use get_package_results() and format_prj_results() instead"""
> +    xml_string = show_prj_results_meta(apiurl, prj)

Hmm the variable name is a bit misleading: it is list instead of a string.
Maybe just do something like
xml_string = ''.join(show_prj_results_meta(apiurl, prj))

> +    return format_prj_results(xml_string, hide_legend, csv, status_filter, name_filter, arch, repo, vertical, show_excluded)
> +
> +def format_prj_results(xml_string, hide_legend=False, csv=False, status_filter=None, name_filter=None, arch=None, repo=None, vertical=None, show_excluded=None):

Hmm maybe call it "format_and_filter_prj_results" (since it does more than
formatting (unfortunately...)).

>      global buildstatus_symbols
>  
>      r = []
>  
> -    f = show_prj_results_meta(apiurl, prj)
> +    f = xml_string

Huh? just use xml_string in the ET.fromstring call:)

>      root = ET.fromstring(''.join(f))



> Subject: [PATCH 3/5] Support osc prjresults --watch without --xml

> index 6cd4e27d..3c4839ae 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py

> -        print('\n'.join(get_prj_results(apiurl, project, hide_legend=opts.hide_legend, \
> +        for results in get_package_results(apiurl, project, **kwargs):
> +            if opts.xml:
> +                print(results)
> +            else:
> +                print('\n'.join(format_prj_results(results, hide_legend=opts.hide_legend, \
>                                          csv=opts.csv, status_filter=opts.status_filter, \
>                                          name_filter=opts.name_filter, repo=opts.repo, \
>                                          arch=opts.arch, vertical=opts.vertical, \
> -                                        show_excluded=opts.show_excluded)))
> +                                        show_excluded=opts.show_excluded))+'\n')
>  
Hmm why the extra newline? It probably makes sense in the opts.watch case,
but let's not "break" the current output format.


> Subject: [PATCH 4/5] Exit 3 if osc prjresults has not successful
> 
> Exit with value 3 if the osc prjresults has repositories / architectures
> that are not finished and succesful.

Hmm this will break existing scripts that take the return code of
"osc prjresults" into account. Actually, I don't see a good reason why
we should use a non-zero exit status in this case (if we would do it here,
then, for instance, "osc ls -u prj pkg" should also use a non-zero exit
status in case of a link error or service error (which is not a good idea
(IMHO))). What's your use case?

> index 870a9502..712a3c7a 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -5776,6 +5776,23 @@ def get_package_results(apiurl, project, package=None, wait=False, *args, **kwar
>              yield xml 
>      yield xml
>  
> +def is_package_results_success(xmlstring):
> +    ok = ('succeeded', 'disabled', 'excluded', 'published', 'unpublished')
> +
> +    root = ET.fromstring(xmlstring)
> +    for result in root.findall('result'):
> +        if result.get('dirty') is not None:
> +            return False
> +        if result.get('code') not in ok:
> +            return False
> +        if result.get('state') not in ok:
> +            return False
> +        packages = result.find('status')

This not doing the right thing (see below (PATCH 5)).

> +        for p in packages:
> +            if p.get('code') not in ok:
> +                return False
> +    return True
> +

> diff --git a/tests/test_results.py b/tests/test_results.py
> index 19843b23..b9d14101 100644
> --- a/tests/test_results.py
> +++ b/tests/test_results.py
> @@ -18,47 +18,58 @@ def _get_fixtures_dir(self):
>          return os.path.join(os.path.dirname(__file__), self._get_fixtures_name())
>  
>      def _run_osc(self, *args):
> -        """Runs osc, returning captured STDOUT as a string."""
> +        """Runs osc, returning (exit value, captured STDOUT as a string)."""
>          cli = osc.commandline.Osc()
>          argv = ['osc', '--no-keyring', '--no-gnome-keyring']
>          argv.extend(args)
> -        cli.main(argv=argv)
> -        return sys.stdout.getvalue()
> +        exit = cli.main(argv=argv)
> +        return (exit, sys.stdout.getvalue())
>  
Cosmetics/style: return exit, sys.stdout.getvalue()


> Subject: [PATCH 5/5] osc result: check all packages for wait states
> 
> The first package could be in a state that doesn't need waiting. Before
> 5e03ffcec96f3c86ae2c8ff75bc346f0ca9c8874 this worked, as only one
> package was asked of the server. I.e. osc prjresults didn't use this
> function, only osc results.

Good catch!

> index 712a3c7a..ef077f1d 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -5765,9 +5765,12 @@ def get_package_results(apiurl, project, package=None, wait=False, *args, **kwar
>                  waiting = True
>                  break
>              else:
> -                pkg = result.find('status')
> -                if pkg is not None and pkg.get('code') in waiting_states:
> -                    waiting = True
> +                packages = result.find('status')

This just returns the first status element (or None!)
Something like

status = result.findall('status')
for st in status:
    ...

should do the trick.

> +                for p in packages:
> +                    if p.get('code') in waiting_states:
> +                        waiting = True
> +                        break
> +                if waiting:
>                      break

```

lethliel:
Since there is no reply from the author we will close this PR. If you think this is not right. Please reopen it.

PR #466, opened 2018-11-09
[python3] change to open() with 'rb' to get bytes in python3

lethliel:
With this change you get bytes with python3 and string in python2

@marcus-h: This is needed that the fixtures are returned as bytes not as string and to mock the same behavior as urllib.request.urlopen()

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: python
 python:
 - '2.7'
-- '3.3'
-- '3.4'
-- 3.5-dev
+#- '3.3'
+#- '3.4'
+#- 3.5-dev
 addons:
   apt:
     packages:
--- a/tests/common.py
+++ b/tests/common.py
@@ -20,6 +20,8 @@ except ImportError:
     from urllib.request import HTTPHandler, addinfourl, build_opener
     from urllib.parse import urlparse, parse_qs
 
+from io import BytesIO
+
 def urlcompare(url, *args):
     """compare all components of url except query string - it is converted to
     dict, therefor different ordering does not makes url's different, as well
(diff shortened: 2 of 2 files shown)


lethliel:
@marcus-h: I had to deactivate the python3.x travis checks. Otherwise they will fail until the complete python3 branch is merged.

marcus-h:
On 2018-11-09 09:48:09 +0000, Marco Strigl wrote:
> @marcus-h: I had to deactivate the python3.x travis checks. Otherwise they will fail until the complete python3 branch is merged.
> 
Ok. Sounds reasonable:)

marcus-h:
On 2018-11-09 01:40:00 -0800, Marco Strigl wrote:
> With this change you get bytes with python3 and string in python2
> 
> @marcus-h: This is needed that the fixtures are returned as bytes not as string and to mock the same behavior as urllib.request.urlopen()

> diff --git a/tests/common.py b/tests/common.py
> index 65bd907e..a75f83f5 100644
> --- a/tests/common.py
> +++ b/tests/common.py
> @@ -105,7 +105,7 @@ def __get_response(self, url, **kwargs):
>          if 'exception' in kwargs:
>              raise kwargs['exception']
>          if 'text' not in kwargs and 'file' in kwargs:
> -            f = StringIO(open(os.path.join(self.__fixtures_dir, kwargs['file']), 'r').read())
> +            f = open(os.path.join(self.__fixtures_dir, kwargs['file']), 'rb')

Hmm this will "leak" a file object. Can't we simply use io.BytesIO?

>          elif 'text' in kwargs and 'file' not in kwargs:
>              f = StringIO(kwargs['text'])

Hmm isn't it a bit inconsistent that a "text" is treated as a str and
a "file" is treated as bytes? This would make osc's life much harder
than it is in reality (the file-like object returned by urllib should
always be in "byte" mode).

lethliel:
@marcus-h:

Now using BytesIO and treating files as bytes objects too.

marcus-h:
On 2018-11-13 15:00:47 +0000, Marco Strigl wrote:
> Now using BytesIO and treating files as bytes objects too. 
> 
Thanks! Merged.

Issue #468, opened 2018-11-17
osc submitrequest should accept all kinds of revisions


johanneskastl:
Currently `osc submitrequest -r` needs to be called with the revision of the expanded sources, see [this mail](https://lists.opensuse.org/opensuse-buildservice/2018-11/msg00016.html) on the buildservice list.

It would be great if any of the other revision "descriptors" would work. I.e. the one you get from `osc log` that reads `r7`, or the plain number (as can be seen on the Revisions tab of the package in the WebUI)  or the hash that also is given in `osc log`.

johanneskastl:
Hi Marcus,

sorry, you lost me there. ;-)

In my (simple) case nothing other than the buildhist revisions did work. I would not call that "revisions are already supported". It seems to me like the job to find out which exact "correct revision" I need when I say "take revision 7" is something that osc should do, not me. 

Kind regards,
Johannes

johanneskastl:
Adrian, I could successfully checkout the revisions I wanted to submit. I saw no errors and no conflicts. That was the first thing I tried.

So, I'm puzzled. Do you actually mean "our revision" (somewhere inside OBS or something) or did you want to say "your revision" (aka the package in `home:ojkastl_buildservice:...`?

adrianschroeter:
at least the mentioned revision 7 from your revision has an expansion error:

```
# osc api /source/home:ojkastl_buildservice:Branch_Virtualization_vagrant/vagrant?rev=7
<directory name="vagrant" rev="7" vrev="7" srcmd5="7dc081528e717bbe3f1c2f9c77cbebb0">
  <linkinfo project="Virtualization:vagrant" package="vagrant" baserev="02db93e63a7b53d65ede64c657288ded" xsrcmd5="267ab3e3fe08bab816667de195b522b5" error="conflict in file vagrant.spec" />
...
```

adrianschroeter:
btw, we could support non fixed (or not versioned) revisions in requests in osc (api is supporting this). But openSUSE submissions would be refused automatically, since they are against the policy there (review must happen on fixed sources).

johanneskastl:
Hi Adrian, yes, now it has errors, as I submitted several revisions of the package with Andreas' workaround. I just tested, I can no longer check out those revisions, which worked fine when I sent the mail. Believe me, that was the first thing I tried.

I have no clue what "non fixed (or not versioned)" are, I would just like to have osc do the magic, so that a simple `osc sr -r 7 -m "here it is"` would work.

[...]

PR #469, opened 2018-11-17
add explanation to 'osc help sr' regarding the revision

johanneskastl:
add explanation to 'osc help sr', that explains that the revision for the '-r' option needs to be the one from 'osc buildhist', others won't work...

Created issue #468 to fix the behaviour of osc, if possible.


johanneskastl:
Marcus, let's discuss this in #468, but I think hinting at which revision might work in half a line of text is better than not hinting at all. And your explanation over in #468 is a lot longer than that half a line, and I still don't understand it. ;-)

lethliel:
It is surprising that the linkrev points to the latest revision. But it can happen and not only for this command and we do not want to provide special cases for a single command.

Issue #471, opened 2018-11-20
not existing key (path to key) gets added to buildinfo (local builds)


lethliel:
The not existing key (path to key) of the project openSUSE_Leap_15.0 gets added to the buildinfo
and verify_pacs tries to verify with the non existing key then.

This is due to a logic error in fetch.py. The path to the key should only be added if the key is found. 

```
osc build --no-init --offline openSUSE_Leap_15.0    x86_64
WARNING: source service from package or project will not be executed. This may not be the same build as on server!
Building osc.spec for openSUSE_Leap_15.0/x86_64
Use local '/home/marco/obs/home:mstrigl/osc/.osc/_buildinfo-openSUSE_Leap_15.0-x86_64.xml' file as buildinfo
Use local '/home/marco/obs/home:mstrigl/osc/.osc/_buildconfig-openSUSE_Leap_15.0-x86_64' file as buildconfig
Updating cache of required packages
0.0% cache miss. 139/139 dependencies cached.
Verifying integrity of cached packages
using keys from openSUSE:Leap:15.0, home:mstrigl, openSUSE:Leap
Traceback (most recent call last):
  File "/home/marco/git_r/osc_python3/osc/osc-wrapper.py", line 41, in <module>
    r = babysitter.run(osccli)
  File "/home/marco/git_r/osc_python3/osc/osc/babysitter.py", line 61, in run
    return prg.main(argv)
  File "/home/marco/git_r/osc_python3/osc/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/home/marco/git_r/osc_python3/osc/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/home/marco/git_r/osc_python3/osc/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/home/marco/git_r/osc_python3/osc/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/home/marco/git_r/osc_python3/osc/osc/commandline.py", line 6348, in do_build
    return osc.build.main(self.get_api_url(), opts, args)
  File "/home/marco/git_r/osc_python3/osc/osc/build.py", line 1106, in main
    verify_pacs(bi)
  File "/home/marco/git_r/osc_python3/osc/osc/fetch.py", line 401, in verify_pacs
    checker.readkeys(bi.keys)
[...]

PR #472, opened 2018-11-20
do not add key to buildinfo if not existent

lethliel:
fixes issue: https://github.com/openSUSE/osc/issues/471

Add the key_path only to buildinfo(bi) if it exists (try_parent == false)


marcus-h:
Merged. Thanks.

PR #476, opened 2018-11-28
Do not choke on operators < and > in spec files

fridrich:


lethliel:
As already discussed in IRC this is needed if a local repository is used with -p /path/to/repository

@fridrich Thanks again for your contribution.

PR #479, opened 2018-12-13
[python3] enable full python3 support for man page building

lethliel:
enable full python3 support for man page building


lethliel:

PR #480, opened 2018-12-13
[python3] the result of a division is a float

lethliel:
The result of a division is a float but we need int to continue.


marcus-h:
On 2018-12-13 04:37:21 -0800, Marco Strigl wrote:
> The result of a division is a float but we need int to continue.

LGTM. Directly merged because it is a low-hanging fruit. Thanks!

PR #481, opened 2018-12-13
[python3] self[i] is now self.args[i]

lethliel:
self[i] is now self.args[i]

### DIFF
--- a/osc/oscerr.py
+++ b/osc/oscerr.py
@@ -82,7 +82,7 @@ class WorkingCopyOutdated(OscBaseError):
     def __str__(self):
         return ('Working copy \'%s\' is out of date (rev %s vs rev %s).\n'
                'Looks as if you need to update it first.' \
-                    % (self[0], self[1], self[2]))
+                    % (self.args[0], self.args[1], self.args[2]))
 
 class PackageError(OscBaseError):
     """Base class for all Package related exceptions"""


marcus-h:
On 2018-12-13 04:38:05 -0800, Marco Strigl wrote:
> diff --git a/osc/oscerr.py b/osc/oscerr.py
> index 7eb0f7ae..5f77f79d 100644
> --- a/osc/oscerr.py
> +++ b/osc/oscerr.py
> @@ -82,7 +82,7 @@ class WorkingCopyOutdated(OscBaseError):
>      def __str__(self):
>          return ('Working copy \'%s\' is out of date (rev %s vs rev %s).\n'
>                 'Looks as if you need to update it first.' \
> -                    % (self[0], self[1], self[2]))
> +                    % (self.args[0], self.args[1], self.args[2]))

For the change itself: LGTM.
However, the change is not as innocent as it looks because self.args[0]
is a pathname that is obtained via os.path.absdir(<some str>). For instance,

- create a directory that contains an invalid utf-8 encoding
  (for example via os.mkdir(b'foo\xff\xffx'))
- cd into this directory
- osc co some prj/pkg
- provoke a WorkingCopyOutdated

This should (_untested_) result in an exception in babysitter.py.

The following script shows the underlying issue

marcus@linux:~/brk> cat test.py 
#!/usr/bin/python3

import os


dirname = b'xxx/foo\xff\xffbar'
os.makedirs(dirname)
os.chdir(dirname)
abspath = os.path.abspath('some_file_as_a_str')
print(abspath)  # this raises an exception
marcus@linux:~/brk>


Something like this will probably hit us in numerous places... :/
(In short: I'm OK with merging this patch)

PR #482, opened 2018-12-13
[python3] magic is now a bytestring in python3

lethliel:
magic is a bytestring in python3


marcus-h:
On 2018-12-13 04:38:41 -0800, Marco Strigl wrote:
> magic is a bytestring in python3

LGTM.

PR #483, opened 2018-12-13
[python3] rpmquery.py now python3 ready

lethliel:
* new function cmp (not available in python3)
  * decode unpack data

### DIFF
--- a/osc/util/rpmquery.py
+++ b/osc/util/rpmquery.py
@@ -5,6 +5,10 @@ import os
 import re
 import struct
 from . import packagequery
+from osc.util.helper import decode_it
+
+def cmp(a, b):
+    return (a > b) - (a < b)
 
 class RpmError(packagequery.PackageError):
     pass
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2018-12-13 12:39:23 +0000, Marco Strigl wrote:
> diff --git a/osc/util/rpmquery.py b/osc/util/rpmquery.py
> index 597b916d..9570ad54 100644
> --- a/osc/util/rpmquery.py
> +++ b/osc/util/rpmquery.py
> @@ -6,6 +6,9 @@
>  import struct
>  from . import packagequery
>  
> +def cmp(a, b):
> +    return (a > b) - (a < b)
> +

OK.


>  class RpmError(packagequery.PackageError):
>      pass
>  
> @@ -374,7 +377,7 @@ def unpack_string(data):
>      """unpack a '\\0' terminated string from data"""
>      val = ''
>      for c in data:
> -        c, = struct.unpack('!c', c)
> +        c = struct.unpack('!c', bytes(bytearray([c])))[0].decode('utf-8')
>          if c == '\0':
>              break
>          else:

Hmm this is wrong. For instance, if data contains a valid utf-8 encoding
of a non-ascii character, this will lead to an exception because you try
to decode a single byte as an utf-8 str (which will fail for the prefix
byte).

Just test what happens if you set data to b'This is a \xe2\x82\xac sign\0'.

I just simplified unpack_string in commit
28eeff86830b4d2cc62356aad23c1abd7be77834 ("Simplify rpmquery.unpack_string")
and added an optional encoding parameter that could be used by the python3
port. However, blindly assuming an utf-8 encoding is actually wrong because
an rpm can contain illegal utf-8 sequences... :/

So the question is if we "force" an utf-8 encoding for now or make this
simply a "bytes" interface?

lethliel:

PR #484, opened 2018-12-13
[python3] setup.py now supports python3

lethliel:
setup.py now supports python3


lethliel:

PR #485, opened 2018-12-13
[python3] babysitter.py msg and body are byte strings

lethliel:


lethliel:

PR #486, opened 2018-12-13
[python3] switch to bytestrings

lethliel:


marcus-h:
On 2018-12-13 13:59:16 +0000, Marco Strigl wrote:
> diff --git a/osc/util/archquery.py b/osc/util/archquery.py
> index 72fa0242..ea464d82 100644
> --- a/osc/util/archquery.py
> +++ b/osc/util/archquery.py
> @@ -6,6 +6,7 @@
>  import tarfile
>  from . import packagequery
>  import subprocess
> +from osc.util.helper import decode_it
>  
Can you remove this import? (See below).

> @@ -28,11 +29,11 @@ def read(self, all_tags=True, self_provides=True, *extra_tags):
>          fn = open('/dev/null', 'wb')
>          pipe = subprocess.Popen(['tar', '-O', '-xf', self.__path, '.PKGINFO'], stdout=subprocess.PIPE, stderr=fn).stdout
>          for line in pipe.readlines():
> -            line = line.rstrip().split(' = ', 2)
> +            line = line.rstrip().split(b' = ', 2)
>              if len(line) == 2:
>                  if not line[0] in self.fields:
> -                    self.fields[line[0]] = []
> -                self.fields[line[0]].append(line[1])
> +                    self.fields[decode_it(line[0])] = []
> +                self.fields[decode_it(line[0])].append(line[1])

Currently, all supported fields are plain ascii. Instead of using
decode_it, can you make the decoding explicit?

>          if self_provides:
>              prv = '%s = %s' % (self.name(), self.fields['pkgver'][0])
>              self.fields.setdefault('provides', []).append(prv)
> @@ -115,7 +116,8 @@ def enhances(self):
>  
>      def canonname(self):
>          pkgver = self.fields['pkgver'][0] if 'pkgver' in self.fields else None

(Hmm the else part looks fishy - the code below will break in case of None
(but that's a different issue/has nothing to do with the python3 port).)

> -        return self.name() + '-' + pkgver + '-' + self.arch() + '.' + self.pkgsuffix
> +        canonname = self.name() + b'-' + pkgver + b'-' + self.arch() + b'.' + self.pkgsuffix
> +        return decode_it(canonname)

Just return the canonname without decoding it (for instance, pkgver could
be an illegal (utf-8) encoding).

The port is a bit incomplete (version, release, epoch, ... are missing).
If you change the 3 things from above, I'll merge this PR and take care of
the rest. (Maybe you could also state in the commit message what you changed
to bytestrings (ArchQuery.read, ArchQuery.filename).)

lethliel:
@marcus-h: done

marcus-h:
Merged. Thanks!

PR #487, opened 2018-12-13
[python3] switch to bytestrings

lethliel:


marcus-h:
I ported the ar module a bit differently to have a bytes-only API.
For the details and the rationale, see commit
6fdce86fc9994a7b28569fbdd800b38632075442 ("Port the ar module to
python3"). The bytes-only API affects your PR#489 (the debquery
stuff) - I'll add a comment there. Additionally, I did some minor
cleanups here and there.

PR #488, opened 2018-12-13
[python3] data and other vars are byte strings

lethliel:

### DIFF
--- a/osc/util/cpio.py
+++ b/osc/util/cpio.py
@@ -21,6 +21,8 @@ import stat
 import struct
 import sys
 
+from osc.util.helper import decode_it
+
 # workaround for python24
 if not hasattr(os, 'SEEK_SET'):
     os.SEEK_SET = 0
@@ -161,7 +163,7 @@ class CpioRead:
             self.__file.seek(0, os.SEEK_SET)
         self._init_datastructs()
         data = self.__file.read(6)
-        self.format = data
+        self.format = decode_it(data)
         if not self.format in self.sfmt.values():
             raise CpioError(self.filename, '\'%s\' is not a supported cpio format' % self.format)
         pos = 0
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2018-12-13 14:00:54 +0000, Marco Strigl wrote:
> diff --git a/osc/util/cpio.py b/osc/util/cpio.py
> index cd3c4065..6746abc3 100644
> --- a/osc/util/cpio.py
> +++ b/osc/util/cpio.py
> @@ -21,6 +21,8 @@

<SNIP>

> @@ -173,7 +175,7 @@ def read(self):
>              pos += self.hdr_len
>              data = struct.unpack(self.hdr_fmt, data)
>              hdr = CpioHdr(*data)
> -            hdr.filename = self.__file.read(hdr.namesize - 1)
> +            hdr.filename = decode_it(self.__file.read(hdr.namesize - 1))

In general, a filename is not necessarily a valid encoding. If we
limit ourselves to the OBS case, it might work (I'm not sure if it
is possible to sneak in an arbitrary filename).

> @@ -212,12 +214,12 @@ class CpioWrite:
>      def __init__(self):
>          self.cpio = ''
> 
> -    def add(self, name=None, content=None, perms=0x1a4, type=0x8000):
> -        namesize = len(name) + 1
> +    def add(self, name=None, content=None, perms=0x1a4, my_type=0x8000):
> +        namesize = len(name.encode('utf-8')) + 1

Hmm this could result in a mismatch of the namesize and the actual
length of the filename (namesize is the length of the encoding but
you "store" the plain str in the archive).


>          if namesize % 2:
>              name += '\0'
>          filesize = len(content)
> -        mode = perms | type
> +        mode = perms | my_type
> 
>          c = []
>          c.append('070701') # magic
> @@ -237,8 +239,10 @@ def add(self, name=None, content=None, perms=0x1a4, type=0x8000):
> 
>          c.append(name + '\0')
>          c.append('\0' * (len(''.join(c)) % 4))
> -
> -        c.append(content)
> +        if isinstance(content, str):
> +            c.append(content)
> +        else:
> +            c.append(decode_it(content))

In the bytes case, there could be a mismatch because filesize is the
number of bytes but you store the decoded str.

I ported the cpio module a bit differently (see commit
5387744d36bfed24c32b0e31facef9a4b73d9414 ("Port CpioWrite to python3")
and commit 3e326b1bb42283556a51b2e5fa5c37b4dacd64a6 ("Port CpioRead
and CpioHdr to python3")). The end result is in both cases a bytes-only
API. If you plan to integrate these changes into your python3 branch,
feel free to ping me so that I can fix the breakage (which I introduced).

PR #489, opened 2018-12-13
[python3] debin and data are byte strings

lethliel:

### DIFF
--- a/osc/util/debquery.py
+++ b/osc/util/debquery.py
@@ -5,8 +5,12 @@ from . import ar
 import os.path
 import re
 import tarfile
-import StringIO
+try:
+    from io import StringIO
+except ImportError:
+    import StringIO
 from . import packagequery
+from osc.util.helper import decode_it
 
 HAVE_LZMA = True
 try:
@@ -34,7 +38,7 @@ class DebQuery(packagequery.PackageQuery, packagequery.PackageQueryResult):
         debbin = arfile.get_file('debian-binary')
         if debbin is None:
             raise DebError(self.__path, 'no debian binary')
-        if debbin.read() != '2.0\n':
+        if debbin.read() != b'2.0\n':
             raise DebError(self.__path, 'invalid debian binary format')
         control = arfile.get_file('control.tar.gz')
         if control is not None:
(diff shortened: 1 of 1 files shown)


marcus-h:
Commit 6fdce86fc9994a7b28569fbdd800b38632075442 ("Port the ar
module to python3") "breaks" this change. It can be fixed by
passing a bytes to each ar.get_file(...) call in debquery.py.

marcus-h:

On 2018-12-13 14:01:31 +0000, Marco Strigl wrote:
> diff --git a/osc/util/debquery.py b/osc/util/debquery.py
> index 4c2e0e5f..e6f9725e 100644
> --- a/osc/util/debquery.py
> +++ b/osc/util/debquery.py
> @@ -64,38 +68,39 @@ def read(self, all_tags=False, self_provides=True, *extra_tags):
>      def __parse_control(self, control, all_tags=False, self_provides=True, *extra_tags):
>          data = control.readline().strip()
>          while data:
> -            field, val = re.split(':\s*', data.strip(), 1)
> +            field, val = re.split(b':\s*', data.strip(), 1)
>              data = control.readline()
> -            while data and re.match('\s+', data):
> -                val += '\n' + data.strip()
> +            while data and re.match(b'\s+', data):
> +                val += b'\n' + data.strip()
>                  data = control.readline().rstrip()
> -            field = field.replace('-', '_').lower()
> +            field = field.replace(b'-', b'_').lower()
> +            field = decode_it(field)

If we want to decode a field, which is ok for a sane deb, we can
simply do "field = field.decode('ascii')" instead of the decode_it
stuff. (Note: if we wouldn't assume an ascii encoding, we would need
to decode first and then perform the .lower() call)

<SNIP>

> -        verep = self.fields['version'].split(':', 1)
> +        verep = self.fields['version'].split(b':', 1)
>          if len(verep) == 2:
>              self.fields['epoch'] = verep[0]
>              self.fields['version'] = verep[1]
>          else:
>              self.fields['epoch'] = '0'

This should be

self.fields['epoch'] = b'0'

(otherwise we might run into trouble when doing a vercmp).

<SNIP>

>          if self_provides:
>              # add self provides entry
> -            self.fields['provides'].append('%s (= %s)' % (self.name(), '-'.join(versrel)))
> +            self.fields['provides'].append('%s (= %s)' % (self.name(), b'-'.join(versrel)))

The constructed provides entry will contain the b'...' stuff
(see below (def filename)) - just append a bytes: b'%s (= %s)' % ...


Apart from that the patch looks good to me - nice work!:)


What's missing:
- correct filename construction: the computed filename is
  formatted with bytes => the resulting str contains the
  bytes "representation" (for instance "b'foo'-b'1.2.3'...")
- debvercmp does not work with bytes atm
- cmp function is missing

Probably, the best is if def filename returns bytes and def debvercmp
works with bytes as well (though, from a technical POV strs should be
fine here (since a valid/sane deb only allows a subset of ascii chars)).

marcus-h:
On 2019-01-16 01:16:14 +0100, Marcus Hüwe wrote:
> On 2018-12-13 14:01:31 +0000, Marco Strigl wrote:
> > diff --git a/osc/util/debquery.py b/osc/util/debquery.py
> > index 4c2e0e5f..e6f9725e 100644
> > --- a/osc/util/debquery.py
> > +++ b/osc/util/debquery.py
> > @@ -64,38 +68,39 @@ def read(self, all_tags=False, self_provides=True, *extra_tags):
> >      def __parse_control(self, control, all_tags=False, self_provides=True, *extra_tags):
> >          data = control.readline().strip()
> >          while data:
> > -            field, val = re.split(':\s*', data.strip(), 1)
> > +            field, val = re.split(b':\s*', data.strip(), 1)
> >              data = control.readline()
> > -            while data and re.match('\s+', data):
> > -                val += '\n' + data.strip()
> > +            while data and re.match(b'\s+', data):
> > +                val += b'\n' + data.strip()
> >                  data = control.readline().rstrip()
> > -            field = field.replace('-', '_').lower()
> > +            field = field.replace(b'-', b'_').lower()
> > +            field = decode_it(field)
> 
> If we want to decode a field, which is ok for a sane deb, we can
> simply do "field = field.decode('ascii')" instead of the decode_it
> stuff. (Note: if we wouldn't assume an ascii encoding, we would need
> to decode first and then perform the .lower() call)
> 
Hmm update: assuming an ascii encoding is only correct for the current
"standard" fields. It is possible to register _arbitrary_ custom fields
(via a 'register-custom-fields' hook). Hence, we are probably better
off if we do not decode the field (even though it is probably very
unlikely that we will ever encounter a binary field...).

I'll take care of it.

marcus-h:
I used a different approach for porting this module. For the details,
see commit cd5f46984d64664c13ab0c8c6decc1229935e7de ("Port debquery
module to python3").

PR #490, opened 2018-12-13
[python3] make oscssl.py python3 ready.

lethliel:
* works also with python2

Should be safe to merge immediately.


lethliel:

PR #492, opened 2018-12-13
[python3] python3 support for core library

lethliel:
* use cmp_to_key from functools for python 2.7 and higer
* use self written cmp_to_key for python 2.6
* new functions compare und cmp (used in python2 and python3)
* a lot of bytestring handling and decoding

Merge only together with https://github.com/openSUSE/osc/pull/493


lethliel:

Issue #499, opened 2019-01-07
Missing progress bar in getbinaries (python3)


mcepl:
```
matej@milic: rpms$ osc getbinaries Virtualization libvirt openSUSE_Factory x86_64
Creating binaries
Traceback (most recent call last):
  File "/home/matej/.local/bin/osc", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/matej/archiv/knihovna/repos/osc/osc-py3-wrapper.py", line 41, in <module>
    r = babysitter.run(osccli)
  File "/home/matej/archiv/knihovna/repos/osc/osc/babysitter.py", line 61, in run
    return prg.main(argv)
  File "/home/matej/archiv/knihovna/repos/osc/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/home/matej/archiv/knihovna/repos/osc/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/home/matej/archiv/knihovna/repos/osc/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/home/matej/archiv/knihovna/repos/osc/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/home/matej/archiv/knihovna/repos/osc/osc/commandline.py", line 7188, in do_getbinaries
    progress_meter = not opts.quiet)
  File "/home/matej/archiv/knihovna/repos/osc/osc/core.py", line 4659, in get_binary_file
    from .meter import TextMeter
  File "/home/matej/archiv/knihovna/repos/osc/osc/meter.py", line 6, in <module>
    import progressbar as pb
ModuleNotFoundError: No module named 'progressbar'
```

lethliel:
You need to install python3-progressbar. 
You use the python3 branch?

mcepl:
Oh, I see, I thought that it is a leftover from python2. Yes, I (happily) use python3 branch.

lethliel:
As master and the python3 branch share the same code for progressbar generation since https://github.com/openSUSE/osc/commit/65b053abb3b5771303ec998a813d600e2a776ab6 it is also a issue for the master branch. 

I fixed it with this https://github.com/openSUSE/osc/commit/4a28448e14368d7aa66ec5afb541a7b48d5f145d for the python3 branch. The same will go into master. 

@mcepl 
[...]

PR #500, opened 2019-01-08
warn user if python-progressbar is not installed

lethliel:
On ImportError have_pb_module is false and the class NoPBTextMeter gets
returned which prints "Please install progressbar module..." on TextMeter.start()

fixes https://github.com/openSUSE/osc/issues/499 which also affects the master branch.

@marcus-h please review. Thanks.

### DIFF
--- a/osc/meter.py
+++ b/osc/meter.py
@@ -3,10 +3,14 @@
 # and distributed under the terms of the GNU General Public Licence,
 # either version 2, or (at your option) any later version.
 
-import progressbar as pb
+try:
+    import progressbar as pb
+    have_pb_module = True
+except ImportError:
+    have_pb_module = False
 
 
-class TextMeter(object):
+class PBTextMeter(object):
 
     def start(self, basename, size=None):
         if size is None:
@@ -24,4 +28,18 @@ class TextMeter(object):
     def end(self):
         self.bar.finish()
 
+class NoPBTextMeter(object):
+    def start(self, *args, **kwargs):
+        print('Please install the progressbar module...')
+
+    def update(self, *args, **kwargs):
+        pass
+
+    def end(self, *args, **kwargs):
+        pass
+
+if have_pb_module:
+    TextMeter = PBTextMeter
+else:
+    TextMeter = NoPBTextMeter
 # vim: sw=4 et


marcus-h:
On 2019-01-08 05:26:57 -0800, Marco Strigl wrote:
> index 9377dc80..fc1a6b48 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -4614,17 +4614,13 @@ def get_binary_file(apiurl, prj, repo, arch,
>                      package = None,
>                      target_filename = None,
>                      target_mtime = None,
> -                    progress_meter = False):
> -    progress_obj = None
> -    if progress_meter:
> -        from .meter import TextMeter
> -        progress_obj = TextMeter()
> +                    progress_meter = None):
>  
Hmm this changes the API in an incompatible way: all external
callers will break, if they pass a progress_meter=<bool> argument (even
our internal call in build.py breaks) (yes, I agree the current API
"sucks" and in an ideal world we would just pass the progress_obj:)).

I'm not sure if changing the API is a good idea. Maybe we can simply
"fix" this in the meter.py:

try:
    import progressbar as pb
    have_pb = True
except ImportError:
    have_pb = False


# rename existing TextMeter class to PBTextMeter (or something like that)
# add a NoPBTextMeter/DummyTextMeter class
class NoPBTextMeter:
    def start(self, *args, **kwargs):
        print('Please install the progressbar module...')

    def update(self, *args, **kwargs):
        pass

    def end(self, *args, **kwargs):
        pass


if have_pb:
    TextMeter = PBTextMeter
else:
    TextMeter = NoPBTextMeter


This would avoid the ImportError dance(s).

lethliel:
On 1/8/19 3:45 PM, Marcus Hüwe  wrote:
> On 2019-01-08 05:26:57 -0800, Marco Strigl wrote:
>> index 9377dc80..fc1a6b48 100644
>> --- a/osc/core.py
>> +++ b/osc/core.py
>> @@ -4614,17 +4614,13 @@ def get_binary_file(apiurl, prj, repo, arch,
>> package = None,
>> target_filename = None,
>> target_mtime = None,
>> - progress_meter = False):
>> - progress_obj = None
>> - if progress_meter:
>> - from .meter import TextMeter
>> - progress_obj = TextMeter()
>> + progress_meter = None):
>>
> Hmm this changes the API in an incompatible way: all external
> callers will break, if they pass a progress_meter=<bool> argument (even
> our internal call in build.py breaks) (yes, I agree the current API
> "sucks" and in an ideal world we would just pass the progress_obj:)).
> 

Oopsie. You are right. This is probably a bad idea. 
Will fix it in meter.py 

Thanks for the review.

marcus-h:
Merged. Thanks! I just made a small follow up commit (see commit
be94132298691c1d78875a440de97f4d819afd7c ("Remove superfluous
try-except from meter import")).

PR #501, opened 2019-01-09
osc shell/chroot is now handled via build script (working for chroot …

adrianschroeter:
…and KVM only atm)

We had bugreports about not matching mount points when using "osc chroot", so using the build script implementation now only. This should work for chroot and KVM at least, but I might have missed some special cases. So please review :)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,6 @@
+0.167
+  - osc shell/chroot/wipe is now handled via build script (working for chroot and KVM only atm)
+
 0.166.2
   - Don't enforce password reuse (boo#1156501)
   - Config option check_for_requests_on_action is now
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,6 @@
 0.167
   - osc shell/chroot/wipe is now handled via build script (working for chroot and KVM only atm)
+  - osc build --vm-type=qemu support for cross architecture builds
 
 0.166.2
   - Don't enforce password reuse (boo#1156501)
--- a/osc/build.py
+++ b/osc/build.py
@@ -46,15 +46,6 @@ change_personality = {
             'sparcv8': 'linux32',
         }
 
-# FIXME: qemu_can_build should not be needed anymore since OBS 2.3
-qemu_can_build = [ 'armv4l', 'armv5el', 'armv5l', 'armv6l', 'armv7l', 'armv6el', 'armv6hl', 'armv7el', 'armv7hl', 'armv8el',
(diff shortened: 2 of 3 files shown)


adrianschroeter:
JFYI, you need git master of the build script to get "build --shell" working for chroot

marcus-h:
On 2019-01-09 14:19:08 +0000, Adrian Schröter wrote:
> …and KVM only atm)
> 
> We had bugreports about not matching mount points when using "osc chroot", so using the build script implementation now only. This should work for chroot and KVM at least, but I might have missed some special cases. So please review :)
> 
Great! I'm all for outsourcing this to the build script! (do_chroot is a
mess (especially, if you use "su" as a su-wrapper).)

- it'll always login as root (if I get the obs-build code right):
  * what about adding a new "--shell-cmd <cmd>" option to obs-build? Then,
    osc simply passes "--shell-cmd 'su - abuild'" and we have the old behavior
    (this way, we can easily support the existing --login-as-root option, too)
- --wipe is currently not supported => breaks UI (but this can be easily
  fixed)
- there is no dedicated entry for "chroot" in "osc --help" anymore (it is part
  of the "build" help text). But I can live with that:) (maybe we could mention
  the chroot command in do_build's __docstr__)

adrianschroeter:
Finally had time for it again:

> - it'll always login as root (if I get the obs-build code right):
>  * what about adding a new "--shell-cmd <cmd>" option to obs-build? Then,
>    osc simply passes "--shell-cmd 'su - abuild'" and we have the old behavior
>    (this way, we can easily support the existing --login-as-root option, too)

I have implemented the --shell-cmd option in obs-build and osc now. However, I am 
actually unsure atm if I really want to add the option for "--login-as-root" in osc here,
because it is something the build script should do and not osc.

Atm I made sure that the user gets informed that he is root and give the CLI he
needs to switch to the right user (also something what build script would deside).

I would leave that as incompability atm and wait for feedback...

> - --wipe is currently not supported => breaks UI (but this can be easily  fixed)

also supported build script now.

> - there is no dedicated entry for "chroot" in "osc --help" anymore (it is part
>   of the "build" help text). But I can live with that:) (maybe we could mention
>   the chroot command in do_build's __docstr__)

this is actually wanted, since it is just the for compability, but it should not be documented
anymore. It would just made the already large help section more crowded.

adrianschroeter:
of course you need again the build script version from the morning again for this code.

lethliel:
Looks good now

PR #502, opened 2019-01-10
improve handling of missing progressbar module

lethliel:
With the NoPBTextMeter class the build view gets broken.

Old view:
`1/11 (repo) filename`

New view:
```
Please install the progressbar module...
Please install the progressbar module...
Please install the progressbar module...
```

With this commit the old behavior is restored.

The getbinaries call now lists the file he downloads instead of just stating "Please install the progressbar module..." several times. (but only if not called with the option quiet)

@marcus-h: We can talk about this. But if it is ok for you, pleas merge ;)

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -167,7 +167,8 @@ class Osc(cmdln.Cmdln):
         self.download_progress = None
         if conf.config.get('show_download_progress', False):
             from .meter import TextMeter
-            self.download_progress = TextMeter()
+            if TextMeter:
+                self.download_progress = TextMeter()
 
 
     def get_cmd_help(self, cmdname):
--- a/osc/core.py
+++ b/osc/core.py
@@ -4618,9 +4618,13 @@ def get_binary_file(apiurl, prj, repo, arch,
     progress_obj = None
     if progress_meter:
         from .meter import TextMeter
-        progress_obj = TextMeter()
+        if TextMeter:
+            progress_obj = TextMeter()
 
     target_filename = target_filename or filename
+
+    if progress_meter and not progress_obj:
+        print('Downloading %s' % target_filename)
 
     where = package or '_repository'
     u = makeurl(apiurl, ['build', prj, repo, arch, where, filename])
--- a/osc/fetch.py
+++ b/osc/fetch.py
@@ -25,6 +25,8 @@ import re
 
 from .meter import TextMeter
 
+if not TextMeter:
+    print('Please install the progressbar module')
 
 class Fetcher:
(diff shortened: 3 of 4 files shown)


marcus-h:
On 2019-01-10 05:38:19 -0800, Marco Strigl wrote:
> With the NoPBTextMeter class the build view gets broken.
> 
> Old view:
> `1/11 (repo) filename`
> 
> New view:
> ```
> Please install the progressbar module...
> Please install the progressbar module...
> Please install the progressbar module...
> ```
> 
Arghs... good catch!:)

> With this commit the old behavior is restored.
> 
Hmm this could break the caller in commandline.py. Let's try to
fix this in a more "general" way to avoid too much special handling
at the call sites. What about the following:

- a new function in meter.py:

def create_text_meter(*args, **kwargs):
    use_pb_fallback = kwargs.pop('use_pb_fallback', True)
    if have_pb_module or use_pb_fallback:
        return TextMeter(*args, **kwargs)
    return None

Then, the callers in the commandline and core module simply do:

from meter import create_text_meter
...
progress_obj = create_text_meter()

and the callers in the build and fetch module

from meter import create_text_meter
...
progress_obj = None
if sys.stdout.isatty():
    progress_obj = create_text_meter(use_pb_fallback=False)


What do you think?

lethliel:
@marcus-h: fixed it the simple way for now.

marcus-h:
On 2019-01-11 09:17:05 +0000, Marco Strigl wrote:
> @marcus-h: fixed it the simple way for now. 
> 
Hmm I really dislike the current implementation:
- a print in fetch.py, which is always executed in in case of a missing
  progressbar (an external user might just import the verify_* functions)
- an explicit print in get_binary_file, which is actually the task of
  *some* progress_obj

Anyway, merged...

lethliel:
On 1/14/19 12:43 AM, Marcus Hüwe  wrote:
> On 2019-01-11 09:17:05 +0000, Marco Strigl wrote:
>> @marcus-h: fixed it the simple way for now.
>>
> Hmm I really dislike the current implementation:
> - a print in fetch.py, which is always executed in in case of a missing
> progressbar (an external user might just import the verify_* functions)
> - an explicit print in get_binary_file, which is actually the task of
> *some* progress_obj
> 
> Anyway, merged...
> 

The simple way was just a solution. I prepared another PR with the progressbar object solution:
(The way you suggested is much better. I just had to adapt the start function in the PBNoTextMeter class 
to print the filenames if basename is given.) 

https://github.com/openSUSE/osc/pull/503

I am not sure where to print the warning, that the module is not installed.

PR #503, opened 2019-01-14
reworked meter.py based on discussion in PR 503

lethliel:
as discussed in PR https://github.com/openSUSE/osc/pull/502

* new function create_text_meter with fallback selection
* NoPBTextMeter.start() will print the basename (if not stated otherise with
  basename = None)
* The callers that should use an alternare TextMeter class now call create_text_meter()
* The callers that should not use and alternate TextMeter (because of different handling,
  like build.py) call create_text_meter(use_pb_fallback=False)

@marcus-h: please comment :-)

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -33,7 +33,7 @@ except ImportError:
 
 from .conf import config, cookiejar
 
-from .meter import TextMeter
+from .meter import create_text_meter
 
 change_personality = {
             'i686':  'linux32',
@@ -305,10 +305,9 @@ def get_preinstall_image(apiurl, arch, cache_dir, img_info):
                 print('packagecachedir is not writable for you?', file=sys.stderr)
                 print(e, file=sys.stderr)
                 sys.exit(1)
-        if sys.stdout.isatty() and TextMeter:
-            progress_obj = TextMeter()
-        else:
-            progress_obj = None
+        progress_obj = None
+        if sys.stdout.isatty():
+            progress_obj = create_text_meter(use_pb_fallback=False)
         gr = OscFileGrabber(progress_obj=progress_obj)
         try:
             gr.urlgrab(url, filename=ifile_path_part, text='fetching image')
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -166,9 +166,8 @@ class Osc(cmdln.Cmdln):
         self.options.verbose = conf.config['verbose']
         self.download_progress = None
         if conf.config.get('show_download_progress', False):
-            from .meter import TextMeter
-            if TextMeter:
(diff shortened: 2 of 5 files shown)


marcus-h:
On 2019-01-14 13:10:15 +0000, Marco Strigl wrote:
> as discussed in PR https://github.com/openSUSE/osc/pull/502
> 
Thanks a lot! From a quick glance, this looks so much cleaner!:)

> * NoPBTextMeter.start() will print the basename (if not stated otherise with
>   basename = None)

This should never ever happen - if it does our PBTextMeter would be
broken (also, the empty str would indicate a bug).

> +class NoPBTextMeter(object):
> +
> +    def start(self, basename, size=None):
> +        if basename:
> +            print(basename)

Maybe print('Processing:', basename) (or print('Downloading', basename))?
(Or print('Processing:', basename, '(Please install the progressbar module)')
(see below))

> +        else:
> +            print('Progressbar module missing...')
> +

Maybe we could also print a warning once:

class NoPBTextMeter(object):
    _complained = False

    def start(self, basename, size=None):
        if not self._complained:
            print('Please install the progressbar module')
            NoPBTextMeter._complained = True
        print(...)

    ...

(Long story short: we should get rid of the else-branch:) )

lethliel:
On 1/14/19 5:41 PM, Marcus Hüwe  wrote:
> On 2019-01-14 13:10:15 +0000, Marco Strigl wrote:
>> as discussed in PR https://github.com/openSUSE/osc/pull/502
>>
> Thanks a lot! From a quick glance, this looks so much cleaner!:)
> 
>> * NoPBTextMeter.start() will print the basename (if not stated otherise with
>> basename = None)
> 
> This should never ever happen - if it does our PBTextMeter would be
> broken (also, the empty str would indicate a bug).
> 
>> +class NoPBTextMeter(object):
>> +
>> + def start(self, basename, size=None):
>> + if basename:
>> + print(basename)
> 
> Maybe print('Processing:', basename) (or print('Downloading', basename))?
> (Or print('Processing:', basename, '(Please install the progressbar module)')
> (see below))
> 
>> + else:
>> + print('Progressbar module missing...')
>> +
> 
> Maybe we could also print a warning once:
> 
> class NoPBTextMeter(object):
> _complained = False
> 
> def start(self, basename, size=None):
> if not self._complained:
> print('Please install the progressbar module')
> NoPBTextMeter._complained = True
> print(...)
> 
> ...
> 
> (Long story short: we should get rid of the else-branch:) )

You are right. I got rid of the else-branch and used the self._complained construct to print the warning one time. 

Thanks for the input.

marcus-h:
Merged. Thanks a lot!

PR #504, opened 2019-01-14
Delete some dead code

maltek:


marcus-h:
On 2019-01-14 06:37:16 -0800, maltek wrote:
> -- Commit Summary --
> 
>   * improve os_path_samefile

LGTM.

>   * remove dead pathjoin function

Hmm this changes the API but we could do that.

>   * unpack_srcrpm: delete unused vulnerable code
> 
Hmm this changes the API in an incompatible way: there might some
external module/script that uses unpack_srcrpm's files parameter - we
would break it for no good reason (as long as the script "controls" the
filenames there is no vulnerability).

marcus-h:
PS: thanks a lot for the PR!

maltek:
Oh, I see. So this isn't dead code, but library code. Thanks for the quick response and sorry for the noise!

marcus-h:
On 2019-01-15 04:46:08 -0800, maltek wrote:
> Oh, I see. So this isn't dead code, but library code.

Yes (Admittedly, I think nobody is really using the pathjoin
function, but...).
Nevertheless, your first patch ("improve os_path_samefile") looks
good and is indeed an improvement. Can you submit that one again?:)

> Thanks for the quick response and sorry for the noise!
> 
Nah... no need to apologize - patches are always welcome:)

PR #505, opened 2019-01-15
os_path_samefile: only use fallback if library not available

maltek:
previously, any exception related to e.g. file i/o would also trigger the fallback


marcus-h:
On 2019-01-15 16:24:45 +0000, maltek wrote:
> previously, any exception related to e.g. file i/o would also trigger the fallback

Merged. Thanks a lot!

PR #506, opened 2019-01-15
some improvements to the TLS stack

maltek:
`e11f52cfa61bd5f5e7255d86592ea17453b8e0df`: manually close the TLS connection before raising TLS errors. This is already done in some instances before such an exception is raised, but not in all. I don't really think there's any way those connections could be reused later, but it won't hurt and better safe than sorry.

`a6ee7834894e8d195ea443ba988bda24fd36751b`: here, I moved some code for the TLS validation around. Before, the `verify_cb` allowed connection setup for pretty much any certificate error, but stored these errors for later in order to abort the connection for some fatal errors or ask the user for a decision in case of some other errors. With this change, the `verify_cb` immediately returns an error in case of those fatal errors, so that the encrypted connection is never fully set up. In case of the non-fatal errors, the behavior is unchanged. I'm personally not quite happy how this code (before and after) is kind of circumventing the normal TLS validation flow of OpenSSL, but since overriding TLS errors is a useful feature (at least for testing) and very old Python versions seem to still be supported by this project (with much less useful tls support in the standard library), I don't really see how to avoid that.

`532d90442e9ba3c6f3fc8574d55f00e20290b688`: This removes the hard-coded list of TLS ciphers. That list had some bad stuff in it (RC4 is broken, but thankfully removed in current versions of OpenSSL; and it disabled epheremal Diffie-Hellman). Instead of updating the cipher list to be what's considered good today, I've removed it -- as evidenced here, such lists tend to get out of  date, and the defaults should always be fine. To be completely sure that this doesn't break with weirdly configured servers, I've made the list of TLS ciphers configurable (the `tls_ciphers` option in a server section). That also means people can configure a more paranoid list of ciphers if they want to.

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -203,7 +203,7 @@ boolean_opts = ['debug', 'do_package_tracking', 'http_debug', 'post_mortem', 'tr
     'status_mtime_heuristic']
 integer_opts = ['build-jobs']
 
-api_host_options = ['user', 'pass', 'passx', 'aliases', 'http_headers', 'realname', 'email', 'sslcertck', 'cafile', 'capath', 'trusted_prj']
+api_host_options = ['user', 'pass', 'passx', 'aliases', 'http_headers', 'realname', 'email', 'sslcertck', 'cafile', 'capath', 'trusted_prj', 'tls_ciphers']
 
 new_conf_template = """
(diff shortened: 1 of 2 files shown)


dmach:

PR #509, opened 2019-02-01
deleterequest for entire projects needs the --recursive option as add…

adrianschroeter:
…itional protection

We had it two times this week that people dropped entire projects
(important ones)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,5 @@
 0.165
-  -
+  - deleterequest for entire projects needs the --all option as additional protection
 
 0.164.1
   - rewrite cpio handling to support python3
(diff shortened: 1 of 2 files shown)


lethliel:
@adrianschroeter 

There is already a -r option in line 1940 (for repository)
Perhaps we can use -f (--force)? 


And in the raise you have a typo. ff --> if 
```
if (not opts.recursive) and package == None:
            raise oscerr.WrongOptions('No package name has been provided. Use --recursive option, ff you want to request to delete the entire project.')
```

And one last nitpick: 
Please put the new option to the existing options above the `@cmdln.alias("dr")` line.

adrianschroeter:
right, so let's use "--all" as we do with other commands like restart, wipebinaries and friends

lethliel:
Thanks. Merged.

PR #512, opened 2019-02-12
support different token operations (runservice, release and rebuild) …

adrianschroeter:
…(requires OBS 2.10)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,5 @@
-0.165.2
-  -
+0.166
+  - support different token operations (runservice, release and rebuild) (requires OBS 2.10)
 
 0.165.1
   - fix oscssl "urldefrag is not defined error"
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -728,6 +728,8 @@ class Osc(cmdln.Cmdln):
                         help='Create a new token')
     @cmdln.option('-d', '--delete', metavar='TOKENID',
                         help='Delete a token')
+    @cmdln.option('-o', '--operation', metavar='OPERATION',
+                        help='Default is "runservice", but "release" or "rebuild" can also be used')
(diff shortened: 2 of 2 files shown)


marcus-h:
On 2019-02-12 09:31:32 +0000, Adrian Schröter wrote:
> index e96a1170..8b154c8c 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -744,6 +744,10 @@ def do_token(self, subcmd, opts, *args):
>          if opts.create:
>              print("Create a new token")
>              url += "?cmd=create"
> +            if len(args) == 1:
> +                url += "&operation=" + args[0]
> +            if len(args) == 3:
> +                url += "&operation=" + args[2]

Hmm what about passing the operation via an option? For instance,
"--operation release" etc. If we may support project tokens in the
future, we would have a "consistent" UI
(osc token --create <project> --operation release).

>              if len(args) > 1:
>                  url += "&project=" + args[0]
>                  url += "&package=" + args[1]
> @@ -761,7 +765,10 @@ def do_token(self, subcmd, opts, *args):
>              http_DELETE(url)
>          elif opts.trigger:
>              print("Trigger token")
> -            url = apiurl + "/trigger/runservice"
> +            operation = "runservice"
> +            if len(args) == 2:

Hmm shouldn't this be len(args) == 1?

> +               operation = args[1]

args[0] (see above) + indentation.

marcus-h:
On 2019-03-04 16:02:09 +0000, Adrian Schröter wrote:
> you have a point, code changed.
> 
LGTM. Let's merge it, once the API part is in.

(Hmm maybe we should refactor parts of the do_token code into a
core function (in core.py) + some new sanity checks in the future.)

adrianschroeter:
minor updates .... api side has been merged now.

lethliel:
Then this can be merged too. Thanks for the PR

PR #513, opened 2019-02-12
[python3] cpio.py is now a bytes only api.

lethliel:
The decoding of the header needs to be done in fetch.py.


lethliel:

Issue #516, opened 2019-02-25
'NoneType' object has no attribute 'startswith' when a metalink file is downloaded as rpm


marguerite:
## Steps to reproduce

    osc co X11:Cinnamon:Factory cjs
    osc build openSUSE_Factory x86_64

## Output

<pre>
[1/313] (openSUSE:Factory) libdb-4_8-4.8.30-37.5.x86_64.rpm: 100%|#########################################################| Time: 0:00:00
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/osc/fetch.py", line 146, in fetch
    self.move_package(tmpfile.name, pac.localdir, pac)
  File "/usr/lib/python2.7/site-packages/osc/fetch.py", line 178, in move_package
    fullfilename = os.path.join(destdir, canonname)
  File "/usr/lib64/python2.7/posixpath.py", line 68, in join
    if b.startswith('/'):
AttributeError: 'NoneType' object has no attribute 'startswith'
</pre>

## Reason

http://download.opensuse.org/repositories/openSUSE:/Factory/snapshot/x86_64/libdb-4_8-4.8.30-37.5.x86_64.rpm

This is the file osc trying to download. it can be downloaded via browser, redirected to tumbleweed's version, but will download a .metalink xml file via wget. it has everything a `pac` object needed except it is not a `binary`.

## Analysis

<pre>
    def move_package(self, tmpfile, destdir, pac_obj=None):
        import shutil
        canonname = None
        if pac_obj and pac_obj.name.startswith('container:'):
            canonname = pac_obj.canonname
        if canonname is None:
            pkgq = packagequery.PackageQuery.query(tmpfile, extra_rpmtags=(1044, 1051, 1052))
            if pkgq:
                canonname = pkgq.canonname()
            else:
                if pac_obj is None:
                    print('Unsupported file type: ', tmpfile, file=sys.stderr)
                    sys.exit(1)
                canonname = pac_obj.binary

        fullfilename = os.path.join(destdir, canonname)
</pre>

PackageQuery.query will return None if it met a xml file:

          elif magic[:5] == b'<?xml':
            f.close()
            return None

So we go to the else block. 'pac_obj' is not none if there's file downloaded.

But pac_obj.binary is not always string. If is None if the file is not binary.

And posixpath.py can't process join method with the `b` object as None.

## Workaround

run with '--download-api-only'.

## Possible fix

* Let osc knows redirect
* Fix on download.opensuse.org server
[...]

PR #517, opened 2019-02-25
raise kind error when downloaded a metalink file as rpm

marguerite:
see #516


dmach:
I'm unable to reproduce the issue (no wonder, it's 3 years old).
If it occurs again, let's fix it differently - it would be better to detect the invalid package earlier and fall back to API download automatically.

PR #522, opened 2019-04-07
[python3] fix epoch encoding in repodata.py

lethliel:
other.epoch() needs to be encoded to work with the vercmp callers.


lethliel:

PR #523, opened 2019-04-07
[python3] fetch.py python3 ready

lethliel:
* the result of a division is a float
  but we need int to continue.


lethliel:

PR #524, opened 2019-04-07
[python3] fix decoding for packageqeury.py

lethliel:
name, arch, version and release need to be decoded


lethliel:

PR #525, opened 2019-04-07
[python3] reactivate python3.x tests

lethliel:
The python3.x tests were deactivated because they wouldn't have
worked anyway. This is the last PR that should be activated within
the python3 PR stack.


lethliel:

PR #526, opened 2019-04-07
[python3] fix decoding issue in debquery.py

lethliel:
name, version, release and arch are strings, not bytes


lethliel:

PR #527, opened 2019-04-15
Replace osc-wrapper.py with entry_points.

mcepl:
I think that playing with locale should be unnecessary, unless the configuration of the system is seriously wrong.

This allows simple
```sh
$ python3 setup.py clean build install --user
```
to work as expected.

Also, are you sure you have to force compiled ``.egg`` on everybody?

### DIFF
--- a/osc-wrapper.py
+++ b/osc-wrapper.py
@@ -1,46 +1,9 @@
 #!/usr/bin/env python3
 
-# this wrapper exists so it can be put into /usr/bin, but still allows the
-# python module to be called within the source directory during development
+"""
+This wrapper allows osc to be called from the source directory during development.
+"""
 
-import locale
-import sys
-import os
+import osc.babysitter
 
-from osc import commandline, babysitter
-
-try:
-    # this is a hack to make osc work as expected with utf-8 characters,
-    # no matter how site.py is set...
-    reload(sys)
-    loc = locale.getpreferredencoding()
-    if not loc:
-        loc = sys.getpreferredencoding()
-    sys.setdefaultencoding(loc)
-    del sys.setdefaultencoding
-except NameError:
-    #reload, neither setdefaultencoding are in python3
-    pass
-
-# avoid buffering output on pipes (bnc#930137)
-# Note: the following only applies to python2
-# Basically, a "print('foo')" call is translated to a corresponding
-# fwrite call that writes to the stdout stream (cf. string_print
-# (Objects/stringobject.c) and builtin_print (Python/bltinmodule.c));
-# If no pipe is used, stdout is a tty/refers to a terminal =>
-# the stream is line buffered (see _IO_file_doallocate (libio/filedoalloc.c)).
-# If a pipe is used, stdout does not refer to a terminal anymore =>
-# the stream is fully buffered by default (see _IO_file_doallocate).
-# The following fdopen call makes stdout line buffered again (at least on
-# systems that support setvbuf - if setvbuf is not supported, the stream
-# remains fully buffered (see PyFile_SetBufSize (Objects/fileobject.c))).
-if not os.isatty(sys.stdout.fileno()):
-    sys.stdout = os.fdopen(sys.stdout.fileno(), sys.stdout.mode, 1)
-
-if not os.isatty(sys.stderr.fileno()):
-    sys.stderr = os.fdopen(sys.stderr.fileno(), sys.stderr.mode, 1)
-
-osccli = commandline.Osc()
-
-r = babysitter.run(osccli)
-sys.exit(r)
+osc.babysitter.main()
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 tags
 build
 *junit-xml-results
+*.egg
+osc.egg-info/*
--- a/README.md
+++ b/README.md
@@ -21,8 +21,6 @@ To install from git, do
 
     ./setup.py build
     ./setup.py install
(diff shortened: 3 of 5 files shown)


nickbroon:
With the move fairly standard python3 setup for osc, this seems like a reasonable change to make.

zilti:
Will this happen eventually? :) the current osc-wrapper.py starts to break builds (like e.g. the one for the upcoming Mageia 8).

abitrolly:
This will make it impossible to run development version as `./osc-wrapper.py`, right?

mcepl:
> This will make it impossible to run development version as `./osc-wrapper.py`, right?

Yes, because it is not needed any more. You can use development version installed via `python setup.py devel --user`.

abitrolly:
@mcepl why introduce unnecessary install step if `./osc-wrapper.py` just works without install?

Besides, [PEP-517](https://www.python.org/dev/peps/pep-0517/) shows that Python packaging is moving away from `setup.py install`

mcepl:
> @mcepl why introduce unnecessary install step if `./osc-wrapper.py` just works without install?

Because my system actually works with standard commands? Because the installation of `osc` module has to be done anyway, so the only change is that I won’t have to fiddle with non-standard script in non-standard location?

> Besides, [PEP-517](https://www.python.org/dev/peps/pep-0517/) shows that Python packaging is moving away from `setup.py install`

1. Take a look when this pull request is from and then tell me what was the statement of PEP-517 implementation then.

2. You are welcome to provide your own version using PEP-517 files. You will find out, that you will have to do what I did anyway.

3. Do you understand that your second statement sharply contradicts your first one?

abitrolly:
> Because the installation of osc module has to be done anyway, so the only change is that I won’t have to fiddle with non-standard script in non-standard location? 

Why the installation "has to be done anyway"? I see it is only needed for man pages and docs. Everything else works without extra packaging magic.

> Do you understand that your second statement sharply contradicts your first one?

The first statement is that installation for development is not needed, and `osc-wrapper.py` is convenient, especially for different checkouts with different branches. The second statement is that Python ecosystem is moving away from package definitions and build code in `setup.py`. For example, https://python-poetry.org/docs/pyproject/#scripts These statements may be orthogonal, but not contradicting.

dmach:
@mcepl could you rebase your patch please?
Could you also preserve the following?
* py2 compatibility
* `osc-wrapper.py`
  * just call main() from the new location (babysitter.py)
  * to be used by developers that use git checkout directly (we don't want to break this use case (yet))

mcepl:
> @mcepl could you rebase your patch please? Could you also preserve the following?

Done

dmach:
> Done
You've dropped `osc-wrapper.py` and left some changes from master behind, for example:
```
    if not os.isatty(sys.stderr.fileno()):
        sys.stderr = os.fdopen(sys.stderr.fileno(), sys.stderr.mode, 1)
```
I've prepared a fixup to your commit: https://github.com/dmach/osc/commit/0a1b3969d8812f748588540b8638ef384bd04f8b
Could you check if it makes sense to you and merge it into your commit eventually?

mcepl:
> > Done
> 
> You've dropped `osc-wrapper.py` and left some changes from master behind, for example:

Well, dropping of `osc-wrapper.py` is the whole point of this PR, isn’t it?

> to be used by developers that use git checkout directly (we don't want to break this use case (yet))

`python3 setup.py develop --user`

osc-wrapper.py
```diff
@@ -1,46 +0,0 @@
...
-# If a pipe is used, stdout does not refer to a terminal anymore =>
-# the stream is fully buffered by default (see _IO_file_doallocate).
-# The following fdopen call makes stdout line buffered again (at least on
-# systems that support setvbuf - if setvbuf is not supported, the stream
-# remains fully buffered (see PyFile_SetBufSize (Objects/fileobject.c))).
-if not os.isatty(sys.stdout.fileno()):
-    sys.stdout = os.fdopen(sys.stdout.fileno(), sys.stdout.mode, 1)
-
-if not os.isatty(sys.stderr.fileno()):
-    sys.stderr = os.fdopen(sys.stderr.fileno(), sys.stderr.mode, 1)
```
dmach, line 41: These lines have disappeared from your patch

osc-wrapper.py
```diff
@@ -1,46 +0,0 @@
-#!/usr/bin/env python
-
-# this wrapper exists so it can be put into /usr/bin, but still allows the
-# python module to be called within the source directory during development
```
dmach, line 4: Removing the wrapper is no-go.
We'll use entry point in rpm, but this wrapper script is used by many people that develop osc.
Please keep it as a thin wrapper calling babysitter.main()

dmach (changes requested):

osc-wrapper.py
mcepl: Ehm. Why? What’s missing from now updated PR?

dmach: > Ehm. Why? What’s missing from now updated PR?

The whole osc-wrapper.py is missing.
I know that you can always run `setup.py develop` to achieve the same functionality, but that really doesn't always work for us. Please keep the osc-wrapper.py file around.

dmach (changes requested):
The osc-wrapper.py file is still removed in this pull request, please bring it back.

dmach:
Rebased, squashed commits into one and restored `osc-wrapper.py` for development purposes. Merging.

PR #528, opened 2019-04-16
[python3] add exception if encoding fails and try ISO-8859-1

lethliel:
In some rare cases the chardet encoding detection detects
a wrong encoding standard. Then we switch to latin-1 which
covers most if utf-8 does not work.


lethliel:
This module needs definitely more work, but for now it works ;-)

PR #529, opened 2019-04-16
[python3] return value when adding attrs is bytes

lethliel:
When adding attribute with osc meta attribute <prj> -a <attr> -s '<val>'
the api call gets executed successfull but the return value needs to be
decoded to print it correctly.


lethliel:
decoding issue.

PR #530, opened 2019-04-16
[python3] fix sorting when using osc se

lethliel:
osc se did not sort the output anymore. The logic
in the existing results.sort(key=...) was wrong. Now it is using
key=itemgetter(0,1) has two columns and key=itemgetter(0) if there
is only one column.


lethliel:
needed for python3 support

PR #532, opened 2019-04-17
Also remove groups on copypac

lnussel:


marcus-h:
Hmpf... github ignores my mail replies again...
Thanks for patch and the ping/review!

Issue #373, opened 2018-01-26
Disable non-secure connection warning for non-https connections


jberry-suse:
Unless I am missing something, the whole point of the following warning is in relation to checking ssl certificates and warning you about a self-signed certificates.

https://github.com/openSUSE/osc/blob/5e3fe8ba08d7625140f06441c804e82790ea7db0/osc/conf.py#L563

When running against local OBS instance using their provided docker method there is not an `https` connection, not even a self-signed `https`. I would expect an `apiurl` configured to `http` to not print this message since the user in-fact had to specify the `apiurl` and chose without `https`. Clearly there is no certificate to check so a message indicating certificate check is disabled does not make a lot of sense. Instances should disallow non-https connections as desired, but to have the tool complain _all_ the time is just silly.

If there is some push-back at least provide an option that can be set in config to hide the warning.

When running tests one is inundated with these messages.

```
nosetests --with-coverage --cover-package=. --cover-inclusive
WARNING: SSL certificate checks disabled. Connection is insecure!
WARNING: SSL certificate checks disabled. Connection is insecure!
.WARNING: SSL certificate checks disabled. Connection is insecure!
WARNING: SSL certificate checks disabled. Connection is insecure!
.WARNING: SSL certificate checks disabled. Connection is insecure!
WARNING: SSL certificate checks disabled. Connection is insecure!
.WARNING: SSL certificate checks disabled. Connection is insecure!
WARNING: SSL certificate checks disabled. Connection is insecure!
.WARNING: SSL certificate checks disabled. Connection is insecure!
WARNING: SSL certificate checks disabled. Connection is insecure!
.WARNING: SSL certificate checks disabled. Connection is insecure!
...S..........................WARNING: SSL certificate checks disabled. Connection is insecure!
WARNING: SSL certificate checks disabled. Connection is insecure!
.WARNING: SSL certificate checks disabled. Connection is insecure!
WARNING: SSL certificate checks disabled. Connection is insecure!
WARNING: SSL certificate checks disabled. Connection is insecure!
WARNING: SSL certificate checks disabled. Connection is insecure!
.WARNING: SSL certificate checks disabled. Connection is insecure!
WARNING: SSL certificate checks disabled. Connection is insecure!
...WARNING: SSL certificate checks disabled. Connection is insecure!
..WARNING: SSL certificate checks disabled. Connection is insecure!
.WARNING: SSL certificate checks disabled. Connection is insecure!
[...]

PR #542, opened 2019-05-02
Don't warn about disabled SSL for http URLs

coolo:
During development and testing against a local OBS, there is no point in warning about disabled SSL checks, osc code itself disables sslchk for http URLs.


jberry-suse:
Would seemingly fix #373.

PR #549, opened 2019-05-14
new localbuild command: convinience mapper for building just a local source

adrianschroeter:
Esp. our kernel people do test builds of sources and they struggled with the command line, which requires --noservice meanwhile.

However, this is a valid usecase and we should make this more obvious to use, so I added an alias for it.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,6 @@
+0.166.0
+  - new localbuild command: convinience mapper for building just a local source
+
 0.165.1
   - fix oscssl "urldefrag is not defined error"
 
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -6234,6 +6234,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                   help='trust packages from all projects')
     @cmdln.option('--nopreinstallimage', '--no-preinstallimage', action='store_true',
                   help='Do not use preinstall images for creating the build root.')
+    @cmdln.alias('localbuild')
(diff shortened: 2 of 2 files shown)


adrianschroeter:
question is if we could allow local services instead though, have not checked ...

marcus-h:
On 2019-05-14 00:01:13 -0700, Adrian Schröter wrote:
> Esp. our kernel people do test builds of sources and they struggled with the command line, which requires --noservice meanwhile.
>
> However, this is a valid usecase and we should make this more obvious to use, so I added an alias for it.
>
Hmm IMHO "localbuild" is a misnomer ("osc build" does also a "localbuild":) ).
Why do we need --noservice (just for handling the case if the package does not
exist in the alternative project? (maybe --local-package in this case?))?

> +        if subcmd == 'localbuild':
> +            opts.noservice = True
> +            opts.alternative_project = args[0]

Hmm this could result in an IndexError if no args are passed.

> +            args = args[1:]
> +            if len(args) < 1:
> +                raise oscerr.WrongArgs('locabuild needs at least a project and repository name')
> +

Maybe

if len(args) < 2:
    bail out
opts.alternative_project = args.pop(0)

?

>          if len(args) > 3:
>              raise oscerr.WrongArgs('Too many arguments')

dmach:
Based on the comments, this PR doesn't seem to be completed.
Converting to draft.

lethliel:
Closing due to inactivity

Issue #543, opened 2019-05-02
[py3] Adding a directory results in a backtrace


dcermak:
Running `osc add $directory` results in the following backtrace on my system:
```
$ osc add root
root is a directory, do you want to archive it for submission? (y/n) y
Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.7/site-packages/osc/babysitter.py", line 61, in run
    return prg.main(argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.7/site-packages/osc/commandline.py", line 4588, in do_add
    addFiles([arg])
  File "/usr/lib/python3.7/site-packages/osc/core.py", line 7148, in addFiles
    cpio_proc.communicate('\0'.join(todo))
  File "/usr/lib64/python3.7/subprocess.py", line 924, in communicate
    self._stdin_write(input)
  File "/usr/lib64/python3.7/subprocess.py", line 873, in _stdin_write
    self.stdin.write(input)
TypeError: a bytes-like object is required, not 'str'
1 block
```

Fortunately, the archive `root.obscpio` is still created, I only have to add it manually.

lethliel:
Will fix this when i am back from vacation (Wednesday)

dcermak:
I have to retract this statement: 
> Fortunately, the archive root.obscpio is still created, I only have to add it manually.

The archive is created but it is empty, which is worse than not creating the archive at all.

PR #551, opened 2019-05-14
[python3] fix adding dirs as compressed archive

lethliel:
The content in the todo dict are strings. In python3 the communicate
method expects a bytes-like object not a string.

Solution: Encode every element in todo to a new dict (enc_todo) and
pass this instead of todo

This fixes https://github.com/openSUSE/osc/issues/543

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -7144,10 +7144,11 @@ def addFiles(filenames, prj_obj = None):
         todo = [os.path.join(p, elm)
                 for p, dirnames, fnames in os.walk(filename, followlinks=False)
                 for elm in dirnames + fnames]
+        enc_todo = [b'%s' % elem.encode() for elem in todo]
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2019-05-14 01:29:48 -0700, Marco Strigl wrote:
> index 036b2b0a..0683d0b8 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -7144,10 +7144,11 @@ def addFiles(filenames, prj_obj = None):
>          todo = [os.path.join(p, elm)
>                  for p, dirnames, fnames in os.walk(filename, followlinks=False)
>                  for elm in dirnames + fnames]
> +        enc_todo = [b'%s' % elem.encode() for elem in todo]

Nitpick: could be simplified to [elem.encode() for elem in todo]

Issue #546, opened 2019-05-03
triggerreason unhandled exception on empty package


AdamMajer:
This maybe a little pedantic, but running triggerreason on newly created packages result in unhandled exception instead of nothing.

```
osc triggerreason  devel:languages:nodejs/nodejs12 openSUSE_Tumbleweed        x86_64
https://api.opensuse.org devel:languages:nodejs nodejs12 openSUSE_Tumbleweed x86_64
Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.6/site-packages/osc/babysitter.py", line 61, in run
    return prg.main(argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 5649, in do_triggerreason
    reason = root.find('explain').text
AttributeError: 'NoneType' object has no attribute 'text'

```

lethliel:
Will fix this when i am back from vacation (Wednesday)

PR #552, opened 2019-05-14
fix triggerreason on empty packages.

lethliel:
When running osc triggerreason on newly created (empty) packages
the command failes with AttributeError: 'NoneType' object has no attribute 'text'
because root.find('explain') is NoneType.

Solution:

Check if root.find('explain') is None and print "No triggerreaseon found".
In this case also do not try to get the triggertime. It will result in the same error.

This fixes https://github.com/openSUSE/osc/issues/546

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -5646,9 +5646,13 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         print(apiurl, project, package, repository, arch)
         xml = show_package_trigger_reason(apiurl, project, package, repository, arch)
         root = ET.fromstring(xml)
-        reason = root.find('explain').text
-        triggertime = time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(int(root.find('time').text)))
-        print("%s (at %s)" % (reason, triggertime))
+        if root.find('explain') is None:
+            reason = "No triggerreason found"
+            print(reason)
+        else:
+            reason = root.find('explain').text
+            triggertime = time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(int(root.find('time').text)))
+            print("%s (at %s)" % (reason, triggertime))
         if reason == "meta change":
             print("changed keys:")
             for package in root.findall('packagechange'):


lethliel:
@marcus-h: please review. This is something that is not python3 related ;-)

marcus-h:
On 2019-05-14 08:55:03 +0000, Marco Strigl wrote:
> index 6749d1df..d85e7fa3 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -5646,9 +5646,13 @@ def do_triggerreason(self, subcmd, opts, *args):
>          print(apiurl, project, package, repository, arch)
>          xml = show_package_trigger_reason(apiurl, project, package, repository, arch)
>          root = ET.fromstring(xml)
> -        reason = root.find('explain').text
> -        triggertime = time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(int(root.find('time').text)))
> -        print("%s (at %s)" % (reason, triggertime))
> +        if root.find('explain') is None:
> +            reason = "No triggerreason found"
> +            print("%s" % (reason))

print(reason)?:)

> +        else:
> +            reason = root.find('explain').text

Minor: we could avoid the subsequent root.find if we store the previous
result in a local variable (but OK...).

lethliel:
On 5/14/19 10:24 PM, Marcus Hüwe  wrote:
> On 2019-05-14 08:55:03 +0000, Marco Strigl wrote:
>> diff --git a/osc/commandline.py b/osc/commandline.py
>> index 6749d1df..d85e7fa3 100644
>> --- a/osc/commandline.py
>> +++ b/osc/commandline.py
>> @@ -5646,9 +5646,13 @@ def do_triggerreason(self, subcmd, opts, *args):
>> print(apiurl, project, package, repository, arch)
>> xml = show_package_trigger_reason(apiurl, project, package, repository, arch)
>> root = ET.fromstring(xml)
>> - reason = root.find('explain').text
>> - triggertime = time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(int(root.find('time').text)))
>> - print("%s (at %s)" % (reason, triggertime))
>> + if root.find('explain') is None:
>> + reason = "No triggerreason found"
>> + print("%s" % (reason))
> 
> print(reason)?:)

Yes. That is a bit simpler. :-) 

> 
>> + else:
>> + reason = root.find('explain').text
> 
> Minor: we could avoid the subsequent root.find if we store the previous
> result in a local variable (but OK...).

No. Because I ask for root.find('explain') to be None.
If I would to something like this: 

reason = root.find('explain').text
If reason = None: 
  print(reason)
else:
  ...

This would fail because root.find('explain') is a NoneType and has no attribute text.

marcus-h:
On 2019-05-15 02:56:22 -0700, Marco Strigl wrote:
> On 5/14/19 10:24 PM, Marcus Hüwe  wrote:
> > On 2019-05-14 08:55:03 +0000, Marco Strigl wrote:
> >> diff --git a/osc/commandline.py b/osc/commandline.py
> >> index 6749d1df..d85e7fa3 100644
> >> --- a/osc/commandline.py
> >> +++ b/osc/commandline.py
> >> @@ -5646,9 +5646,13 @@ def do_triggerreason(self, subcmd, opts, *args):
> >> print(apiurl, project, package, repository, arch)
> >> xml = show_package_trigger_reason(apiurl, project, package, repository, arch)
> >> root = ET.fromstring(xml)
> >> - reason = root.find('explain').text
> >> - triggertime = time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(int(root.find('time').text)))
> >> - print("%s (at %s)" % (reason, triggertime))
> >> + if root.find('explain') is None:
> >> + reason = "No triggerreason found"
> >> + print("%s" % (reason))
> >
> > print(reason)?:)
>
> Yes. That is a bit simpler. :-)
>
> >
> >> + else:
> >> + reason = root.find('explain').text
> >
> > Minor: we could avoid the subsequent root.find if we store the previous
> > result in a local variable (but OK...).
>
> No. Because I ask for root.find('explain') to be None.
> If I would to something like this:
>
Yes, that's what you're fixing:)

> reason = root.find('explain').text
> If reason = None:
>   print(reason)
> else:
>   ...
>
> This would fail because root.find('explain') is a NoneType and has no attribute text.
>
What I meant was:

explain = root.find('explain')
if explain is None:
    ...
else:
    reason = explain.text

(that is, you avoid the root.find again (but since the xml is quite small
it should not make a huge difference (wrt. "performance")).

lethliel:
ok. Then I will merge this now.

PR #553, opened 2019-05-14
Improve grammar of request supersede message

sknorr:
Just a grammar issue I always found a bit annoying: The old supersede message mixes singular and plural in the same sentence. This replacement correctly accounts for singular and plural.


marcus-h:
Merged. Thanks a lot!

Issue #540, opened 2019-04-29
UnicodeDecodeError: 'utf-8' codec can't decode byte


pgajdos:
```
$ osc --version
0.165.0
$
```

I wanted to commit php5 for SLE 11 and SLE 10sp3 and I get:

```
$ isc ci
Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.7/site-packages/osc/babysitter.py", line 61, in run
    return prg.main(argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.7/site-packages/osc/commandline.py", line 4700, in do_commit
    self._commit(subcmd, opts, args)
  File "/usr/lib/python3.7/site-packages/osc/commandline.py", line 4792, in _commit
    msg = get_commit_msg(prj.absdir, pac_objs[prj_path])
  File "/usr/lib/python3.7/site-packages/osc/core.py", line 7270, in get_commit_msg
    footer.extend([''.join(i) for i in p.get_diff(ignoreUnversioned=True)])
  File "/usr/lib/python3.7/site-packages/osc/core.py", line 7270, in <listcomp>
    footer.extend([''.join(i) for i in p.get_diff(ignoreUnversioned=True)])
  File "/usr/lib/python3.7/site-packages/osc/core.py", line 2056, in get_diff
    yield get_source_file_diff(self.absdir, f.name, self.rev)
  File "/usr/lib/python3.7/site-packages/osc/core.py", line 4768, in get_source_file_diff
    s1 = f1.readlines()
  File "/usr/lib64/python3.7/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb4 in position 2264: invalid start byte
$

$ file php5.spec
php5.spec: ASCII text, with very long lines
$
```

[...]

PR #554, opened 2019-05-15
fix diff of files with non-ASCII/non-UTF8 chars

lethliel:
When diffing a file that already contains a character that
can not be decoded using ASCII or UTF-8 the diff crashes with
UnicodeDecodeError: 'utf-8' codec can't decode byte ...

The solution:

Open the source file as binary and use difflib.diff_bytes in case
we run python >= 3.0

Before we return the diff list we decode it again.

This fixes https://github.com/openSUSE/osc/issues/540

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -4763,12 +4763,24 @@ def get_source_file_diff(dir, filename, rev, oldfilename = None, olddir = None, 
         return ['Binary file \'%s\' has changed.\n' % origfilename]
 
     f1 = f2 = None
+    mode = 'rt'
+    py3 = compat_mode = None
+    if sys.version_info >= (3, 0):
+        py3 = True
+        mode = 'rb'
+        if sys.version_info < (3, 5):
+            compat_mode = True
+            from osc.util.helper import compat_diff_bytes
+            diff_module = compat_diff_bytes
+        else:
+            diff_module = difflib.diff_bytes
+
     try:
-        f1 = open(file1, 'rt')
+        f1 = open(file1, mode)
         s1 = f1.readlines()
         f1.close()
 
-        f2 = open(file2, 'rt')
+        f2 = open(file2, mode)
         s2 = f2.readlines()
         f2.close()
     finally:
@@ -4777,23 +4789,38 @@ def get_source_file_diff(dir, filename, rev, oldfilename = None, olddir = None, 
         if f2:
             f2.close()
 
-    d = difflib.unified_diff(s1, s2,
-        fromfile = '%s\t(revision %s)' % (origfilename, rev), \
-        tofile = '%s\t(working copy)' % origfilename)
+    if py3:
+        # This is ugly, but needs to be done to support Python 3.3 and 3.4
+        # Python 3.3 and Python 3.4 does not appear to support % byte operator
+        # so it needs to be byte concated
+        if compat_mode:
+            from_file = b'' + origfilename.encode() + b'\t(revision ' + str(rev).encode() + b')'
+            to_file = b'' + origfilename.encode() + b'\t(working copy)'
+        else:
+            from_file = b'%s\t(revision %s)' % (origfilename.encode(), str(rev).encode())
+            to_file = b'%s\t(working copy)' % origfilename.encode()
+        d = diff_module(difflib.unified_diff, s1, s2,
+            fromfile = from_file, \
+            tofile = to_file)
+    else:
+        d = difflib.unified_diff(s1, s2,
+            fromfile = '%s\t(revision %s)' % (origfilename, rev), \
+            tofile = '%s\t(working copy)' % origfilename)
     d = list(d)
     # python2.7's difflib slightly changed the format
     # adapt old format to the new format
+
     if len(d) > 1:
-        d[0] = d[0].replace(' \n', '\n')
-        d[1] = d[1].replace(' \n', '\n')
+        d[0] = d[0].replace(b' \n', b'\n')
+        d[1] = d[1].replace(b' \n', b'\n')
 
     # if file doesn't end with newline, we need to append one in the diff result
     for i, line in enumerate(d):
-        if not line.endswith('\n'):
-            d[i] += '\n\\ No newline at end of file'
+        if not line.endswith(b'\n'):
+            d[i] += b'\n\\ No newline at end of file'
             if i+1 != len(d):
-                d[i] += '\n'
-    return d
+                d[i] += b'\n'
+    return decode_list(d)
 
 def server_diff(apiurl,
                 old_project, old_package, old_revision,
--- a/osc/util/helper.py
+++ b/osc/util/helper.py
@@ -66,4 +66,23 @@ def decode_it(obj):
                 return obj.decode(locale.getlocale()[1])
             except:
                 return obj.decode('latin-1')
-            
+
+def compat_diff_bytes(dfunc, a, b, fromfile=b'', tofile=b'',
+               fromfiledate=b'', tofiledate=b'', n=3, lineterm=b'\n'):
+
+    import difflib
+
+    def decode(s):
+        return s.decode('ascii', 'surrogateescape')
+
+    a = list(map(decode, a))
+    b = list(map(decode, b))
+    fromfile = decode(fromfile)
+    tofile = decode(tofile)
+    fromfiledate = decode(fromfiledate)
+    tofiledate = decode(tofiledate)
+    lineterm = decode(lineterm)
+
+    lines = dfunc(a, b, fromfile, tofile, fromfiledate, tofiledate, n, lineterm)
+    for line in lines:
+        yield line.encode('ascii', 'surrogateescape')


marcus-h:
github ignored my mail reply again... Here's the forwarded message:
```

Date: Thu, 16 May 2019 18:46:16 +0200
To: reply+AAIEUTBFBIHMI2K6Q2N4DHV25EJI5EVBNHHBU7AXIE@reply.github.com
Subject: Re: [openSUSE/osc] fix diff of files with non-ASCII/non-UTF8 chars (#554)
User-Agent: NeoMutt/20180716

On 2019-05-15 09:46:54 +0000, Marco Strigl wrote:
> index 036b2b0a..aeb67ef1 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -4763,12 +4763,24 @@ def get_source_file_diff(dir, filename, rev, oldfilename = None, olddir = None,
>          return ['Binary file \'%s\' has changed.\n' % origfilename]
> 
>      f1 = f2 = None
> +    mode = 'rt'
> +    py3 = compat_mode = None
> +    if sys.version_info >= (3, 0):
> +        py3 = True
> +        mode = 'rb'
> +        if sys.version_info < (3, 5):
> +            compat_mode = True
> +            from osc.util.helper import compat_diff_bytes
> +            diff_module = compat_diff_bytes
> +        else:
> +            diff_module = difflib.diff_bytes
> +

Hmm let's always open the file with mode 'rb'. I think the best is to
move the python3 specific clutter into a/the corresponding helper.

>      try:
> -        f1 = open(file1, 'rt')
> +        f1 = open(file1, mode)
>          s1 = f1.readlines()
>          f1.close()
> 
> -        f2 = open(file2, 'rt')
> +        f2 = open(file2, mode)
>          s2 = f2.readlines()
>          f2.close()
>      finally:
> @@ -4777,23 +4789,38 @@ def get_source_file_diff(dir, filename, rev, oldfilename = None, olddir = None,
>          if f2:
>              f2.close()
> 
> -    d = difflib.unified_diff(s1, s2,
> -        fromfile = '%s\t(revision %s)' % (origfilename, rev), \
> -        tofile = '%s\t(working copy)' % origfilename)
> +    if py3:
> +        # This is ugly, but needs to be done to support Python 3.3 and 3.4
> +        # Python 3.3 and Python 3.4 does not appear to support % byte operator
> +        # so it needs to be byte concated
> +        if compat_mode:
> +            from_file = b'' + origfilename.encode() + b'\t(revision ' + str(rev).encode() + b')'
> +            to_file = b'' + origfilename.encode() + b'\t(working copy)'
> +        else:
> +            from_file = b'%s\t(revision %s)' % (origfilename.encode(), str(rev).encode())
> +            to_file = b'%s\t(working copy)' % origfilename.encode()
> +        d = diff_module(difflib.unified_diff, s1, s2,
> +            fromfile = from_file, \
> +            tofile = to_file)
> +    else:
> +        d = difflib.unified_diff(s1, s2,
> +            fromfile = '%s\t(revision %s)' % (origfilename, rev), \
> +            tofile = '%s\t(working copy)' % origfilename)

Hmm this looks overly complicated.
What about the following:

- provide a helper.unified_diff_bytes function (which expects bytes and
  delegates to difflib.diff_bytes or your own implementation)
- keep the current code as is but call helper.unified_diff_bytes and
  encode all non bytes arguments (so far origfilename is a str, hence,
  we can keep the str constructions)
- return a list/generator of bytes instead of str (see below)


>      d = list(d)
>      # python2.7's difflib slightly changed the format
>      # adapt old format to the new format
> +
>      if len(d) > 1:
> -        d[0] = d[0].replace(' \n', '\n')
> -        d[1] = d[1].replace(' \n', '\n')
> +        d[0] = d[0].replace(b' \n', b'\n')
> +        d[1] = d[1].replace(b' \n', b'\n')
> 
>      # if file doesn't end with newline, we need to append one in the diff result
>      for i, line in enumerate(d):
> -        if not line.endswith('\n'):
> -            d[i] += '\n\\ No newline at end of file'
> +        if not line.endswith(b'\n'):
> +            d[i] += b'\n\\ No newline at end of file'
>              if i+1 != len(d):
> -                d[i] += '\n'
> -    return d
> +                d[i] += b'\n'

OK.

> +    return decode_list(d)
> 
Hmm this is problematic because using the "wrong" decoding can potentially
break the diff (that is, we return garbage that cannot be applied by the
patch program). Probably, the best is to return a list/generator of bytes
here and "adjust" the callers.


>  def server_diff(apiurl,
>                  old_project, old_package, old_revision,
> diff --git a/osc/util/helper.py b/osc/util/helper.py
> index 05808d46..11ed2fe4 100644
> --- a/osc/util/helper.py
> +++ b/osc/util/helper.py
> @@ -66,4 +66,23 @@ def decode_it(obj):
>                  return obj.decode(locale.getlocale()[1])
>              except:
>                  return obj.decode('latin-1')
> -
> +
> +def compat_diff_bytes(dfunc, a, b, fromfile=b'', tofile=b'',
> +               fromfiledate=b'', tofiledate=b'', n=3, lineterm=b'\n'):
> +
> +    import difflib
> +
We should avoid imports in functions (yes... a lot of osc code is still
doing it...:/ )

> +    def decode(s):
> +        return s.decode('ascii', 'surrogateescape')
> +
> +    a = list(map(decode, a))
> +    b = list(map(decode, b))
> +    fromfile = decode(fromfile)
> +    tofile = decode(tofile)
> +    fromfiledate = decode(fromfiledate)
> +    tofiledate = decode(tofiledate)
> +    lineterm = decode(lineterm)
> +
> +    lines = dfunc(a, b, fromfile, tofile, fromfiledate, tofiledate, n, lineterm)
> +    for line in lines:
> +        yield line.encode('ascii', 'surrogateescape')
> 

```

lethliel:
> github ignored my mail reply again... Here's the forwarded message:

I try to answer inline. :-) 
Sorry for the late response. I was ill at Friday and at the moment my whole family is ill. 
But I have to release on monday. So here we go. 

> 
> ```
> ----- Forwarded message from Marcus Hüwe <suse-tux@gmx.de> -----
> 
> Date: Thu, 16 May 2019 18:46:16 +0200
> From: Marcus Hüwe <suse-tux@gmx.de>
> To: reply+AAIEUTBFBIHMI2K6Q2N4DHV25EJI5EVBNHHBU7AXIE@reply.github.com
> Subject: Re: [openSUSE/osc] fix diff of files with non-ASCII/non-UTF8 chars (#554)
> User-Agent: NeoMutt/20180716
> 
> On 2019-05-15 09:46:54 +0000, Marco Strigl wrote:
> > index 036b2b0a..aeb67ef1 100644
> > --- a/osc/core.py
> > +++ b/osc/core.py
> > @@ -4763,12 +4763,24 @@ def get_source_file_diff(dir, filename, rev, oldfilename = None, olddir = None,
> >          return ['Binary file \'%s\' has changed.\n' % origfilename]
> > 
> >      f1 = f2 = None
> > +    mode = 'rt'
> > +    py3 = compat_mode = None
> > +    if sys.version_info >= (3, 0):
> > +        py3 = True
> > +        mode = 'rb'
> > +        if sys.version_info < (3, 5):
> > +            compat_mode = True
> > +            from osc.util.helper import compat_diff_bytes
> > +            diff_module = compat_diff_bytes
> > +        else:
> > +            diff_module = difflib.diff_bytes
> > +
> 
> Hmm let's always open the file with mode 'rb'. I think the best is to
> move the python3 specific clutter into a/the corresponding helper.
> 

Yes. Opening the file with 'rb' works in python2 and 3.

> >      try:
> > -        f1 = open(file1, 'rt')
> > +        f1 = open(file1, mode)
> >          s1 = f1.readlines()
> >          f1.close()
> > 
> > -        f2 = open(file2, 'rt')
> > +        f2 = open(file2, mode)
> >          s2 = f2.readlines()
> >          f2.close()
> >      finally:
> > @@ -4777,23 +4789,38 @@ def get_source_file_diff(dir, filename, rev, oldfilename = None, olddir = None,
> >          if f2:
> >              f2.close()
> > 
> > -    d = difflib.unified_diff(s1, s2,
> > -        fromfile = '%s\t(revision %s)' % (origfilename, rev), \
> > -        tofile = '%s\t(working copy)' % origfilename)
> > +    if py3:
> > +        # This is ugly, but needs to be done to support Python 3.3 and 3.4
> > +        # Python 3.3 and Python 3.4 does not appear to support % byte operator
> > +        # so it needs to be byte concated
> > +        if compat_mode:
> > +            from_file = b'' + origfilename.encode() + b'\t(revision ' + str(rev).encode() + b')'
> > +            to_file = b'' + origfilename.encode() + b'\t(working copy)'
> > +        else:
> > +            from_file = b'%s\t(revision %s)' % (origfilename.encode(), str(rev).encode())
> > +            to_file = b'%s\t(working copy)' % origfilename.encode()
> > +        d = diff_module(difflib.unified_diff, s1, s2,
> > +            fromfile = from_file, \
> > +            tofile = to_file)
> > +    else:
> > +        d = difflib.unified_diff(s1, s2,
> > +            fromfile = '%s\t(revision %s)' % (origfilename, rev), \
> > +            tofile = '%s\t(working copy)' % origfilename)
> 
> Hmm this looks overly complicated.
> What about the following:
> 
> - provide a helper.unified_diff_bytes function (which expects bytes and
>   delegates to difflib.diff_bytes or your own implementation)
> - keep the current code as is but call helper.unified_diff_bytes and
>   encode all non bytes arguments (so far origfilename is a str, hence,
>   we can keep the str constructions)
> - return a list/generator of bytes instead of str (see below)
> 

I will remove the python3.x (where x < 5) parts and then the helper is not needed anymore. 
Nobody should use those versions anyway. 

> 
> >      d = list(d)
> >      # python2.7's difflib slightly changed the format
> >      # adapt old format to the new format
> > +
> >      if len(d) > 1:
> > -        d[0] = d[0].replace(' \n', '\n')
> > -        d[1] = d[1].replace(' \n', '\n')
> > +        d[0] = d[0].replace(b' \n', b'\n')
> > +        d[1] = d[1].replace(b' \n', b'\n')
> > 
> >      # if file doesn't end with newline, we need to append one in the diff result
> >      for i, line in enumerate(d):
> > -        if not line.endswith('\n'):
> > -            d[i] += '\n\\ No newline at end of file'
> > +        if not line.endswith(b'\n'):
> > +            d[i] += b'\n\\ No newline at end of file'
> >              if i+1 != len(d):
> > -                d[i] += '\n'
> > -    return d
> > +                d[i] += b'\n'
> 
> OK.
> 
> > +    return decode_list(d)
> > 
> Hmm this is problematic because using the "wrong" decoding can potentially
> break the diff (that is, we return garbage that cannot be applied by the
> patch program). Probably, the best is to return a list/generator of bytes
> here and "adjust" the callers.
> 

I don't understand what the difference is if I decode here or If I decode in the 
do_diff / do_commit caller. They use both chardet to "guess" the decoding. 

Moving the decoding in the caller just makes it more complicated. 

> 
> >  def server_diff(apiurl,
> >                  old_project, old_package, old_revision,
> > diff --git a/osc/util/helper.py b/osc/util/helper.py
> > index 05808d46..11ed2fe4 100644
> > --- a/osc/util/helper.py
> > +++ b/osc/util/helper.py
> > @@ -66,4 +66,23 @@ def decode_it(obj):
> >                  return obj.decode(locale.getlocale()[1])
> >              except:
> >                  return obj.decode('latin-1')
> > -
> > +
> > +def compat_diff_bytes(dfunc, a, b, fromfile=b'', tofile=b'',
> > +               fromfiledate=b'', tofiledate=b'', n=3, lineterm=b'\n'):
> > +
> > +    import difflib
> > +
> We should avoid imports in functions (yes... a lot of osc code is still
> doing it...:/ )

Just out of curiosity: Why is this bad? For my understanding it is better to import function related modules in the function. Otherwise if I call osc.helper.decode_it the difflib module will be imported too. 
(Or do I get something wrong here) 

Isn't this something one want to avoid?

> 
> > +    def decode(s):
> > +        return s.decode('ascii', 'surrogateescape')
> > +
> > +    a = list(map(decode, a))
> > +    b = list(map(decode, b))
> > +    fromfile = decode(fromfile)
> > +    tofile = decode(tofile)
> > +    fromfiledate = decode(fromfiledate)
> > +    tofiledate = decode(tofiledate)
> > +    lineterm = decode(lineterm)
> > +
> > +    lines = dfunc(a, b, fromfile, tofile, fromfiledate, tofiledate, n, lineterm)
> > +    for line in lines:
> > +        yield line.encode('ascii', 'surrogateescape')
> > 
> 
> ----- End forwarded message -----
> ```

marcus-h:
On 2019-05-18 12:55:29 -0700, Marco Strigl wrote:
> > github ignored my mail reply again... Here's the forwarded message:
>
> I try to answer inline. :-)
> Sorry for the late response. I was ill at Friday and at the moment my whole family is ill.

Get well soon!

> > ----- Forwarded message from Marcus Hüwe <suse-tux@gmx.de> -----
> > On 2019-05-15 09:46:54 +0000, Marco Strigl wrote:

<SNIP>

> > > +    return decode_list(d)
> > >
> > Hmm this is problematic because using the "wrong" decoding can potentially
> > break the diff (that is, we return garbage that cannot be applied by the
> > patch program). Probably, the best is to return a list/generator of bytes
> > here and "adjust" the callers.
> >
>
> I don't understand what the difference is if I decode here or If I decode in the
> do_diff / do_commit caller. They use both chardet to "guess" the decoding.
>
Decoding + print()ing the diff result is "wrong". The following example
demonstrates this:

```
marcus@linux:~/dif> cat d.py
import chardet
import sys
import difflib


def create_orig_file():
    data = b'''foo
bar
foobar
'''
    with open('orig', 'wb') as f:
        f.write(data)


def create_new_file():
    data = b'''foo
bar
a: \xc2
foobar
'''
    with open('new', 'wb') as f:
        f.write(data)  # ignore retval - keep it simple...


def diff():
    with open('orig', 'rb') as orig, open('new', 'rb') as new:
        odata = orig.readlines()
        ndata = new.readlines()
    d = difflib.diff_bytes(difflib.unified_diff, odata, ndata,
                           fromfile=b'orig', tofile=b'new')
    return b''.join(d)


def decode_via_chardet(b):
    return b.decode(chardet.detect(b)['encoding'])


if __name__ == '__main__':
    if len(sys.argv) != 2 or sys.argv[1] not in ('raw', 'chardet'):
        print("Usage: {} raw|chardet".format(sys.argv[0]), file=sys.stderr)
        sys.exit(1)
    create_orig_file()
    create_new_file()
    d = diff()
    if sys.argv[1] == 'raw':
        sys.stdout.buffer.write(d)
    else:
        sys.stdout.write(decode_via_chardet(d))
marcus@linux:~/dif> python3 d.py chardet | md5sum
6f705a3a225ca1692138f318616e0ff4  -
marcus@linux:~/dif> python3 d.py raw | md5sum
420ba4bf4d90641ddd478354095be929  -
marcus@linux:~/dif> diff -u --label orig --label new orig new | md5sum
420ba4bf4d90641ddd478354095be929  -
marcus@linux:~/dif>
```

That is, decoding + print()ing can result in an incorrect diff.

> Moving the decoding in the caller just makes it more complicated.
>
Yes... :/

<SNIP>

> > > +def compat_diff_bytes(dfunc, a, b, fromfile=b'', tofile=b'',
> > > +               fromfiledate=b'', tofiledate=b'', n=3, lineterm=b'\n'):
> > > +
> > > +    import difflib
> > > +
> > We should avoid imports in functions (yes... a lot of osc code is still
> > doing it...:/ )
>
> Just out of curiosity: Why is this bad?

It is not bad - it is more a matter of style. PEP8 says/dictates that
imports should be at the top of the file.

> For my understanding it is better to import function related modules in the function. Otherwise if I call osc.helper.decode_it the difflib module will be imported too.
> (Or do I get something wrong here)
>
> Isn't this something one want to avoid?
>
Hmm it depends I guess... if the difflib module would run some expensive
initialization code, then it is probably a good idea to import it only
if needed.

frispete:
Since it bites TW users of osc on a regular base, it would be nice to have this fixed and released rather sooner than later.

Latest occurrence for me: UserManual.pdf from Virtualization/virtualbox revision 487 -> 488 (version 6.0.6 -> 6.0.8). Diffing pdfs and expecting a utf-8 save outcome is somewhat adventurous. ;-)

lethliel:
@frispete I know that this is kind of annoying at the moment. I was at the oSC'19 over the weekend and could not have a look at this problem so far. 
The complete diff handling needs to be rewritten to fully satisfy all python3 encoding/decoding corner cases. 
It will be part of 0.166.0

marcus-h:
On 2019-05-28 02:12:04 -0700, Hans-Peter Jansen wrote:
> Since it bites TW users of osc on a regular base, it would be nice to have this fixed and released rather sooner than later.
>
Hmm maybe we should/could also ship the python2 version of osc.
(Actually, I still don't understand why we only ship the python3
version:/)

lethliel:
Sorry for the long delay. I was sick the last 1 1/2 weeks and I am still recovering. 

@marcus-h: We decided to ship only the python3 version for newer distributions because python2 will no longer be part of the distributions (at least i was told so at that point)

But feel free to change the packaging.

lethliel:
https://github.com/openSUSE/osc/pull/598 is a better approach and this PR will be closed.

PR #555, opened 2019-05-15
fix getbinaries quiet logic

lethliel:
opts.quiet is passed to get_binary_file which is False by default.
So the following `if progress_meter` is always False unless -q is given.

So it needs to be `if not progress_meter`.

This was reported by @bugfinder

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -4665,7 +4665,7 @@ def get_binary_file(apiurl, prj, repo, arch,
                     target_mtime = None,
                     progress_meter = False):
     progress_obj = None
-    if progress_meter:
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2019-05-15 08:26:46 -0700, Marco Strigl wrote:
> opts.quiet is passed to get_binary_file which is False by default.
> So the following `if progress_meter` is always False unless -q is given.
>
> So it needs to be `if not progress_meter`.
>
Hmm no. The bug is in the caller (see do_getbinaries). We have
to fix the logic there. This change breaks the caller in the build
module.

Can we please revert this commit?:)

PR #558, opened 2019-05-16
fix getbinaries in commandline (progress_meter)

lethliel:
progress_meter = opts.quiet is wrong, because if -q is not
set progress_meter will be false and no output will be shown.

This is fixed by using a new var use_progress_meter. 

This fixes an issue reported by @bugfinder

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -7289,7 +7289,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                                     package = pac,
                                     target_filename = fname,
                                     target_mtime = i.mtime,
-                                    progress_meter = opts.quiet)
+                                    progress_meter = not opts.quiet)
 
 
     @cmdln.option('-b', '--bugowner', action='store_true',


lethliel:
@marcus-h: This should be the right place to fix this :-)

marcus-h:
On 2019-05-16 04:23:41 -0700, Marco Strigl wrote:
> index 6749d1df..880e8091 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -7203,6 +7203,10 @@ def do_getbinaries(self, subcmd, opts, *args):
>          package = None
>          binary = None
>
> +        use_progress_meter = True
> +        if opts.quiet:
> +            use_progress_meter = False
> +
>          if len(args) < 1 and is_package_dir('.'):
>              self.print_repos()
>
> @@ -7289,7 +7293,7 @@ def do_getbinaries(self, subcmd, opts, *args):
>                                      package = pac,
>                                      target_filename = fname,
>                                      target_mtime = i.mtime,
> -                                    progress_meter = opts.quiet)
> +                                    progress_meter = use_progress_meter)

Hmm correct, but quite "complicated". If you insist on a local
variable just initialize it by inverting opts.quiet:

    use_progress_meter = not opts.quiet

But I would simply move this to the get_binary_file call:

    get_binary_file(..., progress_meter=not opts.quiet)

That's it:)

lethliel:
@marcus-h You are right. Was thinking way to complicated. 
switched to progress_meter = not opts.quiet

PR #564, opened 2019-05-24
Add "prioritize" as (preferred) alias for "priorize" command

sknorr:
"priorize" is not an English word, "prioritize" is the correct
version:
* https://www.merriam-webster.com/dictionary/priorize
* https://www.merriam-webster.com/dictionary/prioritize

The "priorize" command will continue to exist after this commit, but the
documented  version will be "prioritize".

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -2148,7 +2148,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
 
         "checkout" will checkout the request's source package ("submit" requests only).
 
-        "priorize" change the prioritity of a request to either "critical", "important", "moderate" or "low"
+        "prioritize" change the priority of a request to either "critical", "important", "moderate" or "low"
 
 
         The 'review' command has the following sub commands:
@@ -2177,7 +2177,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             osc request setincident [-m TEXT] ID INCIDENT
             osc request supersede [-m TEXT] ID SUPERSEDING_ID
             osc request approvenew [-m TEXT] PROJECT
-            osc request priorize [-m TEXT] ID PRIORITY
+            osc request prioritize [-m TEXT] ID PRIORITY
 
             osc request checkout/co ID
             osc request clone [-m TEXT] ID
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2019-05-24 04:28:37 -0700, Stefan Knorr wrote:
> Subject: [PATCH 1/2] Add "prioritize" as (preferred) alias for "priorize"
>  command
>

This patch looks perfect. Thanks!

> Subject: [PATCH 2/2] Random drive-by typo fixes for osc command-line messages
>
> ---
>  osc/commandline.py | 42 +++++++++++++++++++++---------------------
>  1 file changed, 21 insertions(+), 21 deletions(-)
>
> index 888be1d9..048135fa 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -1,6 +1,6 @@
>  # Copyright (C) 2006 Novell Inc.  All rights reserved.
>  # This program is free software; it may be used, copied, modified
> -# and distributed under the terms of the GNU General Public Licence,
> +# and distributed under the terms of the GNU General Public License,

Hmm "Licence" (BE) is OK. I'm not sure if we should change it here because
it looks like some "standard" copyright bla...


> @@ -4189,7 +4189,7 @@ def _prdiff_output_matching_requests(self, opts, requests,
>      def do_prdiff(self, subcmd, opts, *args):
>          """${cmd_name}: Server-side diff of two projects
>
> -        Compares two projects and either summarises or outputs the
> +        Compares two projects and either summarizes or outputs the

Actually, "summarises" (BE) is OK, but I don't mind changing to the
AE variant.

> @@ -6244,7 +6244,7 @@ def do_build(self, subcmd, opts, *args):
>          of the 'osc repos' output. BUILD_DESCR is either a RPM spec file, or a
>          Debian dsc file.
>
> -        The command honours packagecachedir, build-root and build-uid
> +        The command honors packagecachedir, build-root and build-uid

As above.

Could you revert the "Licence" -> "License" part of your patch? Otherwise
LGTM. Thanks a lot for the patches!

sknorr:
Hi Marcus, thanks! --
I just changed the "licence"/"license" thing back the way it was. The FSF is a US organization, so their spelling of the word is ["license"](https://www.gnu.org/licenses/licenses.html#GPL). But I don't fancy changing all the headers in the osc repo to use the American spelling. :)

In any case, it seems osc might have once started with BE messages but right now, it's not particularly consistent (i.e. there are plenty uses of "initialize", "serialize" etc. already). I could reverse my other BE->AmE changes as well, I guess.

marcus-h:
On 2019-05-27 04:32:25 -0700, Stefan Knorr wrote:
> I just changed the "licence"/"license" thing back the way it was. The FSF is a US organization, so their spelling of the word is ["license"](https://www.gnu.org/licenses/licenses.html#GPL).

Hmm fair point.

> But I don't fancy changing all the headers in the osc repo to use the American spelling. :)
>
> In any case, it seems osc might have once started with BE messages but right now, it's not particularly consistent (i.e. there are plenty uses of "initialize", "serialize" etc. already). I could reverse my other BE->AmE changes as well, I guess.
>
I'm fine with the BE->AmE changes. Is there anything you still want to
change? If not, we can merge the PR.

sknorr:
@marcus-h No more changes, please merge. If I come across more things, I will do another PR later.

marcus-h:
On 2019-05-28 01:23:10 -0700, Stefan Knorr wrote:
> @marcus-h No more changes, please merge. If I come across more things, I will do another PR later.
>
Merged. Thanks!

PR #565, opened 2019-05-25
Fix encode method call

bmwiedemann:
This broke with
--define="%_buildhost foo"


marcus-h:
Merged. Thanks a lot!

PR #568, opened 2019-05-28
cast bytes object to string during os.path.join

MaximilianMeister:
this makes sure not to mix objects to prevent a TypeError

TypeError: Can't mix strings and bytes in path components

Fixes#567

Signed-off-by: Maximilian Meister <mmeister@suse.de>

### DIFF
--- a/osc/fetch.py
+++ b/osc/fetch.py
@@ -176,7 +176,7 @@ class Fetcher:
                     sys.exit(1)
                 canonname = pac_obj.binary
 
-        fullfilename = os.path.join(destdir, canonname)
+        fullfilename = os.path.join(destdir, decode_it(canonname))
         if pac_obj is not None:
             pac_obj.canonname = canonname
             pac_obj.fullfilename = fullfilename


osc/fetch.py
```diff
@@ -176,7 +176,7 @@ def move_package(self, tmpfile, destdir, pac_obj=None):
                     sys.exit(1)
                 canonname = pac_obj.binary
 
-        fullfilename = os.path.join(destdir, canonname)
+        fullfilename = os.path.join(destdir, canonname.decode("utf-8"))
```
lethliel: Can you please use the helper function `decode_it()` from util/helper.py?

Like this:
`fullfilename = os.path.join(destdir, decode_it(canonname))`

MaximilianMeister: sure :+1: wasn't aware that there's a helper. fixed now

lethliel: Thank you very much for this catch.

lethliel:
Thanks for the PR

MaximilianMeister:
> Thanks for the PR

thanks for the quick review

marcus-h:
On 2019-05-28 02:43:53 -0700, Maximilian Meister wrote:
> this makes sure not to mix objects to prevent a TypeError
>
> TypeError: Can't mix strings and bytes in path components
>
> Fixes#567
>
s/Fixes/Works around/
Actually, we should work with bytes there - that's what the cannoname()
method is supposed to return (but for most *Query classes it returns a
str atm, which is wrong).
_If_ the method should work with strs, we should better move the
decode_it call to line 168 or preferrably to line 85.

MaximilianMeister:
@marcus-h if you think it's not fixed, then please re-open the issue with the comment. commenting on a closed PR is not very effective

Issue #569, opened 2019-05-28
osc meta -e pkg <name> returns bytes as output


lethliel:
When running into an error using osc meta the return text is bytes: 

```
Try again? ([y/N]): y
Sending meta data...
BuildService API error: validation_failed (400)
b'package validation error: 4:0: ERROR: Element package has extra content: disable'
```

marcus-h:
On 2019-05-28 04:59:36 -0700, Marco Strigl wrote:
> When running into an error using osc meta the return text is bytes:
>
> ```
> Try again? ([y/N]): y
> Sending meta data...
> BuildService API error: validation_failed (400)
> b'package validation error: 4:0: ERROR: Element package has extra content: disable'
> ```
>
Fixed via PR#571 ("Properly parse error message in metafile.edit").

PR #570, opened 2019-05-28
decode retrundata before printing it (osc meta -e)

lethliel:
Otherwise the return text of a failed meta -e call will look
like this:

```
Sending meta data...
BuildService API error: validation_failed (400)
b'package validation error: 8:12: FATAL: ...'
Try again? ([y/N]):
```

This is not critical but it looks nicer without the b''

Fixes https://github.com/openSUSE/osc/issues/569

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -3697,9 +3697,9 @@ class metafile:
                     print('BuildService API error:', error_help, file=sys.stderr)
                     # examine the error - we can't raise an exception because we might want
                     # to try again
-                    data = e.read()
-                    if b'<summary>' in data:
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2019-05-28 05:20:54 -0700, Marco Strigl wrote:
> Otherwise the return text of a failed meta -e call will look
> like this:
>
> ```
> Sending meta data...
> BuildService API error: validation_failed (400)
> b'package validation error: 8:12: FATAL: ...'
> Try again? ([y/N]):
> ```
>
> This is not critical but it looks nicer without the b''
>
Definitely:)
Fixed via PR#571.

PR #573, opened 2019-05-31
Do not attempt to run source services when --local-package is set

### LINKED ISSUE #572: osc build --local-package does not work with master
With 0.165.1 and master, when I try to use `osc build --alternative-project=home:Pharaoh_Atem CentOS_7 x86_64 --no-verify --local-package`, it does not work.

I get messages like the following:
```
Error: "/home/ngompa/Sources/package-backports/bcc" is not an osc package working copy.
```

With 0.164.2, this worked. My guess is that https://github.com/openSUSE/osc/commit/c39c3b8cae826723c23e5115e5c051438d7aa8ba broke it, but I'm not sure.

Conan-Kudo:
Since c39c3b8cae826723c23e5115e5c051438d7aa8ba, osc has accidentally refused
to permit local builds for package sources that are not from a working copy,
even when `--local-package` is passed in. This is because it attempts to
instantiate a `Package()` object, which fails because there's no osc working
copy to instantiate from.

Since any build passing `--local-package` probably can't run source services
anyway, let's restore the capability to build with it by having it skip
any attempt to run them automatically when the option is passed.

Fixes #572.

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -727,7 +727,7 @@ def main(apiurl, opts, argv):
         buildargs.append('--noinit')
 
     # check for source services
-    if not opts.offline and not opts.noservice:
+    if not opts.offline and not opts.noservice and not opts.local_package:
         p = osc.core.Package(os.curdir)
         r = p.run_source_services(verbose=True)
         if r:


marcus-h:
On 2019-05-30 19:25:31 -0700, Neal Gompa (ニール・ゴンパ) wrote:
> Since c39c3b8cae826723c23e5115e5c051438d7aa8ba, osc has accidentally refused
> to permit local builds for package sources that are not from a working copy,
> even when `--local-package` is passed in. This is because it attempts to
> instantiate a `Package()` object, which fails because there's no osc working
> copy to instantiate from.
>
Use the "--noservice" option. Alternatively, we could check if the package
package is a package wc.

> Since any build passing `--local-package` probably can't run source services
> anyway,

Huh? There is no reason why --local-package should imply --noservice.

Conan-Kudo:
@marcus-h You're right, but I couldn't figure out how to fix it so that `--local-package` still ran source services. The logic seemed to indicate that it's not supposed to work if it's not associated with an osc working copy, so I left it be.

As an aside, `--local-package` functionality has been ping-ponging between working and broken for several releases now, and it's extremely annoying.

Conan-Kudo:
@adrianschroeter This is probably a better alternative to #549.

lethliel:
@Conan-Kudo @marcus-h 

Wouldn't that do the trick? 
```
if is_package_dir('.'):
        p = Package('.')
        r = p.run_source_services(verbose=True)
        if r:
                print('Source service run failed!', file=sys.stderr)
                sys.exit(1)
else:
        opts.local_package = True
```

marcus-h:
On 2019-07-29 06:01:58 -0700, Marco Strigl wrote:
> @Conan-Kudo @marcus-h
>
> Wouldn't that do the trick?

It depends a bit on the context:
- if this is going to replace the current service execution code in
  do_build, then no

> ```
> if is_package_dir('.'):
>         p = Package('.')
>         r = p.run_source_services(verbose=True)
>         if r:
>                 print('Source service run failed!', file=sys.stderr)
>                 sys.exit(1)
> else:
>         opts.local_package = True
> ```

We could do the following:

if not is_package_dir('.'):
    opts.noservice = True

# insert current service execution code here...

Conan-Kudo:
@lethliel The problem is when you're working from a Git checkout. At `$DAYJOB`, we store all our packaging content in Git repos, and have hooks and jobs set up to do the equivalent of Koji scratch builds and do pushes to the build system on tagging for release.

However, `Package('.')` also fails if it detects _any other SCM content_ in there. This causes everything to fail and bomb out. We work around it by declaring `--local-package`, but then *that* broke (for the Nth time...).

Right now, I'm shipping this as a patch in the Fedora package for our CI, but I'd *really* rather this be fixed upstream.

@marcus-h's suggestion is functionally similar to this patch, except that it'll not require setting `--local-package`, which might lead to unexpected effects... IMHO, it's non-obvious behavior that failure to initialize a working copy means that source services aren't running.

That said, the real issue is that source services are tied to osc working copies, which doesn't make sense, but that's a bigger problem...

Conan-Kudo:
I've rebased this on top of current git master.

marcus-h:
Hmm github ate my mail again:

```

Date: Sat, 14 Aug 2021 16:28:45 +0200
To: reply+AAIEUTDHEP2TYO6R2MRPSSV4CNTC7EVBNHHBVW3DS4@reply.github.com
Subject: Re: [openSUSE/osc] Do not attempt to run source services when
 --local-package is set (#573)

We went with the

> if not is_package_dir('.'):
>     opts.noservice = True

approach (see commit b61b79cf85f73130a687d4b8793fac2c330500bf ("Don't run
source services when building outside of an OSC package working copy")
and [1] for its discussion/rationale).

[1] https://github.com/openSUSE/osc/pull/937

```

Issue #579, opened 2019-06-11
osc lock <project> is broken


lethliel:
Running osc lock to lock a project crashes

```
marco@linux-k38d:~/obs> osc lock home:mstrigl
Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.6/site-packages/osc/babysitter.py", line 61, in run
    return prg.main(argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 3669, in do_lock
    edit_meta(kind, path_args=path_args, data=meta, msg=opts.message)
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 3841, in edit_meta
    f = metafile(url, data, change_is_required, metatypes[metatype]['file_ext'])
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 3668, in __init__
    f.write(b''.join(input))
TypeError: sequence item 0: expected a bytes-like object, int found
```

PR #580, opened 2019-06-11
simplified class metafile __init__ method

lethliel:
simplified and fixed the `__init__` method of the metafile class.
input can be string, bytes-like object, list of strings or list
of bytes-like objects now.

Based on the input now always a list is generated and joined to
a string for writing in the fd. (This is ugly but needed for
compat reasons)

fixes https://github.com/openSUSE/osc/issues/579
Thanks to @darix

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -3660,12 +3660,19 @@ class metafile:
         self.url = url
         self.change_is_required = change_is_required
         (fd, self.filename) = tempfile.mkstemp(prefix = 'osc_metafile.', suffix = file_ext)
-        if not input or isinstance(input[0], str) or isinstance(input, str):
-            f = os.fdopen(fd, 'w')
-            f.write(''.join(input))
+
+        open_mode = 'w'
+        input_as_str = None
+
+        if not isinstance(input, list):
+            input = [input]
+        if isinstance(input[0], str):
+            input_as_str = ''.join(input)
         else:
-            f = os.fdopen(fd, 'wb')
-            f.write(b''.join(input))
+            open_mode = 'wb'
+            input_as_str = b''.join(input)
+        f = os.fdopen(fd, open_mode)
+        f.write(input_as_str)
         f.close()
         self.hash_orig = dgst(self.filename)
 


marcus-h:
On 2019-06-11 06:19:12 -0700, Marco Strigl wrote:
> simplified and fixed the `__init__` method of the metafile class.
> input can be string, bytes-like object, list of strings or list
> of bytes-like objects now.
>
> Based on the input now always a list is generated and joined to
> a string for writing in the fd. (This is ugly but needed for
> compat reasons)
>
Merged. Thanks!
(JFTR: the old code could potentially break if input == b'', which
is fixed in the new code - but the new code could break if input == [],
where the old code didn't break (osc itself shouldn't pass such
inputs, hence: merged).)

darix:
Just curious wouldnt it make sense to have a method like this and share it among all places that want to write a "string" to a file?

```python
  def write_something_like_string_to_file(fd, input):
        open_mode = 'w'
        input_as_str=None

        if not(isinstance(input, list)):
           input = [input]

        if isinstance(input[0], str):
            input_as_str=''.join(input)
        else:
            open_mode = 'wb'
            input_as_str=b''.join(input)

        f = os.fdopen(fd, open_mode)
        f.write(input_as_str)
        f.close()
```

marcus-h:
On 2019-06-12 03:53:55 -0700, darix wrote:
> Just curious wouldnt it make sense to have a method like this and share it among all places that want to write a "string" to a file?

Hmm actually I'm not fond of such function. We should rather
(try to) get our datatypes right. In case of the metafile class,
we would ideally do something like

mf = metafile(some_input_as_str, mode='w')
# work with mf

mf = metafile(some_input_as_bytes, mode='wb')
# work with mf

That is, it should be callers duty to specify the correct
mode etc. (The "strange" list handling is only needed
for the metafile class because the meta_exists function
returns a list...)

>
> ```python
>   def write_something_like_string_to_file(fd, input):
>         open_mode = 'w'
>         input_as_str=None
>
>         if not(isinstance(input, list)):
>            input = [input]
>
>         if isinstance(input[0], str):
>             input_as_str=''.join(input)
>         else:
>             open_mode = 'wb'
>             input_as_str=b''.join(input)
>
>         f = os.fdopen(fd, open_mode)
>         f.write(input_as_str)
>         f.close()
> ```

darix:
well yeah not using bytes and strings at the same time anymore would be 
even better. stringify everything!

PR #590, opened 2019-06-27
fix local build using simpleimage

adrianschroeter:


osc/build.py
```diff
@@ -695,7 +702,7 @@ def main(apiurl, opts, argv):
         s = ''
         for i in opts.define:
             s += "%%define %s\n" % i
-        build_descr_data = s.encode() + build_descr_data
+        build_descr_data = s.encode + build_descr_data
```
lethliel: can you please add the () to the encode method call? Otherwise it breaks when using build --define ...

lethliel:
@adrianschroeter 

Can you please add the brackets back to the .encode method call in the define case? 
Otherwise this change will break all osc build --define .... calls.

lethliel:
Thanks. Looks good now.

PR #599, opened 2019-07-17
remove python 3.3 and 3.4 from testing

lethliel:
Those versions are not supported anymore and thus testing
is not needed anymore.


marcus-h:
On 2019-07-17 02:36:22 -0700, Marco Strigl wrote:
> Those versions are not supported anymore and thus testing
> is not needed anymore.

Merged. Thanks!

Issue #602, opened 2019-07-18
can't fetch rpm with osc api call with osc > 0.165.0  


mgoppold:
with osc from Leap 15.1 I can't mirror projects with obs_mirror_project due to wrong file type of downloaded files:

```
# osc -A https://api.opensuse.org api -m GET /build/openSUSE:Leap:15.1:Update/standard/x86_64/_repository/osc.rpm > /tmp/osc_0.165.1.rpm
# file /tmp/osc_0.165.1.rpm
/tmp/osc_0.165.1.rpm: data
```

with osc 0.162.1 from Leap 15.0 I get:

```
# osc -A https://api.opensuse.org api -m GET /build/openSUSE:Leap:15.1:Update/standard/x86_64/_repository/osc.rpm > /tmp/osc.rpm
# file /tmp/osc.rpm
/tmp/osc.rpm: RPM v3.0 bin i386/x86_64
```

lethliel:
Thanks for the issue. I can confirm this. Will investigate and fix it.

PR #603, opened 2019-07-19
Do not decode output from osc api calls

lethliel:
Do not guess the decoding of the returned output on osc api calls.
This could lead to unexptected behavior. Just use sys.stdout.buffer.write()
to print the bytes-like object in python3.

fixes https://github.com/openSUSE/osc/issues/602


marcus-h:
Yes! That's the kind of patch I like:) (getting rid of an
explicit decode if we cannot say anything/much about the content.)
Happily merged:)

Issue #600, opened 2019-07-18
backtrace for '-h' 


rfrohl:
```
$ osc -h
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/osc/babysitter.py", line 61, in run
    return prg.main(argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 325, in main
    self.options, args = self.optparser.parse_args(argv[1:])
  File "/usr/lib64/python3.6/optparse.py", line 1387, in parse_args
    stop = self._process_args(largs, rargs, values)
  File "/usr/lib64/python3.6/optparse.py", line 1431, in _process_args
    self._process_short_opts(rargs, values)
  File "/usr/lib64/python3.6/optparse.py", line 1536, in _process_short_opts
    option.process(opt, value, values, self)
  File "/usr/lib64/python3.6/optparse.py", line 785, in process
    self.action, self.dest, opt, value, values, parser)
  File "/usr/lib64/python3.6/optparse.py", line 807, in take_action
    parser.print_help()
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 1093, in print_help
    self.cmdln.onecmd(["help"])
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 1216, in _dispatch_cmd
    return handler(argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 610, in do_help
    doc = self._help_preprocess(doc, cmdname)
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 206, in _help_preprocess
    help_msg = cmdln.Cmdln._help_preprocess(self, help, cmdname)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 730, in _help_preprocess
    help = preprocessor(help, cmdname)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 845, in _help_preprocess_cmd_name
    handler = self._get_cmd_handler(cmdname)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 983, in _get_cmd_handler
    handler = getattr(self, 'do_' + cmdname)
TypeError: must be str, not NoneType

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #604, opened 2019-07-19
prevent plugins to break help

lethliel:
prevent call of preprocessor if ${cmd_name} marker is accidentaly set,
but cmdname is None.

fixes https://github.com/openSUSE/osc/issues/600

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -206,6 +206,20 @@ class Osc(cmdln.Cmdln):
         help_msg = cmdln.Cmdln._help_preprocess(self, help, cmdname)
         return help_msg % conf.config
 
+    def _help_preprocess_cmd_name(self, help, cmdname=None):
+        if cmdname is None:
+            return help
+        return cmdln.Cmdln._help_preprocess_cmd_name(self, help, cmdname)
+
+    def _help_preprocess_cmd_option_list(self, help, cmdname=None):
+        if cmdname is None:
+            return help
+        return cmdln.Cmdln._help_preprocess_cmd_option_list(self, help, cmdname)
+
+    def _help_preprocess_cmd_usage(self, help, cmdname=None):
+        if cmdname is None:
+            return help
+        return cmdln.Cmdln._help_preprocess_cmd_usage(self, help, cmdname)
 
     def do_init(self, subcmd, opts, project, package=None):
         """${cmd_name}: Initialize a directory as working copy


marcus-h:
On 2019-07-19 04:43:30 -0700, Marco Strigl wrote:
> diff --git a/osc/cmdln.py b/osc/cmdln.py
> index 918a1da1..7c930e5a 100644
> --- a/osc/cmdln.py
> +++ b/osc/cmdln.py
> @@ -727,7 +727,10 @@ def _help_preprocess(self, help, cmdname):
>
>          for marker, preprocessor in preprocessors.items():
>              if marker in help:
> -                help = preprocessor(help, cmdname)
> +                if marker == '${cmd_name}' and cmdname is None:
> +                    next

Hmm you probably want to use "continue" here.

> +                else:
> +                    help = preprocessor(help, cmdname)
>          return help
>

Doing this check for the cmd_name marker is probably not enough
(it could also break on cmd_option_list and cmd_usage). In general,
the whole code is quite fragile because it depends on the
"non-deterministic" order of preprocessor.items()...

I see two alternatives:
- enforce a specific order (this could also lead to problems...)
- override _get_cmd_handler to handle the "None" case more
  gracefully (it should simply return None)

I think we should go for the latter... (I'm a bit hesitant to
do the first approach because it would require changes in the
cmdln.py... (if we want a reproducible result we should probably
do both, though))

lethliel:
On 7/30/19 4:31 PM, Marcus Hüwe  wrote:
> On 2019-07-19 04:43:30 -0700, Marco Strigl wrote:
>> diff --git a/osc/cmdln.py b/osc/cmdln.py
>> index 918a1da1..7c930e5a 100644
>> --- a/osc/cmdln.py
>> +++ b/osc/cmdln.py
>> @@ -727,7 +727,10 @@ def _help_preprocess(self, help, cmdname):
>>
>> for marker, preprocessor in preprocessors.items():
>> if marker in help:
>> - help = preprocessor(help, cmdname)
>> + if marker == '${cmd_name}' and cmdname is None:
>> + next
> 
> Hmm you probably want to use "continue" here.
> 
>> + else:
>> + help = preprocessor(help, cmdname)
>> return help
>>
> 
> Doing this check for the cmd_name marker is probably not enough
> (it could also break on cmd_option_list and cmd_usage). In general,
> the whole code is quite fragile because it depends on the
> "non-deterministic" order of preprocessor.items()...
> 
> I see two alternatives:
> - enforce a specific order (this could also lead to problems...)
> - override _get_cmd_handler to handle the "None" case more
> gracefully (it should simply return None)
> 
> I think we should go for the latter... (I'm a bit hesitant to
> do the first approach because it would require changes in the
> cmdln.py... (if we want a reproducible result we should probably
> do both, though))
> 

_get_cmd_handler now returns None if cmdname is None. But the calling 
functions _help_preprocess_cmd_* needed to be changed too to handle 
an None-handler correctly (weird sentence)

marcus-h:
On 2019-07-31 04:40:22 -0700, Marco Strigl wrote:
> On 7/30/19 4:31 PM, Marcus Hüwe  wrote:
> > On 2019-07-19 04:43:30 -0700, Marco Strigl wrote:
> >> diff --git a/osc/cmdln.py b/osc/cmdln.py
> >> index 918a1da1..7c930e5a 100644
> >> --- a/osc/cmdln.py
> >> +++ b/osc/cmdln.py
> >> @@ -727,7 +727,10 @@ def _help_preprocess(self, help, cmdname):
> >>
> >> for marker, preprocessor in preprocessors.items():
> >> if marker in help:
> >> - help = preprocessor(help, cmdname)
> >> + if marker == '${cmd_name}' and cmdname is None:
> >> + next
> >
> > Hmm you probably want to use "continue" here.
> >
> >> + else:
> >> + help = preprocessor(help, cmdname)
> >> return help
> >>
> >
> > Doing this check for the cmd_name marker is probably not enough
> > (it could also break on cmd_option_list and cmd_usage). In general,
> > the whole code is quite fragile because it depends on the
> > "non-deterministic" order of preprocessor.items()...
> >
> > I see two alternatives:
> > - enforce a specific order (this could also lead to problems...)
> > - override _get_cmd_handler to handle the "None" case more
> > gracefully (it should simply return None)
> >
> > I think we should go for the latter... (I'm a bit hesitant to
> > do the first approach because it would require changes in the
> > cmdln.py... (if we want a reproducible result we should probably
> > do both, though))
> >
>
> _get_cmd_handler now returns None if cmdname is None. But the calling
> functions _help_preprocess_cmd_* needed to be changed too to handle
> an None-handler correctly (weird sentence)
>
Hmm I thought a more meaningful exception would be ok in this case, but
the exception is probably not too helpful from a user's POV. If I
understand you correctly, you would prefer to continue - so let's do
that:) But we should fix this differently:

We should probably override the three methods (in the Osc class) like
this:

def _help_preprocess_cmd_name(self, help, cmdname=None):
    if cmdname is None:
        return help
    return cmdln.Cmdln._help_preprocess_cmd_name(self, help, cmdname)

(so that we do not have to modify the cmdln.py)

lethliel:
On 8/2/19 1:29 PM, Marcus Hüwe  wrote:
> On 2019-07-31 04:40:22 -0700, Marco Strigl wrote:
>> On 7/30/19 4:31 PM, Marcus Hüwe wrote:
>> > On 2019-07-19 04:43:30 -0700, Marco Strigl wrote:

[SNIP]

>> >
>>
>> _get_cmd_handler now returns None if cmdname is None. But the calling
>> functions _help_preprocess_cmd_* needed to be changed too to handle
>> an None-handler correctly (weird sentence)
>>
> Hmm I thought a more meaningful exception would be ok in this case, but
> the exception is probably not too helpful from a user's POV. If I
> understand you correctly, you would prefer to continue - so let's do
> that:) But we should fix this differently:
> 
> We should probably override the three methods (in the Osc class) like
> this:
> 
> def _help_preprocess_cmd_name(self, help, cmdname=None):
> if cmdname is None:
> return help
> return cmdln.Cmdln._help_preprocess_cmd_name(self, help, cmdname)
> 
> (so that we do not have to modify the cmdln.py)
> 

Ahhh. Now I know what you mean. I added the _help_preprocess_cmd* function
to the Osc class in commandline.py and reverted my changes in the particular
functions in cmdln.py

marcus-h:
On 2019-08-02 05:18:32 -0700, Marco Strigl wrote:
> On 8/2/19 1:29 PM, Marcus Hüwe  wrote:
> > On 2019-07-31 04:40:22 -0700, Marco Strigl wrote:
> >> On 7/30/19 4:31 PM, Marcus Hüwe wrote:
> >> > On 2019-07-19 04:43:30 -0700, Marco Strigl wrote:
>
> [SNIP]
>
> >> >
> >>
> >> _get_cmd_handler now returns None if cmdname is None. But the calling
> >> functions _help_preprocess_cmd_* needed to be changed too to handle
> >> an None-handler correctly (weird sentence)
> >>
> > Hmm I thought a more meaningful exception would be ok in this case, but
> > the exception is probably not too helpful from a user's POV. If I
> > understand you correctly, you would prefer to continue - so let's do
> > that:) But we should fix this differently:
> >
> > We should probably override the three methods (in the Osc class) like
> > this:
> >
> > def _help_preprocess_cmd_name(self, help, cmdname=None):
> > if cmdname is None:
> > return help
> > return cmdln.Cmdln._help_preprocess_cmd_name(self, help, cmdname)
> >
> > (so that we do not have to modify the cmdln.py)
> >
>
> Ahhh. Now I know what you mean. I added the _help_preprocess_cmd* function
> to the Osc class in commandline.py and reverted my changes in the particular
> functions in cmdln.py
>
Can you also revert the change in cmdln.py? (It is not needed anymore).

> index 8ea37ba3..177250d5 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py

> +    def _help_preprocess_option_list(self, help, cmdname=None):
> +        if cmdname is None:
> +            return help
> +        return cmdln.Cmdln._help_preprocess_option_list(self, help, cmdname)

s/_help_preprocess_option_list/_help_preprocess_cmd_option_list/

Issue #607, opened 2019-07-23
Package filtering on osc rq list fails


lethliel:
Filtering requests with --target-package-filter fails on python2 and python3 when tgt_package is None. (for example add_role requests)

`osc rq list openSUSE:Tools --target-package-filter=osc`

results in: 

```
Traceback (most recent call last):
  File "/home/marco/git_r/osc/osc-wrapper.py", line 41, in <module>
    r = babysitter.run(osccli)
  File "/home/marco/git_r/osc/osc/babysitter.py", line 61, in run
    return prg.main(argv)
  File "/home/marco/git_r/osc/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/home/marco/git_r/osc/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/home/marco/git_r/osc/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/home/marco/git_r/osc/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/home/marco/git_r/osc/osc/commandline.py", line 2430, in do_request
    if not filter_pattern.match(action.tgt_package):

```

PR #608, opened 2019-07-23
Fix package filtering for osc rq

lethliel:
Check if action.tgt_package is None.

Fixes https://github.com/openSUSE/osc/issues/607

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -2426,7 +2426,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                 for action in result.actions:
                     if action.type == 'group' or not opts.target_package_filter:
                         continue
-                    if not filter_pattern.match(action.tgt_package):
+                    if action.tgt_package is not None and not filter_pattern.match(action.tgt_package):
                         filtered = True
                         break
                 if not filtered:


marcus-h:
On 2019-07-23 13:53:35 +0000, Marco Strigl wrote:
> index 94c2c7d2..6fd6b731 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -2426,7 +2426,7 @@ def do_request(self, subcmd, opts, *args):
>                  for action in result.actions:
>                      if action.type == 'group' or not opts.target_package_filter:
>                          continue
> -                    if not filter_pattern.match(action.tgt_package):
> +                    if action.tgt_package is not None and not filter_pattern.match(action.tgt_package):

Hmm we could do this but does this comply with the description of the
--target-package-filter option? An alternative would be:

+                    if action.tgt_package is None or not filter_pattern.match(action.tgt_package):

(I'm not sure if the alternative complies with the description, too;) )

>                          filtered = True
>                          break
>                  if not filtered:

lethliel:
On 7/30/19 4:41 PM, Marcus Hüwe  wrote:
> On 2019-07-23 13:53:35 +0000, Marco Strigl wrote:
>> diff --git a/osc/commandline.py b/osc/commandline.py
>> index 94c2c7d2..6fd6b731 100644
>> --- a/osc/commandline.py
>> +++ b/osc/commandline.py
>> @@ -2426,7 +2426,7 @@ def do_request(self, subcmd, opts, *args):
>> for action in result.actions:
>> if action.type == 'group' or not opts.target_package_filter:
>> continue
>> - if not filter_pattern.match(action.tgt_package):
>> + if action.tgt_package is not None and not filter_pattern.match(action.tgt_package):
> 
> Hmm we could do this but does this comply with the description of the
> --target-package-filter option? An alternative would be:
> 
> + if action.tgt_package is None or not filter_pattern.match(action.tgt_package):
> 
> (I'm not sure if the alternative complies with the description, too;) )

The problem is that if there is no action.tgt_package (for example for add_role requests)
the option naming makes no sense. 

The description reads like "Just show requests for the package matchin the filter". It is 
the option name that is misleading. But renaming it would probably break some scripts. 

So I think the fix is fine, but the option naming is kind of fishy.

> 
>> filtered = True
>> break
>> if not filtered:
>

marcus-h:
On 2019-07-31 02:08:49 -0700, Marco Strigl wrote:
> On 7/30/19 4:41 PM, Marcus Hüwe  wrote:
> > On 2019-07-23 13:53:35 +0000, Marco Strigl wrote:
> >> diff --git a/osc/commandline.py b/osc/commandline.py
> >> index 94c2c7d2..6fd6b731 100644
> >> --- a/osc/commandline.py
> >> +++ b/osc/commandline.py
> >> @@ -2426,7 +2426,7 @@ def do_request(self, subcmd, opts, *args):
> >> for action in result.actions:
> >> if action.type == 'group' or not opts.target_package_filter:
> >> continue
> >> - if not filter_pattern.match(action.tgt_package):
> >> + if action.tgt_package is not None and not filter_pattern.match(action.tgt_package):
> >
> > Hmm we could do this but does this comply with the description of the
> > --target-package-filter option? An alternative would be:
> >
> > + if action.tgt_package is None or not filter_pattern.match(action.tgt_package):
> >
> > (I'm not sure if the alternative complies with the description, too;) )
>
> The problem is that if there is no action.tgt_package (for example for add_role requests)
> the option naming makes no sense.
>
> The description reads like "Just show requests for the package matchin the filter". It is
> the option name that is misleading. But renaming it would probably break some scripts.
>
Renaming is no option:)
Merged. Thanks!

Issue #596, opened 2019-07-11
Certificate changed unexpectedly. This should not happen


wfrisch:
Checking out certain packages fails with an SSL error.
Most operations, however, work fine, including checking out other packages.

osc --version: 0.165.3.git (9a1b2d980a26aad35e1bb26433b7d24c0f2cb6fc)
```
osc -A ibs co SUSE:SLE-12:Update ImageMagick
A    SUSE:SLE-12:Update
A    SUSE:SLE-12:Update/ImageMagick
A    SUSE:SLE-12:Update/ImageMagick/ImageMagick-6.6.8.9-doc.patch
[...]
A    SUSE:SLE-12:Update/ImageMagick/ImageMagick-CVE-2016-7532.patch
A    SUSE:SLE-12:Update/ImageMagick/ImageMagick-CVE-2016-7533.patch
A    SUSE:SLE-12:Update/ImageMagick/ImageMagick-CVE-2016-7534.patch
Certificate changed unexpectedly. This should not happen
SSL Error: certificate verify failed
```

```
osc -A ibs co SUSE:SLE-11-SP3:Update kvm
[...]
A    SUSE:SLE-11-SP3:Update/kvm/1123-ahci-fix-buffer-overrun-on-invalid-.patch
A    SUSE:SLE-11-SP3:Update/kvm/1124-vmstate-add-VMS_MUST_EXIST.patch
Certificate changed unexpectedly. This should not happen
SSL Error: certificate verify failed
```

osc-0.165.1 from openSUSE Tumbleweed stops at the same point but with a segmentation fault:
```
(gdb) file python3
(gdb) run /usr/bin/osc -A ibs co SUSE:SLE-12:Update ImageMagick

[...]

A    SUSE:SLE-12:Update/ImageMagick/ImageMagick-CVE-2016-7534.patch

Program received signal SIGSEGV, Segmentation fault.
i2c_ASN1_INTEGER (a=0x6f6c168256018230, pp=0x0) at crypto/asn1/a_int.c:203
203         return i2c_ibuf(a->data, a->length, a->type & V_ASN1_NEG, pp
```
The problem seems to be related to https://github.com/openSUSE/osc/issues/451

wfrisch:
My initial uninformed impression is that the issue occurs only during longer operations, for example when checking out packages with an unusually large number of patches.

marcus-h:
Just for the record (github seemed to ignore my mail again):
```
[...]

PR #609, opened 2019-07-23
fix oscssl.py not verifying TLS cert verification [CVE-2013-3685]

lethliel:
use own implementation of HTTPSConnection (myHTTPSConnection)
instead the one provided by M2Crypto (httpslib.HTTPConnection)

all credits go to @wfrisch

fixes https://github.com/openSUSE/osc/issues/596
reported in bug https://bugzilla.suse.com/show_bug.cgi?id=1142518

### DIFF
--- a/osc/oscssl.py
+++ b/osc/oscssl.py
@@ -199,13 +199,13 @@ class myHTTPSHandler(M2Crypto.m2urllib2.HTTPSHandler):
 
         if target_host != host:
             request_uri = urldefrag(full_url)[0]
-            h = httpslib.ProxyHTTPSConnection(host=host, ssl_context=self.ctx)
+            h = myProxyHTTPSConnection(host=host, ssl_context=self.ctx)
         else:
             try:     # up to python-3.2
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2019-07-23 08:45:42 -0700, Marco Strigl wrote:
> use own implementation of HTTPSConnection (myHTTPSConnection)
> instead the one provided by M2Crypto (httpslib.HTTPConnection)
>
Hmm this seems be caused by commit 6ddf771c9f753cf052d9eb2965000021f8ecd26c.
We should do the same for the proxy connection.
Maybe we should do a "post" review of that commit.

> all credits go to @wfrisch
>
Thanks!

> fixes https://github.com/openSUSE/osc/issues/596

Hmm I'm not convinced that this fixes the segfault (at least it is
not obvious to me atm (I haven't checked the code, though)).

lethliel:
Yes. You are right. We missed the proxy case. Will add that. 

We can of course do a Post review of the oscssl.py commit

Issue #605, opened 2019-07-22
Better error message on failing submitrequest for mutliple packages


jonsger:
I wanted to make a single SR for some 20 packages, but it failed and I didn't noticed why.

```
home:jbrielmaier:branches:X11:MATE:Factory> osc submitrequest X11:MATE:Factory -m "Update MATE to 1.23"
Submitting package  atril
[...]
Submitting package  python-caja
The request contains no actions. Submit requests without source changes may have skipped!
```

The actual problem was that I didn't check in the changes in the `mate-applets` package, so in this package there was no source change.

It would be nice to have a error message leading you to the actual problem like
```
The request contains no actions. The following packages didn't contain a source change:
* mate-applets
Submit request may have skipped!
```

~Jonathan

lethliel:
as discussed in IRC we will have a look at this together tomorrow

adrianschroeter:
this is an api side message in first place

adrianschroeter:
but the message should only come when in _no_ package is no change. It should proceed if in at least one package is a change. The other packages will just not be part of the request action list.

lethliel:
@adrianschroeter 

The api returns a 400 Bad Request error and no sr is created. 
I will open an issue for the API.

lethliel:
This is solved in https://github.com/openSUSE/osc/pull/618/commits/3da7b9b3d8d2f46ae0ce1d9f84bd8c1666335603

PR #611, opened 2019-07-24
Better error message when submit of project fails

lethliel:
fetch error on osc sr on project level when one or more
packages do not contain a change.

The user now gets a hint what went wrong and how to check it.

Co-Authored with: Jonathan Brielmaier @jonsger 

fixes https://github.com/openSUSE/osc/issues/605


lethliel:
@marcus-h: please review

lethliel:
not needed.

PR #615, opened 2019-07-26
return the obj if None type is passed to decode_it

lethliel:
If a obj of type None is passed to decode_it just return it and do not try to decode it as this will fail

### DIFF
--- a/osc/util/helper.py
+++ b/osc/util/helper.py
@@ -54,7 +54,7 @@ def decode_it(obj):
         based on the chardet module if possible
     """
 
-    if isinstance(obj, str):
+    if obj is None or isinstance(obj, str):
         return obj
     else:
         try:


marcus-h:
On 2019-07-26 02:00:42 -0700, Marco Strigl wrote:
> This should be handled by the caller, but just to be sure
> if an obj in None we just return it.

Hmm if I read the description it sounds to me as if we should
raise an exception if None is passed.

lethliel:
On 7/26/19 12:44 PM, Marcus Hüwe  wrote:
> On 2019-07-26 02:00:42 -0700, Marco Strigl wrote:
>> This should be handled by the caller, but just to be sure
>> if an obj in None we just return it.
> 
> Hmm if I read the description it sounds to me as if we should
> raise an exception if None is passed.
> 

I stumbled over this, when I tried to pass decode_it(self.release) to a function 
and it is ok, when the passed argument is None. 

So I would not raise an exception.

marcus-h:
On 2019-07-26 04:17:39 -0700, Marco Strigl wrote:
> On 7/26/19 12:44 PM, Marcus Hüwe  wrote:
> > On 2019-07-26 02:00:42 -0700, Marco Strigl wrote:
> >> This should be handled by the caller, but just to be sure
> >> if an obj in None we just return it.
> >
> > Hmm if I read the description it sounds to me as if we should
> > raise an exception if None is passed.
> >
>
> I stumbled over this, when I tried to pass decode_it(self.release) to a function
> and it is ok, when the passed argument is None.
>
> So I would not raise an exception.
>
I'm fine with the patch in general, but the commit description
confuses me a bit. For me [1], it reads like "do not pass None to
this function" (that's the "[t]his should be handled by the
caller" part).

I know, this is a really really really minor nitpick but let's
try to get this straight.


[1] Maybe it is the heat that drives me crazy...

marcus-h:
Merged. Thanks!

PR #616, opened 2019-07-28
Use correct appname for trusted-certs store

andreas-schwab:


marcus-h:
Arghs... *blush*
Merged. Thanks a lot!

PR #618, opened 2019-08-02
simplify osc sr code in submitting prj

lethliel:
Most of the stuff is now done in the API.
For just submitting all in one request the following is sufficient:

`<action type="submit"> <source project="%s" /> %s </action>`

We don't need package name or source package and targets anymore.
This is all handled by the API now.

For the `--seperate-requests` case the check for the entries in the linkinfo
is not working on most cases. So we handle this in do_submitrequest call
and skip the package if `X-Opensuse-Errorcode` is `missing_action` which means
basically nothing to do for this package

solves https://github.com/openSUSE/osc/issues/605

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -1236,73 +1236,64 @@ class Osc(cmdln.Cmdln):
             project = store_read_project(os.curdir)
 
             sr_ids = []
-            # for single request
-            actionxml = ""
-            options_block = "<options>"
-            if src_update:
-                options_block += """<sourceupdate>%s</sourceupdate>""" % (src_update)
-            if opts.update_link:
-                options_block  + """<updatelink>true</updatelink></options> """
-            options_block += "</options>"
 
-            # loop via all packages for checking their state
-            for p in meta_get_packagelist(apiurl, project):
-                # get _link info from server, that knows about the local state ...
-                u = makeurl(apiurl, ['source', project, p])
-                f = http_GET(u)
-                root = ET.parse(f).getroot()
-                target_project = None
-                if len(args) == 1:
-                    target_project = args[0]
-                linkinfo = root.find('linkinfo')
-                if linkinfo == None:
-                    if len(args) < 1:
-                        print("Package ", p, " is not a source link and no target specified.")
-                        sys.exit("This is currently not supported.")
-                else:
-                    if linkinfo.get('error'):
-                        print("Package ", p, " is a broken source link.")
-                        sys.exit("Please fix this first")
-                    t = linkinfo.get('project')
-                    if t:
-                        if target_project == None:
-                            target_project = t
-                        if len(root.findall('entry')) > 1: # This is not really correct, but should work mostly
-                                                           # Real fix is to ask the api if sources are modificated
-                                                           # but there is no such call yet.
-                            print("Submitting package ", p)
-                        else:
-                            print("  Skipping not modified package ", p)
-                            continue
+            target_project = None
+            if len(args) == 1:
+                target_project = args[0]
+            if opts.separate_requests or opts.seperate_requests:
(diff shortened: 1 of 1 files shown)


lethliel:
@marcus-h: please review :)

marcus-h:
On 2019-08-02 04:57:36 -0700, Marco Strigl wrote:
> index 032f62fa..04c675ca 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -4268,6 +4268,8 @@ def create_submit_request(apiurl,
>              tproject = project.get('name')
>              r = create_maintenance_request(apiurl, src_project, [src_package], tproject, dst_project, src_update, message)
>              r = r.reqid
> +        elif e.hdrs.get('X-Opensuse-Errorcode') == "missing_action":
> +            r = e.hdrs.get('X-Opensuse-Errorcode')

Hmm this would change the api and may break 3rd party code. What about
moving this check into do_submitrequest? Alternatively, we could simply
error out (as we do now).

lethliel:
On 8/15/19 5:54 PM, Marcus Hüwe  wrote:
> On 2019-08-02 04:57:36 -0700, Marco Strigl wrote:
>> index 032f62fa..04c675ca 100644
>> --- a/osc/core.py
>> +++ b/osc/core.py
>> @@ -4268,6 +4268,8 @@ def create_submit_request(apiurl,
>> tproject = project.get('name')
>> r = create_maintenance_request(apiurl, src_project, [src_package], tproject, dst_project, src_update, message)
>> r = r.reqid
>> + elif e.hdrs.get('X-Opensuse-Errorcode') == "missing_action":
>> + r = e.hdrs.get('X-Opensuse-Errorcode')
> 
> Hmm this would change the api and may break 3rd party code. What about
> moving this check into do_submitrequest? Alternatively, we could simply
> error out (as we do now).
> 

Ok. I don't see any scenarios where people would rely on raising on X-Opensuse-Errorcode is missing_action,
but you are probably right that the caller should handle this. 

Moved the exception handling for missing_action to do_submitrequest and edited the PR and commit description.

marcus-h:
On 2019-08-16 05:56:15 -0700, Marco Strigl wrote:
> On 8/15/19 5:54 PM, Marcus Hüwe  wrote:
> > On 2019-08-02 04:57:36 -0700, Marco Strigl wrote:
> >> diff --git a/osc/core.py b/osc/core.py
> >> index 032f62fa..04c675ca 100644
> >> --- a/osc/core.py
> >> +++ b/osc/core.py
> >> @@ -4268,6 +4268,8 @@ def create_submit_request(apiurl,
> >> tproject = project.get('name')
> >> r = create_maintenance_request(apiurl, src_project, [src_package], tproject, dst_project, src_update, message)
> >> r = r.reqid
> >> + elif e.hdrs.get('X-Opensuse-Errorcode') == "missing_action":
> >> + r = e.hdrs.get('X-Opensuse-Errorcode')
> >
> > Hmm this would change the api and may break 3rd party code. What about
> > moving this check into do_submitrequest? Alternatively, we could simply
> > error out (as we do now).
> >
>
> Ok. I don't see any scenarios where people would rely on raising on X-Opensuse-Errorcode is missing_action,
> but you are probably right that the caller should handle this.
>
Let's consider the following made up code-snippet:

...
rid = create_submit_request(apiurl, src_prj, src_pkg, tgt_prj)
if rid is None:
    # would indicate a broken xml, but well...
    sys.exit('failure')
# ok SR was created, insert rid in some db whatever...
...

Now, in case of a missing_action, the code above would continue and
could fail in some very strange way...

Back to your patch: "cd prj; osc sr <target_prj>" will ignore the specified
target project. Otherwise, the patch seems to be OK from a refactoring POV!
(Some parts of the old code are pretty buggy/broken - we should probably
tackle them in another patch.)

Issue #617, opened 2019-08-01
 regression in deleterequest --accept-in-hours


olafhering:
see openSUSE/open-build-service/issues/8010

Looks like osc 0.165.1 sends the time via `accept_at`, instead of letting the server figuring out what now+N means for the server.

hennevogel:
I guess you're on an UTC machine? :-)

lethliel:
@olafhering 

I think too, that something with your local time is wrong. 
osc deleterequest --accept-in-hours=1 one works fine for me. 

osc has always calculated the accept time. Not the server (at least the code is in there since 2013)

olafhering:
What could go wrong with a plain osc call? According to the --whatever-debug output that timestamp was in UTC (I think) and it was in the past.

lethliel:
Can you reproduce the issue and provide the -H output?

olafhering:
````
 home:olh:bax/pbzip2 $ osc -H dr --accept-in-hours=1 -m 617


-- POST https://api.suse.de/request?cmd=create
send: b'POST /request?cmd=create HTTP/1.1\r\nAccept-Encoding: identity\r\nContent-type: application/octet-stream\r\nContent-length: 197\r\nHost: api.suse.de\r\nUser-agent: osc/0.165.1\r\nConnection: close\r\n\r\n'
send: b'<request>\n  <action type="delete">\n    <target package="pbzip2" project="home:olh:bax" />\n  </action>\n  <description>617</description>\n  <accept_at>2019-08-05T14:05:14.189729</accept_at>\n</request>'
reply: 'HTTP/1.1 400 Bad Request\r\n'
header: Date header: Server header: Strict-Transport-Security header: Cache-Control header: Vary header: Referrer-Policy header: X-Permitted-Cross-Domain-Policies header: X-XSS-Protection header: X-Opensuse-Runtimes header: X-Request-Id header: X-Opensuse-APIVersion header: X-Download-Options header: X-Runtime header: X-Frame-Options header: X-Content-Type-Options header: X-Opensuse-Errorcode header: X-Powered-By header: Status header: Content-Type header: Connection header: Transfer-Encoding Server returned an error: HTTP Error 400: Bad Request
Auto accept time is in the past
````

lethliel:
Mhh. Ok. I see the problem. Need to look for a fix. Thanks for reporting. :-)

PR #621, opened 2019-08-14
make now() UTC aware for accept-in-hours

lethliel:
just send the UTC time and the API will do the rest

could not use datetime.timezone.utc, because it is not python2 compatible. Therefore the
module pytz

fixes https://github.com/openSUSE/osc/issues/617

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -2860,7 +2860,7 @@ class Request:
 
         now = datetime.datetime.utcnow()
         now = now + datetime.timedelta(hours=hours)
-        self.accept_at = now.isoformat()
+        self.accept_at = now.isoformat() + '+00:00'
 
     @staticmethod
     def format_review(review, show_srcupdate=False):


adrianschroeter:
On Mittwoch, 14. August 2019, 13:35:19 CEST Marco Strigl wrote:
> The API uses the offset to calculate the correct time when
> the request should be auto accepted.
> 
> fixes https://github.com/openSUSE/osc/issues/617
> 
>   https://github.com/openSUSE/osc/pull/621
> 
> -- Commit Summary --
> 
>   * append offset to UTC to the accept_at time
> 
> -- File Changes --
> 
>     M osc/core.py (3)
> 
> -- Patch Links --
> 
> https://github.com/openSUSE/osc/pull/621.patch
> https://github.com/openSUSE/osc/pull/621.diff

sorry, can not login into github atm, but

Your code is always using UTC time but sending the local timezone of your 
workstation, right?

so, there will be mismatches, if you send utc time, you can hardcode "UTC" as 
timezone as well


-- 

Adrian Schroeter
SUSE Linux Products GmbH,  Maxfeldstr. 5, 90409 Nuernberg, Germany
email: adrian@suse.de

marcus-h:
On 2019-08-14 04:35:19 -0700, Marco Strigl wrote:
> The API uses the offset to calculate the correct time when
> the request should be auto accepted.
>
> fixes https://github.com/openSUSE/osc/issues/617

Hmm I have to admit that I'm not much into this time* stuff, but
why is this needed? Why can't we simply work with such a "naive"
time object? (There could be some corner cases which might fail,
but the bug you referenced does not seem to be such a corner
case (at least from my naive POV))

marcus-h:
On 2019-09-12 01:51:33 -0700, Marco Strigl wrote:
> > On 2019-08-14 04:35:19 -0700, Marco Strigl wrote: The API uses the offset to calculate the correct time when the request should be auto accepted. fixes #617
> > Hmm I have to admit that I'm not much into this time* stuff, but why is this needed? Why can't we simply work with such a "naive" time object? (There could be some corner cases which might fail, but the bug you referenced does not seem to be such a corner case (at least from my naive POV))
>
> It is. The API does not calculate the time right If we do not give it a UTC aware time object.
>
Hmm I haven't checked the api code, but the only difference I see (from a
quick glance) is that the new code appends the "utc offset" to the isoformat.
That is, '2019-09-16T16:56:03.129347' -> '2019-09-16T16:56:03.129347+00:00'.
Can't we simply concat '+00:00' to the isoformat() and keep the old code?
(This way we would avoid the pytz dependency.)

lethliel:
On 9/16/19 4:00 PM, Marcus Hüwe  wrote:
> On 2019-09-12 01:51:33 -0700, Marco Strigl wrote:
>> > On 2019-08-14 04:35:19 -0700, Marco Strigl wrote: The API uses the
> offset to calculate the correct time when the request should be auto
> accepted. fixes #617
>> > Hmm I have to admit that I'm not much into this time* stuff, but why
> is this needed? Why can't we simply work with such a "naive" time
> object? (There could be some corner cases which might fail, but the bug
> you referenced does not seem to be such a corner case (at least from my
> naive POV))
>>
>> It is. The API does not calculate the time right If we do not give it
> a UTC aware time object.
>>
> Hmm I haven't checked the api code, but the only difference I see (from a
> quick glance) is that the new code appends the "utc offset" to the
> isoformat.
> That is, '2019-09-16T16:56:03.129347' -> '2019-09-16T16:56:03.129347+00:00'.
> Can't we simply concat '+00:00' to the isoformat() and keep the old code?
> (This way we would avoid the pytz dependency.)

You are right. This is way more easier and no additional module is needed.

Changed.

lethliel:
as discussed with @marcus-h this can be merged now

PR #622, opened 2019-08-22
osc/core: parseRevisionOption(): cleanup and make logic consistent.

jberry-suse:
Handle multiple revisions the same as a single revision in terms of what
is allowed (digit, or 32 character string [md5]). Additionally, support
either blank or "latest" to mean latest revision (None).

This allows for new revision to be specified without old revision like the
following example:

osc rdiff ... -r :17

This is needed to complete openSUSE/openSUSE-release-tools#2144 which will alternatively have to wrap this logic to lookup current old revision before calling into `osc`. This is preferred since also usable from CLI directly.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -6841,7 +6841,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             project = args[0]
             package = args[1]
 
-        rev, rev_upper = parseRevisionOption(opts.revision)
+        rev, rev_upper = parseRevisionOption(opts.revision, allow_md5=False)
         if rev and not checkRevision(project, package, rev, apiurl, opts.meta):
             print('Revision \'%s\' does not exist' % rev, file=sys.stderr)
             sys.exit(1)
--- a/osc/core.py
+++ b/osc/core.py
@@ -6603,32 +6603,22 @@ def cmdbuild(apiurl, cmd, project, package=None, arch=None, repo=None, code=None
     return root.get('code')
 
 
-def parseRevisionOption(string):
+def parseRevisionOption(string, allow_md5=True):
     """
     returns a tuple which contains the revisions
     """
 
+    revisions = [None, None]
     if string:
-        if ':' in string:
-            splitted_rev = string.split(':')
-            try:
-                for i in splitted_rev:
-                    int(i)
-                return splitted_rev
-            except ValueError:
+        parts = string.split(':')
+        for i, revision in enumerate(parts[0:2], 0):
+            if revision.isdigit() or (allow_md5 and revision.isalnum() and len(revision) == 32):
+                revisions[i] = revision
+            elif revision != '' and revision != 'latest':
                 print('your revision \'%s\' will be ignored' % string, file=sys.stderr)
                 return None, None
-        else:
-            if string.isdigit():
-                return string, None
-            elif string.isalnum() and len(string) == 32:
-                # could be an md5sum
-                return string, None
-            else:
-                print('your revision \'%s\' will be ignored' % string, file=sys.stderr)
-                return None, None
-    else:
-        return None, None
+
+    return tuple(revisions)
 
 def checkRevision(prj, pac, revision, apiurl=None, meta=False):
     """


marcus-h:
On 2019-08-22 13:55:53 -0700, Jimmy Berry wrote:
> Handle multiple revisions the same as a single revision in terms of what
> is allowed (digit, or 32 character string [md5]). Additionally, support
> either blank or "latest" to mean latest revision (None).
>
Hmm we could do that. The patch looks good to me except that it could yield
to unexpected behavior (from a user's POV) in the do_log caller:
old code: osc log <prj> <pkg> -r 5:<some md5sum> complains about the md5sum
          (and continues as if no -r option was specified)
new code: osc log <prj> <pkg> -r 5:<some md5sum> simply ignores the -r option
          (without complaining)

Long story short: we could do this change, but this could potentially break
other plugins, which rely on the old behavior.

marcus-h:
On 2019-08-26 13:31:17 -0700, Jimmy Berry wrote:
> > Long story short: we could do this change, but this could potentially break other plugins, which rely on the old behavior.
>
> In summary, something that was not supported before is still not supported. If the error message is desired to be restored there are a few options:
>
> - do not allow md5sum with multiple revisions (like before), but still allow only second revision
> - add optional argument to disallow md5sum revisions
> - explicitly check for md5sum revisions in log
>
> My preference would be the middle option. Either way this would seem to be one hell of an edge-case if someone is depending on an error message when running log when still getting the same output.
>
I merely used this example to illustrate that we change/break the
existing API. That is, if you, as plugin developer, rely on the
old behavior (-r x:y implies that x and y can be represented as ints),
then your plugin may break.

I'm fine with all your options. In lieu of your middle option, we could
also add an argument to enable md5sum revisions, which is disabled by
default in order to keep the current API.

jberry-suse:
Let me know if any other changes desired.

dmach:
The new code behaves slightly differently than the old one.

I've created `tests/test_core.py` - could you add it to your PR and eventually modify it according to the expected behavior?
I'll merge the PR afterwards.
```
from osc.core import parseRevisionOption

import unittest


class TestParseRevisionOption(unittest.TestCase):
    def test_empty(self):
        expected = None, None
        actual = parseRevisionOption("")
        self.assertEqual(expected, actual)

    def test_invalid_colon(self):
        expected = None, None
        actual = parseRevisionOption(":")
        # your revision ':' will be ignored
        self.assertEqual(expected, actual)

    def test_invalid_multiple_colons(self):
        expected = None, None
        actual = parseRevisionOption(":::::")
        # your revision ':::::' will be ignored
        self.assertEqual(expected, actual)

    def test_one_number(self):
        expected = ("1", None)
        actual = tuple(parseRevisionOption("1"))
        self.assertEqual(expected, actual)

    def test_two_numbers(self):
        expected = ("1", "2")
        actual = tuple(parseRevisionOption("1:2"))
        self.assertEqual(expected, actual)

    def test_many_numbers(self):
        expected = ("1", "2", "3", "4", "5")
        actual = tuple(parseRevisionOption("1:2:3:4:5"))
        self.assertEqual(expected, actual)

    def test_one_hash(self):
        expected = "c4ca4238a0b923820dcc509a6f75849b", None
        actual = parseRevisionOption("c4ca4238a0b923820dcc509a6f75849b")
        self.assertEqual(expected, actual)

    def test_invalid_two_hashes(self):
        expected = None, None
        actual = parseRevisionOption("d41d8cd98f00b204e9800998ecf8427e:c4ca4238a0b923820dcc509a6f75849b")
        self.assertEqual(expected, actual)


if __name__ == "__main__":
    import unittest
    unittest.main()
```

dmach:
The PR looks good, I'll merge it and submit another one with several more tweaks.

marcus-h:
On 2022-04-04 01:52:40 -0700, Daniel Mach wrote:
> The PR looks good, I'll merge it and submit another one with several more tweaks.
>
The problem is that this now breaks "osc log <srcmd5>":/

marcus-h:
On 2022-04-07 22:21:32 +0000, Marcus Hüwe wrote:
> On 2022-04-04 01:52:40 -0700, Daniel Mach wrote:
> > The PR looks good, I'll merge it and submit another one with several more tweaks.
> >
> The problem is that this now breaks "osc log <srcmd5>":/
>
osc log -r <srcmd5>, of course:)

Issue #623, opened 2019-08-26
getbinaries: wrong reporting of available repositories


jengelh:
osc-0.165.4
Observed:
```
» osc co devel:languages:python/python-hexdump
» cd devel/languages/python/python-hexdump
» osc getbinaries openSUSE:Factory/python-hexdump openSUSE_Tumbleweed x86_64
Valid arguments for this package are:

openSUSE_Tumbleweed        i586
openSUSE_Tumbleweed        x86_64
openSUSE_Leap_15.1         x86_64
openSUSE_Factory_zSystems  s390x
openSUSE_Factory_PowerPC   ppc64
openSUSE_Factory_PowerPC   ppc64le
openSUSE_Factory_ARM       armv7l
openSUSE_Factory_ARM       aarch64
SLE_15_SP1                 x86_64
SLE_12_SP4                 x86_64
Invalid repository 'openSUSE_Tumbleweed'
```
Expected:
```
standard   x86_64
...
```

lethliel:
confirmed. Will fix this bug in osc

PR #626, opened 2019-08-30
Add project parameter to print_repos

lethliel:
This is needed if the project differs from the current
working copy.

Fixes https://github.com/openSUSE/osc/issues/623


lethliel:
reviewed ok

PR #628, opened 2019-09-11
Initialize diff as bytes

mgorse:
Otherwise, python 3 will throw a traceback if the fallback code is
triggered, since it appends bytes.


marcus-h:
Good catch! Merged. Thanks a lot for the PR!

PR #631, opened 2019-09-12
add support for new staging api. (command osc ps)

lethliel:
Create, Add, Delete staging projects
Add requests to staging

It can be issued within a project directory or with the
project as the first argument.

If you operate on a staging project or with requests the
staging project has to be given with `-s` and the request has
to be given with `-r`

You can add multiple requests / stagings comma serparated.
For example:
 ```
 -s myprj:Staging:A, myprj:Staging:B
  -r 1234, 12345
```

osc ps

For example:
    `linux:~/myproject/> osc ps create`
    or
    `osc ps create myproject`

STILL IN BETA

This fixes https://github.com/openSUSE/osc/issues/583

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -9249,6 +9249,238 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             result = delete_comment(apiurl, args[1])
             print(result)
 
+    @cmdln.option('-g', '--group',
+            help='group', metavar='GROUP')
+    @cmdln.option('-s', '--staging', action='append', default=[],
+            help='staging project', metavar='STAGING')
+    @cmdln.option('-r', '--requests', action='append', default=[],
+            help='request(s)', metavar='REQUESTS')
+    @cmdln.option('-v', '--verbose', action='store_true',
+                        help='print extra information')
+    @cmdln.alias('ps')
+    def do_projectstaging(self, subcmd, opts, cmd=None, *args):
+        """${cmd_name}: Handle staging
+
+        Create, Add, Delete staging projects
+        Add requests to staging
+
+        This can be issued within a project directory or with the
+        project as the first argument.
+
+        If you operate on a staging project or with requests the
+        staging project has to be given with -s and the request has
+        to be given with -r
+
+        You can add multiple requests / stagings:
+        For example:
+        "-s myprj:Staging:A -s myprj:Staging:B"
+        "-r 1234 -r 12345"
+
+        osc ps
+
+        For example:
+        linux:~/myproject/> osc ps create
+        or
+        osc ps create myproject
+
+        subcommands:
+            create    Create a staging workflow for the project
+            add       Add a staging project to a staging workflow
+                      using -s <staging project>
+            list      List all staging project associated to this
+                      workflow
+            show      Show information to one specific staging project
+                      using -s <staging project>
+            delete
+                      Delete complete staging workflow from project
+            accept
+                      Accept the staging given with -s. All requests in
+                      the staging project will be accepted
+
+
+        osc ps request
+
+        subcommands:
+            add       Add requests given with -r to a staging given with -s
+            delete    Delete requests given with -r from a staging given with -s
+            list      List all requests in a staging given with -s
+
+
+
+        Example Workflow:
+        osc ps create myprj -g group_for_review
+        osc ps add myprj -s myprj:Staging:A
+        osc ps request add myprj -s myprj:Staging:A -r 1234, 4567
+        osc ps request delete myprj -s myprj:Staging:A -r 1234
+        osc ps request list -s myprj:Staging:A
+        osc ps accept myprj -s myprj:Staging:A
+        osc ps show myprj -s myprj:Staging:A
+
+
+        """
+
+        args = slash_split(args)
+
+        subcmdlist = ['create', 'add', 'delete', 'show', 'request', 'rq', 'list', 'accept']
+        rqcmdlist = ['add', 'delete', 'list']
+
+        if cmd not in subcmdlist:
+            raise oscerr.WrongArgs('Unknown operation. Choose one of %s.' \
+                                               % ', '.join(subcmdlist))
+
+        staging_list = opts.staging
+
+        is_subcmd = None
+        rqcmd = None
+        if cmd in ['rq', 'request']:
+            if not args or args[0] not in rqcmdlist:
+                raise oscerr.WrongArgs('Unknown operation. Choose one of %s.' \
+                                                    % ', '.join(rqcmdlist))
+            if args[0] in ['delete', 'add'] and not opts.requests:
+                raise oscerr.WrongArgs('You need to specify at least one request-id with -r')
+            rqcmd = args[0]
+            is_subcmd = True
+            del args[0]
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by coolo** [comment]
```
coolo@f102#osc>osc ps
Traceback (most recent call last):
  File "/suse/coolo/prod/osc/osc-wrapper.py", line 41, in <module>
    r = babysitter.run(osccli)
  File "/space/prod/osc/osc/babysitter.py", line 62, in run
    return prg.main(argv)
  File "/space/prod/osc/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/space/prod/osc/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/space/prod/osc/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/space/prod/osc/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
TypeError: do_projectstaging() missing 1 required positional argument: 'cmd'

coolo:
why do I need to give the name of the staging project to remove from?

lethliel:
> why do I need to give the name of the staging project to remove from?

Because the http_request to remove a sr is: 

`DELETE      staging/home:mstrigl/staging_projects/Staging:A/staged_requests`

and expects a xml with the requests to delete. osc does not know which request is in 
which staging. 

I would have to iterate over all stagings and search for the request to build the url

coolo:
So you let the user do it instead?

coolo:
But if that's the API, the API sucks - easy as that :)

lethliel:
> So you let the user do it instead?

Just explaining why I did it this way :-) I will implement that osc determines the staging for the request when no staging is given with -s

coolo:
I think the API should handle this, really.

lethliel:
> ```
> coolo@f102#osc>osc ps
> Traceback (most recent call last):
>   File "/suse/coolo/prod/osc/osc-wrapper.py", line 41, in <module>
>     r = babysitter.run(osccli)
>   File "/space/prod/osc/osc/babysitter.py", line 62, in run
>     return prg.main(argv)
>   File "/space/prod/osc/osc/cmdln.py", line 344, in main
>     return self.cmd(args)
>   File "/space/prod/osc/osc/cmdln.py", line 367, in cmd
>     retval = self.onecmd(argv)
>   File "/space/prod/osc/osc/cmdln.py", line 501, in onecmd
>     return self._dispatch_cmd(handler, argv)
>   File "/space/prod/osc/osc/cmdln.py", line 1232, in _dispatch_cmd
>     return handler(argv[0], opts, *args)
> TypeError: do_projectstaging() missing 1 required positional argument: 'cmd'
> ```

fixed

hennevogel:
@lethliel whatever you need, please file issues, we want to make this work for you :cowboy_hat_face:

lethliel:
> I think the API should handle this, really.

I put a work around in osc. osc now determines in which staging project the request is, when omitting the -s option

vpereira:
> `DELETE staging/home:mstrigl/staging_projects/Staging:A/staged_requests`

@lethliel 

I'm not sure, but why asking the user to provide the staging workflow project, if it could be inferred from the staging project, why not use it? 

i.e:

`dosc ps rq add home:Admin -s home:Admin:Staging:A -r 1`  IMO should be `dosc ps rq add  -s home:Admin:Staging:A -r 1`

coolo:
How to retrieve the staging workflow for a project through API?

vpereira:
> How to retrieve the staging workflow for a project through API?

Nothing is set in stone. The API is under development and should and will be adapted if necessary. The question should be always if the design make sense or not.  Do you need an endpoint for that? Couldn't we solve it with one extra request?

coolo:
The question is basically: does it make sense at all that the route is `/staging/openSUSE:Leap:15.2/staging_projects/openSUSE:Leap:15.2:Staging:N/staged_requests` if it could be `/staging_projects/openSUSE:Leap:15.2:Staging:N/staged_requests`

vpereira:
> The question is basically: does it make sense at all that the route is `/staging/openSUSE:Leap:15.2/staging_projects/openSUSE:Leap:15.2:Staging:N/staged_requests` if it could be `/staging_projects/openSUSE:Leap:15.2:Staging:N/staged_requests`

It is a good question.. I think we were following the rails convention and the route ended up like that.. we can file an issue for that. How should the osc command look like for you @coolo ?

coolo:
`osc staging -p openSUSE:Leap:15.2 select N bash` is what I use

vpereira:
If in the server side we had something like `Project.find_by(name: 'home:Admin:Staging:A').staging_workflow` then `osc sp rq add -s home:Admin:Staging:A -r 1` would do the trick.

coolo:
I really see no point in typing the extra :Staging: - it's a convention we will stick to

vpereira:
>  I really see no point in typing the extra :Staging: - it's a convention we will stick to

That's a requirement from our PO. The Staging projects are just projects. No special names or hierarchy (example, sub-projects)

coolo:
Yeah, then just don't ask me. Because I can only tell you what users want not what POs want.

vpereira:
> Yeah, then just don't ask me. Because I can only tell you what users want not what POs want.

right @coolo but thinking about the user point of view, would it make sense to have something like ` dosc ps rq list -s home:Admin:Staging:A` instead of ` dosc ps rq list home:Admin  -s home:Admin:Staging:A` or are you ok having to enter the project twice?

coolo:
No, because in my world `ps rq list home:Admin -s A` is sufficient information

vpereira:
@lethliel I see some difference between the example workflows given my the `osc ps` and how it really should be used:

i.e:

```
Example Workflow:
...
osc ps request list -s myprj:Staging:A
```

but what really works is:

`osc ps request list  myprj -s myprj:Staging:A`

vpereira:
As it is today, this command is broken. The XML format that we should send while staging a new request is similar with:

```
<requests><request id="9"/><request id="12"/></requests>
```

Today we are sending something like:

```
<requests><number>1</number></requests>
```

Please update it @lethliel

lethliel:
This will be closed.. Nobody seems to be using it.. 
And rebasing on the current codebase would be more effort than to reimplement this

vpereira:
> This will be closed.. Nobody seems to be using it..
> And rebasing on the current codebase would be more effort than to reimplement this

I think after 3 years, people that need it, just gave up of using `osc` and instead coded their own glue code around staging.

lethliel:
> > This will be closed.. Nobody seems to be using it..
> > And rebasing on the current codebase would be more effort than to reimplement this
> 
> I think after 3 years, people that need it, just gave up of using `osc` and instead coded their own glue code around staging.

There is also a staging plugin in osc-relese-tools. But I guess this is for the old staging API?

lethliel:
> We still want this...

@hennevogel Then I will rebase and commit what I have so far.. So that the people can use it and report bugs ;-)

dmach:

dmach:
Since this new command has subcommands, it might be good to rewrite it to use the new code from #1285 once it gets merged.

dmach:
This PR is still in draft and didn't get any code updates for 6 years. Closing.

PR #632, opened 2019-09-13
tmpfs as new password store backend

lethliel:
With the new TmpFSCredentialsManager it is possible
to store the password in /run/user/<uid> and to keep it
while the session lasts.

As soon as the session is terminated, the directory with
the suffix 'osc' will be deleted and the user has to enter
his password again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ addons:
     - diffstat
 sudo: false
 script: cd tests; python suite.py
+install:
+- pip install fs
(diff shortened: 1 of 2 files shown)


lethliel:
@marcus-h: can you please have a look

dmach:
@lethliel do we still want this considering we have the keyctl backend for python-keyring?

lethliel:
No. This is not needed anymore.

PR #634, opened 2019-09-16
fix checkout of meta files

adrianschroeter:
typo leaded to wrong parameter


bmwiedemann:
Looks more like a copy&paste error.

This patch fixes `osc co -M openSUSE:Factory/_project`

lethliel:
Yes. Nice catch. Thanks for the PR

Issue #635, opened 2019-09-17
[python 3.7] DEFAULT values cannot be None Type. 


lethliel:
osc ls home:mstrigl on a fresh install throws the following when entering a new
user/password (python3.7):

python3.7/configparser.py expects strings not None type.

```
Traceback (most recent call last):
  File "/home/lethliel/git_r/osc/osc-wrapper.py", line 41, in <module>
    r = babysitter.run(osccli)
  File "/home/lethliel/git_r/osc/osc/babysitter.py", line 62, in run
    return prg.main(argv)
  File "/home/lethliel/git_r/osc/osc/cmdln.py", line 336, in main
    self.postoptparse()
  File "/home/lethliel/git_r/osc/osc/commandline.py", line 152, in postoptparse
    conf.interactive_config_setup(e.file, apiurl)
  File "/home/lethliel/git_r/osc/osc/conf.py", line 1013, in interactive_config_setup
    write_initial_config(conffile, config, creds_mgr_descriptor=creds_mgr_descr)
  File "/home/lethliel/git_r/osc/osc/conf.py", line 759, in write_initial_config
    cp = OscConfigParser.OscConfigParser(DEFAULTS)
  File "/home/lethliel/git_r/osc/osc/OscConfigParser.py", line 213, in __init__
    configparser.SafeConfigParser.__init__(self, defaults)
  File "/usr/lib64/python3.7/configparser.py", line 1225, in __init__
    super().__init__(*args, **kwargs)
  File "/usr/lib64/python3.7/configparser.py", line 638, in __init__
    self._read_defaults(defaults)
  File "/usr/lib64/python3.7/configparser.py", line 1216, in _read_defaults
    self.read_dict({self.default_section: defaults})
  File "/usr/lib64/python3.7/configparser.py", line 753, in read_dict
    self.set(section, key, value)
  File "/usr/lib64/python3.7/configparser.py", line 1197, in set
    self._validate_value_types(option=option, value=value)
  File "/usr/lib64/python3.7/configparser.py", line 1182, in _validate_value_types
    raise TypeError("option values must be strings")
TypeError: option values must be strings
```

PR #636, opened 2019-09-17
fix TypeError with python3.7 configparser

lethliel:
configparser.py in python3.7 does not take None Type. It expects only
strings.

fixes https://github.com/openSUSE/osc/issues/635


marcus-h:
Please tell me when done - at least the diff changed in between;)
Since we are willing to "allow" a None value (it will never be
serialized), the easiest might be to override _validate_value_types
in the OscConfigParser class.

PR #639, opened 2019-09-18
do not decode diff before passing it to run_pager

lethliel:
run_pager will take care of the bytes-like object and
will directly write it to sys.stdout.buffer.write() or
NamedTemporaryFile.


marcus-h:
Merged. Thanks!

PR #640, opened 2019-09-18
switch shebang of wrapper script to python3

lethliel:
The osc-wrapper script in git master should use python3 as default.

### DIFF
--- a/osc-wrapper.py
+++ b/osc-wrapper.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # this wrapper exists so it can be put into /usr/bin, but still allows the
 # python module to be called within the source directory during development


marcus-h:
On 2019-09-18 05:35:18 -0700, Marco Strigl wrote:
> The osc-wrapper script in git master should use python3 as default.

Hmm before we switch it to python3, we should probably reconsider the
osc-wrapper.py itself - at least I'm not sure if all the comments
still apply to python3.

mcepl:
Hmm, that’s true … whole wrapper script could (and should) be eliminated in favor of [entry points](https://packaging.python.org/specifications/entry-points/) (or [this HOWTO](https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation) and this answer on [StackOverflow](https://stackoverflow.com/questions/774824/explain-python-entry-points)).

nickbroon:
#527 would allow the wrapper script to be removed.

dmach:
Resolved in different PRs already.

PR #641, opened 2019-09-18
Make check_for_request_on_action on per default.

mcepl:
Fixes #337


mcepl:
@lethliel Do you think the failing tests are because with different defaults code suddenly got into code paths, which were not tested before? Otherwise, I have no idea, why change of one default should make this trouble.

lethliel:
@mcepl: I think it is because the codepath for the search for the open requests is now activated
But I will have a look

marcus-h:
On 2019-09-18 06:39:54 -0700, Matěj Cepl wrote:
> Fixes #337
>
>   https://github.com/openSUSE/osc/pull/641
>
> -- Commit Summary --
>
>   * Make check_for_request_on_action on per default.
>
Hmm why should we change the default? If it is useful for you, just
turn it on:)

mcepl:
@marcus-h Read #641 … there are too many possible ways how to create in-flight collisions of two contributions, and this seems like the most easy and least painful way how to avoid it.

Cc: @scarabeusiv

scarabeusiv:
@lethliel anything preventing the merge of this?

Issue #642, opened 2019-09-19
Error message contains HTTP escaped data


AdamMajer:
Attempting to checkout a non-existent package results in error message that still contains HTTP escaped characters.

```
/tmp/home:adamm:node_test> osc co node1
Server returned an error: HTTP Error 404: Not Found
Error getting meta for project 'home%3Aadamm%3Anode_test' package 'node1'
home:adamm:node_test/node1
```

Repository     : Main Update Repository      
Name           : osc
Version        : 0.165.4-lp151.2.6.1

PR #643, opened 2019-09-20
unquote prj on osc co error message

lethliel:
was:
`Error getting meta for project 'home%3Amstrigl' package 'foo'`

now:
`Error getting meta for project 'home:mstrigl' package 'foo'`

fixes https://github.com/openSUSE/osc/issues/642


lethliel:
This is just a output fix. So --> merge

PR #645, opened 2019-09-20
Resurrect merge_cmd variable after it got removed in the past

dcermak:
In case `diff3` returns something except 0 or 1, osc errors out because the variable `merge_cmd` is not defined. It used to exist, but got removed in the past and is reintroduced via this commit.


marcus-h:
Actually, the Package.mergefile methods needs a bit more love... (for
the commit 74e1f47574c4ef29ac8e0f84ed586745b04772b2).

PR #646, opened 2019-09-25
New browse command to quickly open web browser

lnussel:


marcus-h:
On 2019-09-25 07:12:48 -0700, Ludwig Nussel wrote:
> -- Commit Summary --
>
>   * New browse command to quickly open web browser
>
Hmm I'm a bit hesitant... "browse" is quite generic. Maybe "browser"?
What about shipping this in a separate plugin?
About the UI itself: would something like "osc browse(r) reqid" be useful
(in the future)?

Sorry for the delay and sorry for the questions:)

lnussel:
I named it like that due to "hub browse" which opens github for the repo in $PWD. sure with requests would be interesting too.

lnussel:
so?

marcus-h:
On 2019-10-16 00:24:59 -0700, Ludwig Nussel wrote:
> so?
>
Hmm well - I can live without this feature, but I can also live with
this feature... :) @lethliel any opinion?

If we are going to merge this, we should change subprocess.call to
run_external (from the core module). (And we have a new dependency
to xdg-utils.)

lethliel:
I like the idea.

@lnussel Please use run_external from the core module:

Just change 
`subprocess.call(['xdg-open', url])`

to 
`run_external('xdg-open', url)`

then squash the commits and I will merge this.

lnussel:
Done. Note there are just as many places that use subprocess.call as there are for run_external so you may want to clean that up to provide guidance for casual commiters like me.

lethliel:
> Done. Note there are just as many places that use subprocess.call as there are for run_external so you may want to clean that up to provide guidance for casual commiters like me.

Yes. You are right. I will clean that up and create a PR. Thanks for the hint and thanks for your contribution.

PR #648, opened 2019-10-08
be sure that destination does not exist before overwriting content

adrianschroeter:
we used to crash on utf encoding errors, be sure not to replace meta
data in that case


adrianschroeter:
right, adapted to crash for non 404

marcus-h:
On 2019-10-08 22:36:50 -0700, Adrian Schröter wrote:
> right, adapted to crash for non 404
>
Merged. Thanks a lot!

Issue #48, opened 2013-09-26
feature: osc rq list prj --incoming


DimStar77:
A feature that could be helpful would be an extension to osc rq list, showing only 'incoming' requests.

Example:
osc rq list GNOME:Factory --incoming
would show me only SRs with GNOME:Factory as target.

Currently, one can work around this missing feature by 'excluding known targets' (--exclude-target-project openSUSE:Factory).

Would just be a small 'help' in daily review jobs.

lethliel:
This is kind of old. But the function would be nice nevertheless.

lethliel:
implemented in git master. Will be part of 0.166.0

PR #651, opened 2019-10-16
implement incoming option for osc rq

lethliel:
There is a config option 'include_request_from_project' which
can be set to 0 to exclude all requests originating from the
project in osc [rq|review] list.

To force this behaviour just once (without always editing oscrc
the '--incoming' option can be used.

implements feature request: https://github.com/openSUSE/osc/issues/48

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -2122,6 +2122,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                         help='non-interactive review of request')
     @cmdln.option('--exclude-target-project', action='append',
                         help='exclude target project from request list')
+    @cmdln.option('--incoming', action='store_true',
+                        help='Show only requests where the project is target')
     @cmdln.option('--involved-projects', action='store_true',
                         help='show all requests for project/packages where USER is involved')
     @cmdln.option('--target-package-filter', metavar='TARGET_PACKAGE_FILTER',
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2019-10-16 05:11:52 -0700, Marco Strigl wrote:
> def get_review_list(apiurl, project='', package='', byuser='', bygroup='', byproject='', bypackage='', states=(),
> -                    req_type=''):
> +                    req_type='', incoming=False):
>      # this is so ugly...
>      def build_by(xpath, val):
>          if 'all' in states:
> @@ -4378,7 +4378,7 @@ def build_by(xpath, val):
>      for kind, val in todo.items():
>          xpath_base = 'action/target/@%(kind)s=\'%(val)s\''
>
> -        if conf.config['include_request_from_project']:
> +        if conf.config['include_request_from_project'] and incoming is not True:

"... and not incoming":)

(and the same in get_request_list)

Another option to implement it would be the following (in commandline.py):

if opts.incoming:
    conf.config['include_request_from_project'] = False


(but your code is, of course, also ok)

lethliel:
On 10/21/19 4:03 PM, Marcus Hüwe wrote:
> On 2019-10-16 05:11:52 -0700, Marco Strigl wrote:

<SNIP>

> 
> Another option to implement it would be the following (in commandline.py):
> 
> if opts.incoming:
> conf.config['include_request_from_project'] = False
> 

I haven't thought of this at all. This is (of course) much simpler than
my approach. So not functions and function calls needed to be adapted.

Switched to that solution.

> 
> (but your code is, of course, also ok)

Yes. But your proposal is better ;)

lethliel:
This is tested and can be merged now.

Issue #83, opened 2014-04-11
maintained command should not be aliased to mbranch


yaccz:
maintained documented as an alias to mbranch implies one could do

``` shell
echo ${osc_command} | sed 's/maintained/mbranch/g'
```

However, maintained really is `mbranch --dryrun` and is poorly documented in the man page.

lethliel:
documentation fixed

PR #652, opened 2019-10-16
improve help text for osc maintained

lethliel:
The help text does not indicate that only osc mbranch
really creates a branch.

osc maintained and osc sm are just aliases for osc mbranch --dryrun

fix for https://github.com/openSUSE/osc/issues/83


marcus-h:
Merged. Thanks!

PR #655, opened 2019-10-21
use _repository if alt prj is used

lethliel:
Use _repository as package name if alternative_project is used


lethliel:
@marcus-h: please review also this one

marcus-h:
On 2019-10-21 04:33:13 -0700, Marco Strigl wrote:
> Use _repository as package name if alternative_project is used
>
Hmm this change breaks some invocations where a build_descr is
omitted. But this can be worked around by using "osc buildinfo
<prj> <pkg> <repo> <arch>". This change makes it consistent with
"osc build --alternative-project <prj> ..." => merged. Thanks!

Issue #559, opened 2019-05-17
osc commit fails cryptically when EDITOR points to a nonexistent executable


wstephenson:
$ EDITOR=gibtsnet osc ci
ERROR: service run failed 

a traceback inserted after this unhelpful message reveals that the failure occurs while trying to run the nonexistent external error, but the problem is masked and the service run fails because there is no commit message

Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 7687, in run_external
    return subprocess.call(cmd, **kwargs)
  File "/usr/lib64/python3.6/subprocess.py", line 267, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/usr/lib64/python3.6/subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "/usr/lib64/python3.6/subprocess.py", line 1344, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'gibtsnet': 'gibtsnet'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 4702, in do_commit
    self._commit(subcmd, opts, args)
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 4795, in _commit
    msg = get_commit_msg(prj.absdir, pac_objs[prj_path])
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 7279, in get_commit_msg
    msg = edit_message(footer='\n'.join(footer), template=template)
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 4110, in edit_message
    return edit_text(data, delim, suffix='.diff', template=template)
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 4121, in edit_text
    file_changed = _edit_message_open_editor(filename, data, mtime)
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 4095, in _edit_message_open_editor
    run_editor(filename)
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 4066, in run_editor
    return run_external(cmd[0], *cmd[1:])
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 7691, in run_external
    raise oscerr.ExtRuntimeError(e.strerror, filename)
osc.oscerr.ExtRuntimeError

PR #657, opened 2019-10-23
Better errmsg oscci editor not presentosc ci improve error message on fail

lethliel:
The error message if the call of _commit is just:
"ERROR: service run failed"

One option why this can fail is that the user Editor in env('EDITOR')
is not present.

We check now if e.msg gives a hint about "not found file" and then
error out with a better error message

fixes https://github.com/openSUSE/osc/issues/559


lethliel:
This is just an error message improvement. So --> merge

PR #659, opened 2019-10-24
make credentials.py python2.6 compatible

lethliel:
importlib does not exist in python2.6 so
we need to use __import__ instead

At least SLE11 is affected by this. So the question is:

Do we drop SLE11  and thus python2.6 support? 

@adrianschroeter @marcus-h opinions please


marcus-h:
On 2019-10-24 06:46:22 -0700, Marco Strigl wrote:
> importlib does not exist in python2.6 so
> we need to use __import__ instead
>
> At least SLE11 is affected by this. So the question is:
>
> Do we drop SLE11  and thus python2.6 support?
>
> @adrianschroeter @marcus-h opinions please

Hmm actually I don't like breaking stuff for no "good" reason (but
if its too much work, I'm fine with dropping python26 support). When
does SLE11 reach its EOL?

> index 59c17ace..f51878a3 100644
> --- a/osc/credentials.py
> +++ b/osc/credentials.py
> @@ -304,7 +304,7 @@ def create_credentials_manager(url, cp):
>          creds_mgr_cls = config_entry
>          options = None
>      mod, cls = creds_mgr_cls.rsplit('.', 1)
> -    return getattr(importlib.import_module(mod), cls).create(cp, options)
> +    return getattr(__import__(mod, fromlist=[cls]), cls).create(cp, options)

Hmm isn't this incomplete? We are still importing "importlib" at the
top of the file (this should fail on python2.6).

If we need to support it, we should do something like this:

try:
    import importlib
except ImportError:
    importlib = None


def import_module(mod):
    if importlib is None:
        # python2.6 specific code here
        return ...
    return importlib.import_module(mod)


and then use import_module in create_credentials_manager.

lethliel:
On 10/25/19 12:43 AM, Marcus Hüwe wrote:
> On 2019-10-24 06:46:22 -0700, Marco Strigl wrote:
>  > importlib does not exist in python2.6 so
>  > we need to use __import__ instead
>  >
>  > At least SLE11 is affected by this. So the question is:
>  >
>  > Do we drop SLE11 and thus python2.6 support?
>  >
>  > @adrianschroeter @marcus-h opinions please
> 
> Hmm actually I don't like breaking stuff for no "good" reason (but
> if its too much work, I'm fine with dropping python26 support). When
> does SLE11 reach its EOL?
> 

No it is not much work. This PR makes it work on SLE11.

>  > diff --git a/osc/credentials.py b/osc/credentials.py
>  > index 59c17ace..f51878a3 100644
>  > --- a/osc/credentials.py
>  > +++ b/osc/credentials.py
>  > @@ -304,7 +304,7 @@ def create_credentials_manager(url, cp):
>  > creds_mgr_cls = config_entry
>  > options = None
>  > mod, cls = creds_mgr_cls.rsplit('.', 1)
>  > - return getattr(importlib.import_module(mod), cls).create(cp, options)
>  > + return getattr(__import__(mod, fromlist=[cls]), cls).create(cp, 
> options)
> 
> Hmm isn't this incomplete? We are still importing "importlib" at the
> top of the file (this should fail on python2.6).
> 

Yes. This is incomplete. I added the other stuff. Now it should be complete.

marcus-h:
On 2019-10-25 00:52:16 -0700, Marco Strigl wrote:

...

> +def import_module(mod, cls):
> +    if importlib is None:
> +        return __import__(mod, fromlist=[cls])
> +    return importlib.import_module(mod)

__import__(mod) should be sufficient.

lethliel:
On 10/25/19 1:50 PM, Marcus Hüwe wrote:
> On 2019-10-25 00:52:16 -0700, Marco Strigl wrote:
> 
> ...
> 
>  > +def import_module(mod, cls):
>  > + if importlib is None:
>  > + return __import__(mod, fromlist=[cls])
>  > + return importlib.import_module(mod)
> 
> __import__(mod) should be sufficient.

No. Because __import__(mod) results in __import__(osc.credentials)
and __import__ only imports the top level module 'osc'

This does not work.

marcus-h:
```
  
Date: Thu, 5 Dec 2019 23:29:16 +0100
To: reply+AAIEUTCCRM6RITPUKZ4ZCOF3YAMQFEVBNHHB5A6XDI@reply.github.com
Subject: Re: [openSUSE/osc] make credentials.py python2.6 compatible (#659)
User-Agent: NeoMutt/20180716

On 2019-10-25 05:11:46 -0700, Marco Strigl wrote:
> On 10/25/19 1:50 PM, Marcus Hüwe wrote:
> > On 2019-10-25 00:52:16 -0700, Marco Strigl wrote:
> > 
> > ...
> > 
> >  > +def import_module(mod, cls):
> >  > + if importlib is None:
> >  > + return __import__(mod, fromlist=[cls])
> >  > + return importlib.import_module(mod)
> > 
> > __import__(mod) should be sufficient.
> 
> No. Because __import__(mod) results in __import__(osc.credentials)
> and __import__ only imports the top level module 'osc'
> 
> This does not work.
> 
Sorry, I completely forgot about this.
I need to check what __import__ exactly does. From where can I get the
python26 source that is used in SLE11 (the SUSE:SLE-11 and SUSE:SLE-11:GA
obs projects are empty)?

```

marcus-h:
```
  
Date: Mon, 9 Dec 2019 13:00:03 +0100
To: reply+AAIEUTCCRM6RITPUKZ4ZCOF3YAMQFEVBNHHB5A6XDI@reply.github.com
Subject: Re: [openSUSE/osc] make credentials.py python2.6 compatible (#659)
User-Agent: NeoMutt/20180716

On 2019-10-25 05:11:46 -0700, Marco Strigl wrote:
> +def import_module(mod, cls):
> +    if importlib is None:
> +        return __import__(mod, fromlist=[cls])
> +    return importlib.import_module(mod)

Can you remove the cls parameter (+ fix the caller) and use something
like fromlist=[None] (or your favorite non-empty list, [''] etc.). Actually,
the contents of fromlist should be a str (see ensure_fromlist in
Python/import.c) - however this is only relevant for a package import (which
we don't do in our case (except someone messed up the credentials_mgr_class
setting in her config file)).
The only thing we care about is that fromlist is not empty.

```

Issue #660, opened 2019-10-29
osc 0.166.0 lacks attribute 'encode'


olafhering:
After upgrading osc from 0.165.* to 0.166.0 nothing works anymore, on a SLE_15 system:
````
olaf@bax:~/work/ibs $  ibs co -u home:olh:obs:devel:languages:ocaml:410
Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.6/site-packages/osc/babysitter.py", line 62, in run
    return prg.main(argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 336, in main
    self.postoptparse()
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 145, in postoptparse
    override_verbose = self.options.verbose)
  File "/usr/lib/python3.6/site-packages/osc/conf.py", line 899, in get_config
    password = creds_mgr.get_password(url, user)
  File "/usr/lib/python3.6/site-packages/osc/credentials.py", line 97, in get_password
    return self.decode_password(passwd)
  File "/usr/lib/python3.6/site-packages/osc/credentials.py", line 106, in decode_password
    compressed_pw = base64.b64decode(password.encode("ascii"))
AttributeError: 'NoneType' object has no attribute 'encode'
olaf@bax:~/work/ibs $  osc config
Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.6/site-packages/osc/babysitter.py", line 62, in run
    return prg.main(argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 336, in main
    self.postoptparse()
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 145, in postoptparse
    override_verbose = self.options.verbose)
  File "/usr/lib/python3.6/site-packages/osc/conf.py", line 899, in get_config
    password = creds_mgr.get_password(url, user)
  File "/usr/lib/python3.6/site-packages/osc/credentials.py", line 97, in get_password
    return self.decode_password(passwd)
  File "/usr/lib/python3.6/site-packages/osc/credentials.py", line 106, in decode_password
    compressed_pw = base64.b64decode(password.encode("ascii"))
AttributeError: 'NoneType' object has no attribute 'encode'
olaf@bax:~/work/ibs $  ibs config --change-password
[...]

PR #661, opened 2019-10-29
fix credentials with passx entries

lethliel:
Existing passx enries resulted in a stacktrace because
the cp.get() call for the passx entries was missing.

Also added a delete_password function for ObfuscatedPasswordManager
to delete passx entries

fixes https://github.com/openSUSE/osc/issues/660

### DIFF
--- a/osc/credentials.py
+++ b/osc/credentials.py
@@ -65,6 +65,8 @@ class AbstractCredentialsManager(object):
 
 class PlaintextConfigFileCredentialsManager(AbstractCredentialsManager):
     def get_password(self, url, user, defer=True):
+        if self._cp.has_option(url, 'passx', proper=True):
+            return self._cp.get(url, 'passx', raw=True)
         return self._cp.get(url, 'pass', raw=True)
 
     def set_password(self, url, user, password):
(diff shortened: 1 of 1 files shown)


marcus-h:
diff --git a/osc/credentials.py b/osc/credentials.py
index 59c17ace..ec894218 100644
--- a/osc/credentials.py
+++ b/osc/credentials.py
@@ -65,6 +65,8 @@ def _process_options(self, options):
 
>  class PlaintextConfigFileCredentialsManager(AbstractCredentialsManager):
>     def get_password(self, url, user, defer=True):
> +        if self._cp.has_option(url, 'passx', proper=True):

Hmm proper=True looks fishy (at least old version config parsers don't have this).

> +            return self._cp.get(url, 'passx', raw=True)
>          return self._cp.get(url, 'pass', raw=True)

This code should be moved the the ObfuscatedConfigFileCredentialsManager class.
Otherwise LGTM.

marcus-h:
Oh, my bad! proper=True is part of the OscConfigParser... Sorry!
(But the other comment still holds:) )

PR #666, opened 2019-11-03
Don't enforce password reuse

andreas-schwab:


marcus-h:
Arghs... Merged. Thanks a lot!

Issue #664, opened 2019-11-02
Traceback in xml parsing


frispete:
Hi,

with osc 0.166.1 as of TW 20191030:

```
$ osc linkpac home:frispete:KDE4 kdm-theme-lisa home:frispete:15.0
Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.7/site-packages/osc/babysitter.py", line 62, in run
    return prg.main(argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.7/site-packages/osc/commandline.py", line 2874, in do_linkpac
    link_pac(src_project, src_package, dst_project, dst_package, opts.force, rev, opts.cicount, opts.disable_publish, opts.new_package, vrev)
  File "/usr/lib/python3.7/site-packages/osc/core.py", line 5145, in link_pac
    root = ET.fromstring(''.join(dst_meta))
TypeError: sequence item 0: expected str instance, bytes found
```

lethliel:
Thanks for reporting. Will have a look next week...

frispete:
Thanks for fixing.

PR #668, opened 2019-11-04
fix decoding for osc linkpac

lethliel:
in def link_pac dst_meta is a list of bytes-like objects.
so b''.join(dst_meta) is needed

fixes https://github.com/openSUSE/osc/issues/664


marcus-h:
Merged. Thanks!

PR #644, opened 2019-09-20
Transfer the name of the input file to vc

lethliel:
instead of the content and let vc do the rest.
This prevents calling echo -e and interpreting
escape characters.

This is a prototype fix for https://github.com/openSUSE/osc/issues/637
Needs https://github.com/openSUSE/obs-build/pull/525 in obs-buiild/vc to work.

PR #672, opened 2019-11-11
extend cmd list for vc with args

lethliel:
If args is not none extend the list for the command
vc call with the args. Otherwise this will not work:

osc vc --file=proposed.changes package.changes

fixes boo#1155953

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -8980,8 +8980,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                     raise oscerr.WrongOptions('\'%s\': is no file' % opts.file)
                 args = list(args)
                 if not args:
-                    cmd_list.append('')
-                cmd_list.append(opts.file)
+                    args.append('')
+                args.append(opts.file)
 
             if opts.just_edit:
                 cmd_list.append("-e")


marcus-h:
On 2019-11-11 04:09:22 -0800, Marco Strigl wrote:
> If args is not none extend the list for the command
> vc call with the args. Otherwise this will not work:
>
> osc vc --file=proposed.changes package.changes
>
> fixes boo#1155953

Hrm let's just do it as described in [1]. Then, we avoid an additional
else branch + we do not pass excess arguments to the vc script.


[1] https://github.com/openSUSE/osc/pull/644#issuecomment-533514712

lethliel:
I have done the change. 

osc vc --file foo.bar foo.changes has written foo.changes into foo.bar.

But the cmd_list.extend(args) needs to be done before the cmd_list.append(opts.file)
because vc expects the changes filename first and then the file from which the content should be
taken.

After my change everything is correct

marcus-h:
On 2019-11-12 06:49:54 -0800, Marco Strigl wrote:
> After my change everything is correct
>
Hmm won't this break the "osc vc my.changes" codepath? The change should
look like this:

index 8d0649a..2b94b14 100644
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -8980,8 +8980,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                     raise oscerr.WrongOptions('\'%s\': is no file' % opts.file)
                 args = list(args)
                 if not args:
-                    cmd_list.append('')
-                cmd_list.append(opts.file)
+                    args.append('')
+                args.append(opts.file)

             if opts.just_edit:
                 cmd_list.append("-e")

Issue #671, opened 2019-11-10
`osc checkout` with checkout_no_colon still checks out colon directory


mook:
When `checkout_no_colon` is set, `osc checkout $project` still creates the colon directory.  It _also_ creates the subdirectories, if it's checking out packages.

```
$ osc checkout foo:bar:baz
A    foo:bar:baz
A    foo/bar/baz
A    foo/bar/baz/some-package
...
```

In [`do_checkout`]:
1. `project` and `project_dir` are set to `foo:bar:baz` on [line 4462]
2. `prj_dir` is set to `project` (== `foo:bar:baz`) on [line 4507]
3. The coloned version is then created in [line 4516]
4. [line 4541] then calls [`checkout_package`], which handles `checkout_no_colon` on [line 5020]. (Except it doesn't seem to on Windows.)

[`do_checkout`]: https://github.com/openSUSE/osc/blob/f78057e2d53e08b62658561278e9617809e683fa/osc/commandline.py#L4408
[line 4462]: https://github.com/openSUSE/osc/blob/f78057e2d53e08b62658561278e9617809e683fa/osc/commandline.py#L4462
[line 4507]: https://github.com/openSUSE/osc/blob/f78057e2d53e08b62658561278e9617809e683fa/osc/commandline.py#L4507
[line 4516]:https://github.com/openSUSE/osc/blob/f78057e2d53e08b62658561278e9617809e683fa/osc/commandline.py#L4516
[line 4541]: https://github.com/openSUSE/osc/blob/f78057e2d53e08b62658561278e9617809e683fa/osc/commandline.py#L4541
[`checkout_package`]: https://github.com/openSUSE/osc/blob/f78057e2d53e08b62658561278e9617809e683fa/osc/core.py#L5004
[line 5020]: https://github.com/openSUSE/osc/blob/f78057e2d53e08b62658561278e9617809e683fa/osc/core.py#L5020

The fix is probably to make [line 4507] check if `checkout_no_colon` is set, and set `prj_dir` accordingly.  I have not looked enough at the rest of the code base to be confident at producing a patch that uses any helpers correctly, though.

PR #673, opened 2019-11-11
fix checkout_no_colon on project level

lethliel:
prj_dir was not altered when issuing osc co on
project level.

Solution: Replace ":" with "/" on project level

fixes https://github.com/openSUSE/osc/issues/671

Thanks to @mook

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -4505,6 +4505,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
 
         elif project:
             prj_dir = opts.output_dir if opts.output_dir else project
+            if not opts.output_dir and conf.config['checkout_no_colon']:
+                prj_dir = prj_dir.replace(':', '/')
             if sys.platform[:3] == 'win':
                 prj_dir = prj_dir.replace(':', ';')
             if os.path.exists(prj_dir):


marcus-h:
```
  
Date: Thu, 6 Feb 2020 20:50:13 +0100
To: reply+AAIEUTEAHJV3TLSD56EC5QV322FFVEVBNHHB6DHQRY@reply.github.com
Subject: Re: [openSUSE/osc] fix checkout_no_colon on project level (#673)

On 2019-11-11 04:45:14 -0800, Marco Strigl wrote:
> index 8d0649a8..62610176 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -4505,6 +4505,8 @@ def do_checkout(self, subcmd, opts, *args):
>  
>          elif project:
>              prj_dir = opts.output_dir if opts.output_dir else project
> +            if conf.config['checkout_no_colon']:
> +                prj_dir = prj_dir.replace(':', '/')

Hmm the question is if we should do this if --output-dir foo:bar:baz
was specified. That is,

the code from above

VS.

if not opts.output_dir and conf.config['checkout_no_colon']:
    prj_dir = prj_dir.replace(':', '/')


(I'm fine with either variants)

```

lethliel:
Yes.

If the user has given a output_dir this should be considered. Otherwise
we will override what the user wants.

Changed it.


-- 
SUSE Software Solutions Germany GmbH, Maxfeldstr. 5, 90471 Nürnberg
(HRB 36809, AG Nürnberg), Geschäftsführer: Felix Imendöffer

Issue #675, opened 2019-11-13
traceback with http_debug=1 on osc build


mcepl:
```
-- GET https://api.opensuse.org/source/home:mcepl:neovim/_pubkey
send: b'GET /source/home:mcepl:neovim/_pubkey HTTP/1.1\r\nAccept-Encoding: identity\r\nHost: api.opensuse.org\r\nUser-agent: osc/0.166.1\r\nConnection: close\r\n\r\n'
reply: 'HTTP/1.1 404 home mcepl neovim  no pubkey available\r\n'
header: Date: Wed, 13 Nov 2019 06:41:29 GMT
header: Server: Apache/2.4.23 (Linux/SUSE)
header: Strict-Transport-Security: max-age=31536000
header: Cache-Control: private, no-transform
header: Referrer-Policy: strict-origin-when-cross-origin
header: X-Permitted-Cross-Domain-Policies: none
header: X-XSS-Protection: 1; mode=block
header: X-Opensuse-Runtimes: {"view":null,"db":8.34656460210681,"backend":0}
header: X-Request-Id: 038574a3-c183-4a65-ada0-8538cd5f9559
header: X-Opensuse-APIVersion: 2.11~alpha.20191112T141018.f43864e078
header: X-Download-Options: noopen
header: X-Runtime: 0.030336
header: X-Frame-Options: SAMEORIGIN
header: X-Content-Type-Options: nosniff
header: X-Powered-By: Phusion Passenger 6.0.2
header: Content-Type: text/xml
header: Cache-Control: private, no-transform
header: Content-Length: 154
header: Connection: close
Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.7/site-packages/osc/babysitter.py", line 62, in run
    return prg.main(argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
[...]

PR #676, opened 2019-11-13
fix output in http_debug mode for pubkey not found

lethliel:
Variable e is not known at the time of the call and
e.strerror is None nevertheless. So just ommit e.strerror.

fixes https://github.com/openSUSE/osc/issues/675

### DIFF
--- a/osc/fetch.py
+++ b/osc/fetch.py
@@ -286,7 +286,7 @@ class Fetcher:
 
             if try_parent:
                 if self.http_debug:
-                    print("can't fetch key for %s: %s" % (i, e.strerror), file=sys.stderr)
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2019-11-13 02:19:17 -0800, Marco Strigl wrote:
> index d84474ee..4578e892 100644
> --- a/osc/fetch.py
> +++ b/osc/fetch.py
> @@ -286,7 +286,7 @@ def run(self, buildinfo):
>
>              if try_parent:
>                  if self.http_debug:
> -                    print("can't fetch key for %s: %s" % (i, e.strerror), file=sys.stderr)
> +                    print("can't fetch key for %s:" % (i), file=sys.stderr)

Can you also remove the colon ':'?                  ^
Otherwise, LGTM!

PR #677, opened 2019-11-13
fix decoding for osc aggregatepac

andreas-schwab:


lethliel:
LGTM. Thanks for the patch.

PR #678, opened 2019-11-14
add --lastsucceeded option for logfile display

adrianschroeter:
no single option since it is really a special usecase IMHO
jsc#OBS-40


adrianschroeter:
fixed conflicts again. Can we merge?

lethliel:
LGTM

PR #681, opened 2019-11-18
Swap all usage of cgi.escape() with html.escape()

Conan-Kudo:
In Python 3.8, `cgi.escape()` has been completely removed after
being deprecated since Python 3.2. The suggested replacement
is to use `html.escape()`.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -1241,7 +1241,7 @@ class Osc(cmdln.Cmdln):
         if len(args) < 2 and is_project_dir(os.getcwd()):
             if opts.diff:
                 raise oscerr.WrongOptions('\'--diff\' is not supported in a project working copy')
-            import cgi
+            import html
             project = store_read_project(os.curdir)
 
             sr_ids = []
@@ -1295,7 +1295,7 @@ class Osc(cmdln.Cmdln):
                         (project, target_prj_block, options_block)
                 actionxml += s
                 xml = """<request> %s <state name="new"/> <description>%s</description> </request> """ % \
-                        (actionxml, cgi.escape(opts.message or ""))
+                        (actionxml, html.escape(opts.message or "", False))
                 u = makeurl(apiurl, ['request'], query='cmd=create&addrevision=1')
                 f = http_POST(u, data=xml)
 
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2019-11-17 22:03:54 -0800, Neal Gompa (ニール・ゴンパ) wrote:
> In Python 3.8, `cgi.escape()` has been completely removed after
> being deprecated since Python 3.2. The suggested replacement
> is to use `html.escape()`.

Hmm there are two "bogus" calls:

> @@ -1295,7 +1295,7 @@ def _check_service(root):
>                          (project, target_prj_block, options_block)
>                  actionxml += s
>                  xml = """<request> %s <state name="new"/> <description>%s</description> </request> """ % \
> -                        (actionxml, cgi.escape(opts.message or ""))
> +                        (actionxml, html.escape(opts.message or ""), False)
>                  u = makeurl(apiurl, ['request'], query='cmd=create&addrevision=1')
>                  f = http_POST(u, data=xml)
>
> @@ -1864,9 +1864,9 @@ def do_createrequest(self, subcmd, opts, *args):
>          if not opts.message:
>              opts.message = edit_message()
>
> -        import cgi
> +        import html
>          xml = """<request> %s <state name="new"/> <description>%s</description> </request> """ % \
> -              (actionsxml, cgi.escape(opts.message or ""))
> +              (actionsxml, html.escape(opts.message or ""), False)
>          u = makeurl(apiurl, ['request'], query='cmd=create')
>          f = http_POST(u, data=xml)

The False belongs to the html.escape call.

python2 has no default html module. We should probably move this into
a helper in util/helper.py

def xml_escape(data):
    ...


But maybe we should check first, if python's elementtree module provides
a function that does xml escaping.

Conan-Kudo:
@marcus-h Fixed the bogus calls.

Conan-Kudo:
@marcus-h There is a backports module for Python 2: [`backports.html`](https://pypi.org/project/backports.html/)

marcus-h:
On 2019-11-18 05:43:22 -0800, Neal Gompa (ニール・ゴンパ) wrote:
> @marcus-h There is a backports module for Python 2: [`backports.html`](https://pypi.org/project/backports.html/)
>
In the meantime, we solved it a bit differently. See commits
33bbc57b5f1b4cd3aded4a696b095355a8fa4805 (and
6dbf103e1030494381c2fbb384f9648a78b68ce6).

=> Closing. Anyway, thanks a lot for the PR!

PR #682, opened 2019-11-18
core: Switch from platform.linux_distribution() to distro.id()

Conan-Kudo:
The `platform.linux_distribution()` and `platform.dist()` methods have been removed in Python 3.8.

The suggested replacement is using the third party `distro` module.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -4022,11 +4022,8 @@ def get_default_editor():
     if system == 'Windows':
         return 'notepad'
     if system == 'Linux':
-        try:
-            # Python 2.6
-            dist = platform.linux_distribution()[0]
-        except AttributeError:
-            dist = platform.dist()[0]
+        import distro
+        dist = distro.id()
         if dist == 'debian':
             return 'editor'
         elif dist == 'fedora':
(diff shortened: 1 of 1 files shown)


lethliel:
Your solution needs the module python-distro which is not available on all platforms we support osc (like older SLE12 Versions).

Can we do something like this:
if sys.version_info >= (3, 8):
    import dist
    # your code
else:
    # old code

Then we do not have a problem on older systems where the dist module is not available.

Conan-Kudo:
@lethliel why not just pull a version of `python-distro` into `openSUSE:Tools`?

lethliel:
@Conan-Kudo No. Because we officially don't have repos internally.

Conan-Kudo:
@lethliel Do I still need to keep the Python 2.4.x function? Or can I move up to Python 2.6+?

lethliel:
@Conan-Kudo 2.6+ is fine

tomaskrizek:
Backwards compatible version: https://github.com/openSUSE/osc/pull/695

Conan-Kudo:
Closed in favor of #695

Issue #683, opened 2019-11-18
Python 2 to 3 conversion error in bugowner command (TypeError: sequence item 0: expected str instance, bytes found)


krig:
Command executed:
```
osc bugowner -d <user> <project> <package>
```

Result:
```
TypeError: sequence item 0: expected str instance, bytes found
Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.6/site-packages/osc/babysitter.py", line 62, in run
    return prg.main(argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 8289, in do_maintainer
    delPerson(apiurl, prj, pac, opts.delete, role)
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 6975, in delPerson
    root = ET.fromstring(''.join(data))
TypeError: sequence item 0: expected str instance, bytes found
```

PR #684, opened 2019-11-18
Handle bytes vs. str error when parsing meta (#683)

krig:
In all the cases where meta_exists returns either
string data, bytes data or a list, the output needs
to be parsed correctly.

Signed-off-by: Kristoffer Grönlund <kgronlund@suse.com>

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -7973,7 +7973,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                                        'name': pac,
                                        'user': user}), apiurl=apiurl)
                 if data:
-                    data = ET.fromstring(''.join(data))
+                    data = ET.fromstring(parse_meta_to_string(data))
                     data.find('title').text = ''.join(title)
                     data.find('description').text = ''.join(descr)
                     data.find('url').text = url
--- a/osc/core.py
+++ b/osc/core.py
@@ -3802,6 +3802,19 @@ def make_meta_url(metatype, path_args=None, apiurl=None, force=False, remove_lin
 
     return makeurl(apiurl, [path], query)
 
+def parse_meta_to_string(data):
+    """
+    Converts the output of meta_exists into a string value
+    """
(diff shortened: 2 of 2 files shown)


marcus-h:
On 2019-11-18 00:39:40 -0800, Kristoffer Grönlund wrote:
> In all the cases where meta_exists returns either
> string data, bytes data or a list, the output needs
> to be parsed correctly.
>
Hmm maybe we should simply fix meta_exists once and for all...
I need to think about it.

lethliel:
@marcus-h Have you thought about his? Otherwise I would like to merge this.

marcus-h:
On 2020-02-06 01:58:14 -0800, Marco Strigl wrote:
> @marcus-h Have you thought about his?

Yes... conclusion: the whole meta_exists implementation is completely
insane and misleading (we even have code that does something like
"if meta_exists(...): ... else: ...").
The only sane thing would be a cleanup, but I'm not sure when I have
time for it...

About the patch: the patch itself is OK (except that we call
parse_meta_to_string(...) even if it is not needed (it is only
needed if the "create" parameter is set to True)). Also, it misses
one caller in commandline.py (the one in do_importsrcpkg) (I haven't,
checked the ones in core.py, though).

> Otherwise I would like to merge this.
>
I don't want to block this. Since it works around real bugs, it
is (IMHO) OK to accept the patch as-is...

@krig: are you in the mood for fixing the caller in commandline.py?

(Btw., sorry for the long delay!:/ )

krig:
@marcus-h Sure, I'll take a look at that

krig:
@marcus-h I think I found the one you meant, but please do double-check :)

marcus-h:
On 2020-02-07 01:03:20 -0800, Kristoffer Grönlund wrote:
> @marcus-h I think I found the one you meant, but please do double-check :)
>
Awesome! Thanks a lot for the patch! (And your patience;) )
Merged.

PR #685, opened 2019-11-21
Packaging in git

nickbroon:
It is far easier to track change and modify the packaging files if they are kept in git (particularly the debian one instead of being a tarball).
See https://build.opensuse.org/package/view_file/home:nickbrown:branches:openSUSE:Tools/osc/_service?expand=1 for how these are then used in the _service file for osc on OBS. I'll followup with SR on OBS to https://build.opensuse.org/package/show/openSUSE:Tools/osc to use this if the merge is accepted.

This is based on how obs-build does this:
https://build.opensuse.org/package/show/openSUSE:Tools/build
https://github.com/openSUSE/obs-build/tree/master/dist

With the packaging in Git, I hope to more easily follow this up with work to improve the debian packaging for python3.


Conan-Kudo:
@nickbroon Packages that are maintained in the openSUSE distribution are not allowed to use this packaging strategy.

nickbroon:
I copied the packaging strategy of obs-build. Is that not part of openSUSE?

nickbroon:
To allow the files to be more easily versioned in OBS: https://build.opensuse.org/request/show/752964, which would make this git PR less important. Can this OBS request be accepted?

Conan-Kudo:
It may also be worth considering just switching to using the spec file to build Debian packages...

nickbroon:
I'm not big fan of debbuild as it does not support things like generating package dependencies for python, and does not generate debugs, etc. I'm sure these can all be solved, but given that OBS has support for debian source packages seems reasonable to use the native format.
The SR in OBS I submitted, just expands the native debian tarball, so that contents can be more easily version controlled.
I'd be happy with either this Git PR (based on what obs-build does) being merged or the OBS SR being merged. Either would make maintaining the debian packaging much easier. (and I plan to follow either up with a tidy up of the packaging to have osc use python3).

Conan-Kudo:
> package dependencies for python

That doesn't apply for this project as it doesn't use setuptools. But yes, I'd like to fix this...

> does not generate debugs

I don't know what this is?

nickbroon:
Closing in favour of: https://build.opensuse.org/request/show/752964

PR #691, opened 2019-12-02
"osc shell" uses --noinit by default now if buildroot exists

adrianschroeter:


osc/build.py
```diff
@@ -866,13 +869,14 @@ def main(apiurl, opts, argv):
     # real arch of this machine
     # vs.
     # arch we are supposed to build for
-    if bi.hostarch != None:
```
darix: is this an unrelated change?

lethliel:
@adrianschroeter Please open separate PRs for the commits. As they have nothing in common.

osc/build.py
```diff
@@ -675,6 +673,11 @@ def main(apiurl, opts, argv):
         except:
             pass
 
+    if opts.shell:
+        buildargs.append("--shell")
+        if os.path.exists(build_root) and not opts.clean:
+            buildargs.append("--noinit")
```
lethliel: opts.noinit must be set to True (otherwise the buildinfo and the buildconfig will be downloaded from the server again.)
opts.offline is also needed if the build_root exists.

adrianschroeter: done

lethliel:
Thanks for the patch. And LGTM

PR #692, opened 2019-12-03
fix building for "local" arch

adrianschroeter:


mlschroe:
I think the buildinfo should contain a correct 'repoarch' element for the 'local' packages.

mlschroe:
OTOH maybe osc should not request the buildinfo for arch 'local'?

adrianschroeter:
On Dienstag, 3. Dezember 2019, 13:50:28 CET Michael Schroeder wrote:
> OTOH maybe osc should not request the buildinfo for arch 'local'?

The problem is that "x86_64" itself might not be part of the repository
definition.

So, we would need to map local -> $local_arch on the server side.

-- 

mlschroe:
The problem is that '$local_arch' might not match your system.

adrianschroeter:
On Dienstag, 3. Dezember 2019, 14:37:41 CET Michael Schroeder wrote:
> The problem is that '$local_arch' might not match your system.

yes ... also not the case with my solution proposed here, but we could
at least cover this in osc by checking local arch.

-- 

adrianschroeter:
okay, mlschroe will solve this on the repo server side to support building against "local" arch. No need for this patch anymore.

PR #695, opened 2019-12-04
core: support distro module for platform detection

tomaskrizek:
Python 3.8 no longer supports platform.linux_distribution().

Compatibility with Python 2.6+ is kept. The
platform.linux_distribution() function is available since 2.6:
https://docs.python.org/2.6/library/platform.html?platform.linux_distribution#platform.linux_distribution

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -23,7 +23,13 @@ import socket
 import errno
 import shlex
 import hashlib
+import platform
 
+
+try:
+    import distro
+except ImportError:
+    distro = None
 
 try:
     from urllib.parse import urlsplit, urlunsplit, urlparse, quote_plus, urlencode, unquote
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@tomaskrizek There are typos in your commit message.

tomaskrizek:
Hopefully I got it right this time :)

marcus-h:
On 2019-12-04 08:23:23 -0800, Tomas Krizek wrote:
> index 0683d0b8..b4d959d6 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -4001,17 +4001,26 @@ def read_meta_from_spec(specfile, *args):
>
>      return spec_data
>
> +def get_linux_distro():
> +    try:
> +        import distro
> +    except ImportError:
> +        if sys.version_info >= (3, 8):
> +            return None
> +        else:  # compatibility for Python 2.6 to 3.7
> +            try:
> +                return platform.linux_distribution()[0]
> +            except (AttributeError, KeyError):

Hmm why do you catch a KeyError here? (Since we do not care about
pre python26 anymore, we can also drop the AttributeError.)

> +                return None
> +    return distro.id()

Hmm this won't work because a platform import is missing. I think we
could also simplify the code a bit. Maybe something like this:

def _get_linux_distro():
    try:
        import distro
        return distro.id()
    except ImportError:
        if sys.version_info >= (3, 8)
            return None
    # backward compatibility for python 2.6 to 3.7
    import platform
    return platform.linux_distribution()[0]

Or maybe a bit more concise:

try:
    import distro
except:
    distro = None


import platform


def _get_linux_distro():
    if distro is not None:
        return distro.id()
    elif sys.version_info >= (3, 8):
        return None
    return platform.linux_distribution()[0]

What do you think?

tomaskrizek:
Thanks for the review, I updated the MR as suggested. There was no reason for catching the `KeyError` (or `AttributeError` since only Python 2.6+ is supported).

marcus-h:

PR #696, opened 2019-12-05
fix decoding in fetch_cpio

lethliel:


marcus-h:
On 2019-12-05 04:03:25 -0800, Marco Strigl wrote:
> index 9e3adb5e..920ab591 100644
> --- a/osc/fetch.py
> +++ b/osc/fetch.py
> @@ -73,8 +73,8 @@ def __download_cpio_archive(self, apiurl, project, repo, arch, package, **pkgs):
>                      # getbinarylist instead of the public/... route
>                      # (which is routed to getbinaries)
>                      # getbinaries does not support kiwi builds
> -                    if hdr.filename == '.errors':
> -                        archive.copyin_file(hdr.filename)
> +                    if hdr.filename == b'.errors':

Good catch!

> +                        archive.copyin_file(decode_it(hdr.filename))

Decoding the filename is wrong (this code will result in a CPIOError).
Can you remove the decode_it call again?

(At some point we could also "fix" the CpioError(...) code in cpio.py (because
we "squeeze" a bytes into str), but this is ok for now).

>                          raise oscerr.APIError('CPIO archive is incomplete '
>                                                '(see .errors file)')
>                      if package == '_repository':
>

lethliel:
> + archive.copyin_file(decode_it(hdr.filename))
> Decoding the filename is wrong (this code will result in a CPIOError). Can you remove the decode_it call again? (At some point we could also "fix" the CpioError(...) code in cpio.py (because we "squeeze" a bytes into str), but this is ok for now).
> […](#)
> raise oscerr.APIError('CPIO archive is incomplete ' '(see .errors file)') if package == '_repository':

Ok. Reverted the decoding in https://github.com/openSUSE/osc/pull/704

PR #698, opened 2019-12-06
fix Fetch api to become compatible again.

adrianschroeter:
Broke in db795c81219fad406b7fd19341ed29530cb45c9b as reported by Marcus

### DIFF
--- a/osc/fetch.py
+++ b/osc/fetch.py
@@ -28,8 +28,8 @@ from .meter import create_text_meter
 
 class Fetcher:
     def __init__(self, cachedir='/tmp', api_host_options={}, urllist=[],
-                 modules = [], http_debug=False, cookiejar=None, offline=False,
-                 enable_cpio=True):
+                 http_debug=False, cookiejar=None, offline=False,
+                 enable_cpio=True, modules=[]):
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2019-12-05 23:21:51 -0800, Adrian Schröter wrote:
> index bcf17c03..6ad06896 100644
> --- a/osc/fetch.py
> +++ b/osc/fetch.py
> @@ -28,8 +28,8 @@
>
>  class Fetcher:
>      def __init__(self, cachedir='/tmp', api_host_options={}, urllist=[],
> -                 modules = [], http_debug=False, cookiejar=None, offline=False,
> -                 enable_cpio=True):
> +                 http_debug=False, cookiejar=None, offline=False,
> +                 enable_cpio=True, modules = []):

One minor nitpick: "modules=[]" instead of "modules = []". I know the latter
is still used in numerous places, but we should avoid to introduce more such
"style violations" (PEP8) (IMHO).

Are you in the mood for fixing this?:)

Otherwise, LGTM and thanks!

adrianschroeter:
sure, done

lethliel:
Thanks :)

Issue #709, opened 2019-12-11
Regression: osc chroot --wipe does not work with --root


nickbroon:
```
~$ cd ~
~$ osc chroot --wipe --root=/var/lib/jenkins/workspace/job/build-root/project-standard-x86_64
Error: '/home/user' is not an osc project dir or working copy
```

nickbroon:
See also #706 for related issue when using --force

lethliel:
@nickbroon Are you issuing this command outside the package directory? 

When I use it inside an package working copy it just works fine. 
If I use it outside I get:
`Missing argument: build description (for example a spec, dsc or kiwi file)`

nickbroon:
Yes, It's being used from a user home directory, that is not a package working copy.

lethliel:
You are right. This is a regression. I am working on the fix at the moment.

nickbroon:
Fab, thanks. I watch for a new release with this fix appearing in openSUSE:Tools on OBS.

nickbroon:
@lethliel Did you get a chance to publish a new version to https://build.opensuse.org/package/show/openSUSE:Tools/osc ?

lethliel:
I will release a new version today. But it also needs the new version of the build script to work. But that will be released today as well.

Issue #706, opened 2019-12-10
Regression: osc chroot does not accept `--force` option with --wipe.


nickbroon:
```
osc chroot --wipe --force --root /var/lib/jenkins/workspace/job/build-root/project-standard-x86_64
osc chroot: no such option: --force
Try 'osc help chroot' for info.
```

man page states:

```
                  -f, --force
                              Do not ask confirmation for wipe
                  --wipe      Delete the build root instead of chrooting into it
```

adrianschroeter:
On Dienstag, 10. Dezember 2019, 15:21:41 CET Nick Brown wrote:
> ```
> osc chroot --wipe --force --root /var/lib/jenkins/workspace/job/build-root/project-standard-x86_64
> osc chroot: no such option: --force
> Try 'osc help chroot' for info.
> ```
> 
> man page states:
> 
> ```
>                   -f, --force
>                               Do not ask confirmation for wipe
>                   --wipe      Delete the build root instead of chrooting into it
> ```

I think we should just drop the -f part from the help ... can do so


-- 

Adrian Schroeter <adrian@suse.de>
Build Infrastructure Project Manager

SUSE Software Solutions Germany GmbH,  Maxfeldstr. 5, 90409 Nuernberg, Germany
(HRB 247165, AG München), Geschäftsführer: Felix Imendörffer

nickbroon:
This used to work.
Without the `--force` option how would a non-interactive script remove the build chroot?

```
~$ osc chroot --wipe --force --root=/var/lib/jenkins/workspace/job/build-root/project-standard-x86_64
osc chroot: no such option: --force
Try 'osc help chroot' for info.

~$ osc chroot --wipe --root=/var/lib/jenkins/workspace/job/build-root/project-standard-x86_64
Error: '/home/user' is not an osc project dir or working copy
```

lethliel:
@nickbroon @adrianschroeter 

This is a regression. Wiping the local build root should be possible without the use of a WC. 

We should bypass parse_repoarchdescr() or just invoke the wipe call of the build script before that. 

I am working on a fix. ETA tomorrow morning.

nickbroon:
Fixed by #710

PR #710, opened 2019-12-11
fix regression in osc chroot

lethliel:
This fixes some regressions with osc chroot:
    
- osc chroot --wipe --root=/dir/ can now be called outside
   a working copy
- osc chroot --noinit --root=/dir/ can now be called outside
  a working copy and behaves like the old code (Just entering
  the chroot without any modifications)
- The confirmation of the deletion is implemented again and thus
   the --force option was implemented too.
    
This needs all to be done before self.parse_repoarchdescr() call. If no opts.noinit is set the opts.shell is set to true and the buildroot gets initialized before entering the shell.

fixes https://github.com/openSUSE/osc/issues/709, https://github.com/openSUSE/osc/issues/706 and https://github.com/openSUSE/osc/issues/711

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -524,6 +524,32 @@ def get_kiwipath_from_buildinfo(apiurl, bi_filename, prj, repo):
     kiwipath.insert(0, myprp)
     return kiwipath
 
+def calculate_prj_pac(opts, descr):
+    project = opts.alternative_project or store_read_project('.')
+    if opts.local_package:
+        package = os.path.splitext(os.path.basename(descr))[0]
+    else:
+        package = store_read_package('.')
+    return project, package
+
+def calculate_build_root(apihost, prj, pac, repo, arch):
+    buildroot = os.environ.get('OSC_BUILD_ROOT', config['build-root']) \
+                            % {'repo': repo, 'arch': arch, 'project': prj, 'package': pac, 'apihost': apihost}
+    return buildroot
+
+def run_build(*args):
+    cmd = [config['build-cmd']]
+    cmd += args
+
+    sucmd = os.environ.get('OSC_SU_WRAPPER', config['su-wrapper']).split()
+    if sucmd[0] == 'su':
+        if sucmd[-1] == '-c':
+            sucmd.pop()
+        cmd = sucmd + ['-s', cmd[0], 'root', '--'] + cmd[1:]
+    else:
+        cmd = sucmd + cmd
+    return run_external(cmd[0], *cmd[1:])
+
 def main(apiurl, opts, argv):
 
     repo = argv[0]
@@ -696,12 +722,6 @@ def main(apiurl, opts, argv):
 
     if opts.shell:
         buildargs.append("--shell")
-        if os.path.exists(build_root) and os.path.exists(bi_filename) and not opts.clean and not opts.extra_pkgs:
-            opts.noinit = True
-            opts.offline = True
-            # we should check if the service did run before and only skip it then,
-            # but we have no save point for this atm
-            opts.noservice = True
 
     if opts.noinit:
         buildargs.append('--noinit')
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -6341,6 +6341,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                   help=SUPPRESS_HELP)
     @cmdln.option('--shell-cmd', metavar='COMMAND',
                   help='run specified command instead of bash')
+    @cmdln.option('-f', '--force', action='store_true',
+                  help='Do not ask for confirmation to wipe')
     @cmdln.option('--host', metavar='HOST',
             help='perform the build on a remote server - user@server:~/remote/directory')
     @cmdln.option('--trust-all-projects', action='store_true',
@@ -6425,9 +6427,6 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         if opts.debuginfo and opts.disable_debuginfo:
             raise oscerr.WrongOptions('osc: --debuginfo and --disable-debuginfo are mutual exclusive')
 
-        if subcmd == 'chroot' or subcmd == 'shell':
-            opts.shell = True
-
         if subcmd == 'wipe':
             opts.wipe = True
 
@@ -6449,6 +6448,33 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                args = [ lastbuildroot[0], lastbuildroot[1] ]
                if not opts.vm_type:
                   opts.vm_type = lastbuildroot[2]
+
+        vm_chroot = opts.vm_type or conf.config['build-type']
+        if (subcmd in ('shell', 'chroot') or opts.shell or opts.wipe) and not vm_chroot:
+            if opts.root:
+                build_root = opts.root
+            else:
+                args = self.parse_repoarchdescr(args, opts.noinit or opts.offline, opts.alternative_project, False, opts.vm_type, opts.multibuild_package)
+                repo, arch, build_descr = args
+                prj, pac = osc.build.calculate_prj_pac(opts, build_descr)
+                apihost = urlsplit(self.get_api_url())[1]
+                build_root = osc.build.calculate_build_root(apihost, prj, pac, repo,
+                                                    arch)
+            if opts.wipe and not opts.force:
+                    # Confirm delete
+                    print("Really wipe '%s'? [y/N]: " % build_root)
+                    choice = raw_input().lower()
+                    if choice != 'y':
+                        print('Aborting')
+                        sys.exit(0)
+            build_args = ['--root=' + build_root, '--noinit', '--shell']
+            if opts.wipe:
+                build_args.append('--wipe')
+            sys.exit(osc.build.run_build(*build_args))
+        elif subcmd in ('shell', 'chroot') or opts.shell:
+            print('--shell in combination with build-type %s is experimental.' % vm_chroot)
+            print('The semantics may change at any time!')
+            opts.shell = True
 
         args = self.parse_repoarchdescr(args, opts.noinit or opts.offline, opts.alternative_project, False, opts.vm_type, opts.multibuild_package)
 


nickbroon:
Fab, thanks @lethliel ! :-)

nickbroon:
@marcus-h @lethliel can this be merged and published?

lethliel:
I am working on another bug that is close to this one.
You cannot use `osc chroot -root=/var/tmp/build-root/openSUSE_Factory-x86_64 --noinit ` anymore.
It expects to be executed in the working copy. 

This is definitively a regression that needs to be fixed. As soon as I have a fix for that I will release.

lethliel:
@marcus-h can you please have a look. 
I needed to revive a little of the old code. We need to bypass build.py. Otherwise we will need to completely rewrite build.py

marcus-h:
On 2019-12-13 01:09:24 -0800, Marco Strigl wrote:
> @marcus-h can you please have a look.
> I needed to revive a little of the old code. We need to bypass build.py. Otherwise we will need to completely rewrite build.py
>
Hmm this would break the "new" use cases:
osc shell/chroot --noinit --vm-type kvm

Let's try to get this straight with a few refactorings. In essence, we should
end up with code in do_build that looks like this:

def do_build(...):
    ...
    args = self.parse_repoarchdescr(args, ...)
    if not opts.local_package:
        ...
    if subcmd in ('shell', 'chroot') or opts.shell:
        # opts.shell is debatable because it might do something special
        # in case of --extra-pkgs?! (see build.py (is this intended?))
        repo, arch, build_descr = args
        prj, pac, pacname = build.calculate_prj_pac_pacname(opts)
        build_root = build.calculate_build_root(apiurl, prj, pacname, repo,
                                                arch, opts)
        build_root, vm_opts = build.calculate_vm_options(opts)
        if opts.wipe and not opts.force:
            raw_input('Really wipe blah...')
            ...
        # the --noinit might be superfluous (I haven't checked the build code)
        extra_args = []
        if opts.wipe:
            extra_args.append('--wipe')
        build.run_build('--root', build_root, '--noinit', '--shell',
                        **extra_args)
        sys.exit(0)
    elif opts.wipe:
        raise ...  # not supported


Advantages:
- code reuse in commandline.py and build.py (the needed refactorings are
  mainly about moving existing code into new functions)
- no regressions wrt. the old "osc chroot" (in contrast, we even "fix"
  some bugs in the old "osc chroot" build_root calculation etc.)

Disadvantages:
- it might break the current "osc shell" implementation in case of a kiwi,
  docker, podman, or fissile build

Personally, I can perfectly live with the disadvantages because these are
"new" features (at least from the 0.166 release POV). Alternatively, we
simply route the "osc shell" command through the build.py as we do now (but
then "osc shell" != "osc chroot").

What do you guys think?

nickbroon:
I've not really had a change to play with the newer 'osc shell' can't really comment. But I'd definitely like to see the regressions in 'osc chroot' fixed, as it's impacting the tooling and CI that development engineers in my organisation are using.

lethliel:
@nickbroon @marcus-h 

I discussed it with @mlschroe and we will break the old implementation of osc shell. 
There will be a new switch --build-shell for the old implementation. 

I will work on this PR again today and I think we will have a fix by tomorrow afternoon.

lethliel:
@nickbroon I am working on this. If you need a fix immediately you can just downgrade to 0.166.4.

nickbroon:
I've asked all the development engineers in my organisation to disable automatic upgrades for osc/obs-build package from the https://build.opensuse.org/project/show/openSUSE:Tools repostory and/or downgrade to 0.166.4 manually. A lot of grumbles but it should un-stick most people.
Hopefully the fix will be published soon.

lethliel:
@marcus-h: Can you please have a look again. In addition with https://github.com/openSUSE/obs-build/pull/537 this will completely restore the old behavior.

marcus-h:
On 2019-12-17 05:37:04 -0800, Marco Strigl wrote:
> @marcus-h: Can you please have a look again. In addition with https://github.com/openSUSE/obs-build/pull/537 this will completely restore the old behavior.
>
I just had a very quick glance at it:

> index 8f599b85..a0c1c172 100644
> --- a/osc/build.py
> +++ b/osc/build.py
> @@ -524,6 +524,32 @@ def get_kiwipath_from_buildinfo(apiurl, bi_filename, prj, repo):
>      kiwipath.insert(0, myprp)
>      return kiwipath
>
> +def calculate_prj_pac(opts):
> +    project = opts.alternative_project or store_read_project('.')
> +    if opts.local_package:
> +        package = os.path.splitext(os.path.basename(descr))[0]

descr is not defined.

> +    else:
> +        package = store_read_package('.')
> +    return project, package
> +
> +def calculate_build_root(apihost, prj, pac, repo, arch, opts):
> +    buildroot = os.environ.get('OSC_BUILD_ROOT', conf.config['build-root']) \
> +                            % {'repo': repo, 'arch': arch, 'project': prj, 'package': pac, 'apihost': apihost}
> +    return buildroot
> +
> +def run_build(*args, **kwargs):
> +    cmd = [ config['build-cmd']]
> +    cmd.extend(*args)
> +

This should be cmd.extend(args) (in general, *args is wrong). It just "works"
because the caller in commandline is "wrong" (actually, it should be
"osc.build.run_build(*build_args)").

> +    sucmd = config['su-wrapper'].split()
> +    if sucmd[0] == 'su':
> +        if sucmd[-1] == '-c':
> +            sucmd.pop()
> +        cmd = sucmd + ['-s', cmd[0], 'root', '--' ] + cmd[1:]
> +    else:
> +        cmd = sucmd + cmd
> +    run_external(cmd[0], *cmd[1:])

Actually, the rationale behind these functions was that we reuse them
in the osc.build.main function, too (otherwise, we still "duplicate" code).


> index 78e66e02..8146b59d 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -6451,6 +6450,27 @@ def do_build(self, subcmd, opts, *args):
>                    opts.vm_type = lastbuildroot[2]
>
>          args = self.parse_repoarchdescr(args, opts.noinit or opts.offline, opts.alternative_project, False, opts.vm_type, opts.multibuild_package)

Hmmpf I just realized that this is too early because it would break
"osc chroot --root /path/to/build-root" in a non package wc (yes, my
previously reply was wrong:/).

> +        if subcmd in ('shell', 'chroot') or opts.shell or opts.wipe:
> +            if opts.root:
> +                build_root = opts.root
> +            else:
> +                repo, arch, build_descr = args
> +                prj, pac = osc.build.calculate_prj_pac(opts)
> +                apihost = urlsplit(self.get_api_url())[1]
> +                build_root = osc.build.calculate_build_root(apihost, prj, pac, repo,
> +                                                    arch, opts)
> +            if opts.wipe and not opts.force:
> +                    # Confirm delete
> +                    print("Really wipe '%s'? [y/N]: " % build_root)
> +                    choice = raw_input().lower()
> +                    if choice == 'n':
> +                        print('Aborting')
> +                        sys.exit(0)
> +            build_args = ['--root='+build_root, '--noinit', '--shell']
> +            if opts.wipe:
> +                build_args.append('--wipe')

Hmm that is, we drop the chroot/shell support for VMs, right? If so, what
about "simply" reverting the corresponding commits so that we end up with
the original "osc chroot" implementation again? Then, we can prepare a new
osc release and afterwards we can have a look at the shell/chroot rewrite
again (without "time pressure").

What do you think?

lethliel:
>> + package = os.path.splitext(os.path.basename(descr))[0]
> 
> descr is not defined.
>

fixed. I pass build_descr now into this function.

>> +def run_build(*args, **kwargs):
>> + cmd = [ config['build-cmd']]
>> + cmd.extend(*args)
>> +
> 
> This should be cmd.extend(args) (in general, *args is wrong). It just
> "works"
> because the caller in commandline is "wrong" (actually, it should be
> "osc.build.run_build(*build_args)").
> 

Yes. Fixed.

>> + sucmd = config['su-wrapper'].split()
>> + if sucmd[0] == 'su':
>> + if sucmd[-1] == '-c':
>> + sucmd.pop()
>> + cmd = sucmd + ['-s', cmd[0], 'root', '--' ] + cmd[1:]
>> + else:
>> + cmd = sucmd + cmd
>> + run_external(cmd[0], *cmd[1:])
> 
> Actually, the rationale behind these functions was that we reuse them
> in the osc.build.main function, too (otherwise, we still "duplicate" code).
> 

Yes. I know. But I would to these in a redesign of the build.py in a
separate script.

>> args = self.parse_repoarchdescr(args, opts.noinit or opts.offline,
> opts.alternative_project, False, opts.vm_type, opts.multibuild_package)
> 
> Hmmpf I just realized that this is too early because it would break
> "osc chroot --root /path/to/build-root" in a non package wc (yes, my
> previously reply was wrong:/).
> 

Yes. But It inside the else tree of the if opts.root case.
It is then needed again afterwards. (This is kind of a code duplication
but it is just one line)

>> + if subcmd in ('shell', 'chroot') or opts.shell or opts.wipe:
>> + if opts.root:
>> + build_root = opts.root
>> + else:
>> + repo, arch, build_descr = args
>> + prj, pac = osc.build.calculate_prj_pac(opts)
>> + apihost = urlsplit(self.get_api_url())[1]
>> + build_root = osc.build.calculate_build_root(apihost, prj, pac, repo,
>> + arch, opts)
>> + if opts.wipe and not opts.force:
>> + # Confirm delete
>> + print("Really wipe '%s'? [y/N]: " % build_root)
>> + choice = raw_input().lower()
>> + if choice == 'n':
>> + print('Aborting')
>> + sys.exit(0)
>> + build_args = ['--root='+build_root, '--noinit', '--shell']
>> + if opts.wipe:
>> + build_args.append('--wipe')
> 
> Hmm that is, we drop the chroot/shell support for VMs, right? If so, what
> about "simply" reverting the corresponding commits so that we end up with
> the original "osc chroot" implementation again? Then, we can prepare a new
> osc release and afterwards we can have a look at the shell/chroot rewrite
> again (without "time pressure").

I added the else tree where we set opts.shell if chroot / shell is given
and a vm_type is set.

Then the build script will take care of the setup of the vm and will
provide a shell.

> 
> What do you think?
> 

I really don't like the idea of reverting everything. As of now we are
clean with this PR. And the rest (like build.py code deduplication and
implementetion of the build inline shell) will be implemented afterwards.

marcus-h:
On 2019-12-19 02:50:55 -0800, Marco Strigl wrote:

<SNIP>

> > Hmm that is, we drop the chroot/shell support for VMs, right? If so, what
> > about "simply" reverting the corresponding commits so that we end up with
> > the original "osc chroot" implementation again? Then, we can prepare a new
> > osc release and afterwards we can have a look at the shell/chroot rewrite
> > again (without "time pressure").
>
> I added the else tree where we set opts.shell if chroot / shell is given
> and a vm_type is set.
>
> Then the build script will take care of the setup of the vm and will
> provide a shell.
>
We should also take "build-type" config option into account.


> > What do you think?
> >
>
> I really don't like the idea of reverting everything. As of now we are
> clean with this PR.

IMHO we are "far" from being clean:/ (there are some inconsistencies in
the vm and non-vm use case). Also, in the vm code path several
"strange"/unexpected things can happen (expansion errors, missing packages
etc.).

What about the following:
- let's "fix" the "osc chroot" implementation (that's what your patch
  basically does (I'll need to have a second look at it, though))
- mark the --shell option for vm builds as experimental and inform the user
  that the semantics might change in future osc releases (just some
  print(...) call); then we do not have to worry about regressions when we
  touch the --shell vm codepath in the future.

lethliel:
<SNIP>


> What about the following:
> - let's "fix" the "osc chroot" implementation (that's what your patch
> basically does (I'll need to have a second look at it, though))
> - mark the --shell option for vm builds as experimental and inform the user
> that the semantics might change in future osc releases (just some
> print(...) call); then we do not have to worry about regressions when we
> touch the --shell vm codepath in the future.
> 

That sounds good to me. I now consider also the
conf.config('build-type') entry and print a warning if --shell / chroot
in combination
with a VM is used.

marcus-h:
On 2019-12-19 11:54:42 -0800, Marco Strigl wrote:
> index 8f599b85..2ddbf6ef 100644
> --- a/osc/build.py
> +++ b/osc/build.py

> +def calculate_build_root(apihost, prj, pac, repo, arch, opts):

opts can be dropped (unused).

> +    buildroot = os.environ.get('OSC_BUILD_ROOT', conf.config['build-root']) \
> +                            % {'repo': repo, 'arch': arch, 'project': prj, 'package': pac, 'apihost': apihost}
> +    return buildroot
> +
> +def run_build(*args, **kwargs):

kwargs can be dropped (unused).

> +    cmd = [ config['build-cmd']]
> +    cmd += args
> +
> +    sucmd = config['su-wrapper'].split()

We should also take the OSC_SU_WRAPPER env variable into account.

sucmd = os.environ.get('OSC_SU_WRAPPER', config['su-wrapper']).split()

(The access to the config dict is a bit inconsistent "config" vs.
"conf.config" (see calculate_build_root).)

> +    if sucmd[0] == 'su':
> +        if sucmd[-1] == '-c':
> +            sucmd.pop()
> +        cmd = sucmd + ['-s', cmd[0], 'root', '--' ] + cmd[1:]
> +    else:
> +        cmd = sucmd + cmd
> +    run_external(cmd[0], *cmd[1:])
> +

We should return the retval:
return run_external(cmd[0], *cmd[1:])

(or return run_external(*cmd)).

> @@ -696,7 +722,7 @@ def main(apiurl, opts, argv):
>
>      if opts.shell:
>          buildargs.append("--shell")
> -        if os.path.exists(build_root) and os.path.exists(bi_filename) and not opts.clean and not opts.extra_pkgs:
> +        if os.path.exists(build_root) and os.path.exists(bi_filename) and not opts.clean and not opts.extra_pkgs and not opts.vm_type:

Hmm why?

>              opts.noinit = True
>              opts.offline = True
>              # we should check if the service did run before and only skip it then,
> index 78e66e02..78037e14 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -6341,6 +6341,8 @@ def parse_repoarchdescr(self, args, noinit = False, alternative_project = None,
>                    help=SUPPRESS_HELP)
>      @cmdln.option('--shell-cmd', metavar='COMMAND',
>                    help='run specified command instead of bash')
> +    @cmdln.option('-f', '--force', action='store_true',
> +                  help='Do not ask confirmation for wipe')

(Hmm maybe: "... for confirmation ..."?)

>      @cmdln.option('--host', metavar='HOST',
>              help='perform the build on a remote server - user@server:~/remote/directory')
>      @cmdln.option('--trust-all-projects', action='store_true',
> @@ -6425,9 +6427,6 @@ def do_build(self, subcmd, opts, *args):
>          if opts.debuginfo and opts.disable_debuginfo:
>              raise oscerr.WrongOptions('osc: --debuginfo and --disable-debuginfo are mutual exclusive')
>
> -        if subcmd == 'chroot' or subcmd == 'shell':
> -            opts.shell = True
> -
>          if subcmd == 'wipe':
>              opts.wipe = True
>
> @@ -6450,6 +6449,36 @@ def do_build(self, subcmd, opts, *args):
>                 if not opts.vm_type:
>                    opts.vm_type = lastbuildroot[2]
>
> +        vm_chroot = None
> +        if opts.vm_type or conf.config['build-type']:
> +            vm_chroot = 1

Hmm could be simplified to

vm_chroot = opts.vm_type or conf.config['build-type']

(or maybe "vm_type" or "build_type" instead of "vm_chroot".)

> +        if (subcmd in ('shell', 'chroot') or opts.shell or opts.wipe) and (not vm_chroot):

Superfluous parentheses (around vm_chroot).

> +            if opts.root:
> +                build_root = opts.root
> +            else:
> +                args = self.parse_repoarchdescr(args, opts.noinit or opts.offline, opts.alternative_project, False, opts.vm_type, opts.multibuild_package)
> +                repo, arch, build_descr = args
> +                prj, pac = osc.build.calculate_prj_pac(opts, build_descr)
> +                apihost = urlsplit(self.get_api_url())[1]
> +                build_root = osc.build.calculate_build_root(apihost, prj, pac, repo,
> +                                                    arch, opts)

opts should be dropped.

> +            if opts.wipe and not opts.force:
> +                    # Confirm delete
> +                    print("Really wipe '%s'? [y/N]: " % build_root)
> +                    choice = raw_input().lower()
> +                    if choice == 'n':

Hmm maybe

if choice != 'y':
    ...

(Better safe than sorry.)

> +                        print('Aborting')
> +                        sys.exit(0)
> +            build_args = ['--root='+build_root, '--noinit', '--shell']

Hmm missing whitespaces.

> +            if opts.wipe:
> +                build_args.append('--wipe')
> +            osc.build.run_build(*build_args)

sys.exit(osc.build.run_build(*build_args)) in order to propagate the retval to
the caller (assumption: the build script also returns the correct retval).

> +            sys.exit(0)
> +        elif (subcmd in ('shell', 'chroot') or opts.shell) and (vm_chroot):
> +            print('--shell in combination with build-type kvm/xen is experimental.')
> +            print('The semantic and the behavior may change at any time!')

s/semantic/semantics/

(Hmm using either "semantics" or "behavior" should be sufficient:) )

> +            opts.shell=True

Missing whitespaces around "=" :)

Otherwise, LGTM!

lethliel:
<SNIP>

> On 2019-12-19 11:54:42 -0800, Marco Strigl wrote:
>> - if os.path.exists(build_root) and os.path.exists(bi_filename) and
> not opts.clean and not opts.extra_pkgs:
>> + if os.path.exists(build_root) and os.path.exists(bi_filename) and
> not opts.clean and not opts.extra_pkgs and not opts.vm_type:
> 
> Hmm why?

Otherwise opts.noinit will be set to true even in VM case.
And a VM always needs to be initialized. Otherwise the shell
call will not work with a VM.

> 
>> opts.noinit = True
>> opts.offline = True

<SNIP>

All other requested changes done.

marcus-h:
On 2019-12-20 07:03:06 -0800, Marco Strigl wrote:
> > On 2019-12-19 11:54:42 -0800, Marco Strigl wrote:
> >> - if os.path.exists(build_root) and os.path.exists(bi_filename) and
> > not opts.clean and not opts.extra_pkgs:
> >> + if os.path.exists(build_root) and os.path.exists(bi_filename) and
> > not opts.clean and not opts.extra_pkgs and not opts.vm_type:
> >
> > Hmm why?
>
> Otherwise opts.noinit will be set to true even in VM case.
> And a VM always needs to be initialized. Otherwise the shell
> call will not work with a VM.
>
Hmm then we can remove the whole if-block because this codepath is only
executed if opts.vm_type (or conf.config['build-type']) is set.


> index 8f599b85..edcbca5f 100644
> --- a/osc/build.py
> +++ b/osc/build.py

> +def run_build(*args):
> +    cmd = [ config['build-cmd']]

Minor: style (remove leading whitespace)

> +    cmd += args
> +
> +    sucmd = os.environ.get('OSC_SU_WRAPPER', config['su-wrapper']).split()
> +    if sucmd[0] == 'su':
> +        if sucmd[-1] == '-c':
> +            sucmd.pop()
> +        cmd = sucmd + ['-s', cmd[0], 'root', '--' ] + cmd[1:]

Minor: style (remove trailing whitespace (I know, the old code has it,
too:/ ))


> index 78e66e02..c881625b 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py

> @@ -6450,6 +6449,34 @@ def do_build(self, subcmd, opts, *args):
>                 if not opts.vm_type:
>                    opts.vm_type = lastbuildroot[2]
>
> +        vm_chroot = None

Should be removed.

Feel free to ignore the style remarks:)

marcus-h:
Thanks! Some final remarks:)

On 2019-12-20 10:53:06 -0800, Marco Strigl wrote:
> @@ -696,12 +722,6 @@ def main(apiurl, opts, argv):
>
>      if opts.shell:
>          buildargs.append("--shell")
> -        if os.path.exists(build_root) and os.path.exists(bi_filename) and not opts.clean and not opts.extra_pkgs:
> -            opts.noinit = True
> -            opts.offline = True
> -            # we should check if the service did run before and only skip it then,
> -            # but we have no save point for this atm
> -            opts.noservice = True
>
Hmm I have to ask again: why?:)
In case of a "vm chroot", the stuff makes sense (at least from a quick glance
(but maybe I miss something)). Is there a specific reason/bug why we drop this?

(That is, the condition _without_ the "and not opts.vm_type" might make sense.)

>      if opts.noinit:
>          buildargs.append('--noinit')
> index 78e66e02..a62a35f5 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -6341,6 +6341,8 @@ def parse_repoarchdescr(self, args, noinit = False, alternative_project = None,
>                    help=SUPPRESS_HELP)
>      @cmdln.option('--shell-cmd', metavar='COMMAND',
>                    help='run specified command instead of bash')
> +    @cmdln.option('-f', '--force', action='store_true',
> +                  help='Do not ask for confirmation to wipe')
>      @cmdln.option('--host', metavar='HOST',
>              help='perform the build on a remote server - user@server:~/remote/directory')
>      @cmdln.option('--trust-all-projects', action='store_true',
> @@ -6425,9 +6427,6 @@ def do_build(self, subcmd, opts, *args):
>          if opts.debuginfo and opts.disable_debuginfo:
>              raise oscerr.WrongOptions('osc: --debuginfo and --disable-debuginfo are mutual exclusive')
>
> -        if subcmd == 'chroot' or subcmd == 'shell':
> -            opts.shell = True
> -
>          if subcmd == 'wipe':
>              opts.wipe = True
>
> @@ -6450,6 +6449,33 @@ def do_build(self, subcmd, opts, *args):
>                 if not opts.vm_type:
>                    opts.vm_type = lastbuildroot[2]
>
> +        vm_chroot = opts.vm_type or conf.config['build-type']
> +        if (subcmd in ('shell', 'chroot') or opts.shell or opts.wipe) and not vm_chroot:
> +            if opts.root:
> +                build_root = opts.root
> +            else:
> +                args = self.parse_repoarchdescr(args, opts.noinit or opts.offline, opts.alternative_project, False, opts.vm_type, opts.multibuild_package)
> +                repo, arch, build_descr = args
> +                prj, pac = osc.build.calculate_prj_pac(opts, build_descr)
> +                apihost = urlsplit(self.get_api_url())[1]
> +                build_root = osc.build.calculate_build_root(apihost, prj, pac, repo,
> +                                                    arch)
> +            if opts.wipe and not opts.force:
> +                    # Confirm delete
> +                    print("Really wipe '%s'? [y/N]: " % build_root)
> +                    choice = raw_input().lower()
> +                    if choice != 'y':
> +                        print('Aborting')
> +                        sys.exit(0)
> +            build_args = ['--root=' + build_root, '--noinit', '--shell']
> +            if opts.wipe:
> +                build_args.append('--wipe')
> +            sys.exit(osc.build.run_build(*build_args))
> +        elif (subcmd in ('shell', 'chroot') or opts.shell) and vm_chroot:

Minor: "and vm_chroot" can be dropped.

> +            print('--shell in combination with build-type kvm/xen is experimental.')

Minor: name the specific type

print('--shell in combination with build-type %s is experimental.' % vm_chroot)

> +            print('The semantics may change at any time!')
> +            opts.shell = True
> +

lethliel:
<SNIP>

>> - if os.path.exists(build_root) and os.path.exists(bi_filename) and
> not opts.clean and not opts.extra_pkgs:
>> - opts.noinit = True
>> - opts.offline = True
>> - # we should check if the service did run before and only skip it then,
>> - # but we have no save point for this atm
>> - opts.noservice = True
>>
> Hmm I have to ask again: why?:)
> In case of a "vm chroot", the stuff makes sense (at least from a quick
> glance
> (but maybe I miss something)). Is there a specific reason/bug why we
> drop this?
> 
> (That is, the condition _without_ the "and not opts.vm_type" might make
> sense.)
> 

Because the if block is only reached if vm_chroot is set. A noinit makes
only sense when using a local chroot.

So this breaks osc build --shell --vm-type=kvm if a local buildroot
already exists.

<SNIP>

All other remarks have been changed.

marcus-h:
On 2019-12-20 11:46:11 -0800, Marco Strigl wrote:
> >> - if os.path.exists(build_root) and os.path.exists(bi_filename) and
> > not opts.clean and not opts.extra_pkgs:
> >> - opts.noinit = True
> >> - opts.offline = True
> >> - # we should check if the service did run before and only skip it then,
> >> - # but we have no save point for this atm
> >> - opts.noservice = True
> >>
> > Hmm I have to ask again: why?:)
> > In case of a "vm chroot", the stuff makes sense (at least from a quick
> > glance
> > (but maybe I miss something)). Is there a specific reason/bug why we
> > drop this?
> >
> > (That is, the condition _without_ the "and not opts.vm_type" might make
> > sense.)
> >
>
> Because the if block is only reached if vm_chroot is set. A noinit makes
> only sense when using a local chroot.
>
> So this breaks osc build --shell --vm-type=kvm if a local buildroot
> already exists.
>
Ok.

> <SNIP>
>
> All other remarks have been changed.
>
Merged. Thanks a lot!

PR #715, opened 2020-01-07
Remove unused function dgst_from_string

dcermak:
The function `dgst_from_string` does not appear to be used anywhere in the code base and existed only for compatibility reasons when Python 2.4 still was a thing.


marcus-h:
On 2020-01-07 07:35:00 -0800, Dan Čermák wrote:
> The function `dgst_from_string` does not appear to be used anywhere in the code base and existed only for compatibility reasons when Python 2.4 still was a thing.

Merged. Actually, this breaks the API but that's IMHO ok (users of this,
if at all, should be very rare)...

Thanks for the PR!

PR #718, opened 2020-01-09
- support zstd arch linux files in local build

adrianschroeter:
Note: This requires a tar executable supporting zstd


lethliel:
LGTM

PR #723, opened 2020-01-20
improve SSLError message

lethliel:
Improve the SSLError message if a version mismatch (TLSv1) is
detected and give the user a hint what is wrong.

(Basically the API does not supprt TLS <= 1.2 and the python
version does not support TLS >= 1.2)

### DIFF
--- a/osc/babysitter.py
+++ b/osc/babysitter.py
@@ -172,6 +172,8 @@ def run(prg, argv=None):
     except RPMError as e:
         print(e, file=sys.stderr)
     except SSLError as e:
+        if 'tlsv1' in str(e):
+            print('The python on this system does not support TLSv1.2', file=sys.stderr)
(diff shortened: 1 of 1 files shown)


osc/babysitter.py
```diff
@@ -172,6 +172,10 @@ def run(prg, argv=None):
     except RPMError as e:
         print(e, file=sys.stderr)
     except SSLError as e:
+        if 'tlsv1' in str(e):
+            print('The API does not allow SSL Versions < TLSv1.2 and it\n' \
+                'ssems that you are running on a system with a python\n' \
```
adrianschroeter: typo: seems

But I would not make it so vague. Just say that the system python version is not supporting TLSv1.2. 

In case the detection is wrong, it simply a bug ;)

lethliel:
updated

Issue #194, opened 2016-04-08
FR: osc sr should output URL to created request


aspiers:
Currently it just outputs something like

```
created request id 386220
```

but it would be much more user-friendly if it also outputted the URL:

```
https://build.opensuse.org/request/show/386220
```

so that the submitter can immediately visit it in the browser.

marcus-h:
On 2016-04-08 03:04:13 -0700, Adam Spiers wrote:
> Currently it just outputs something like
> 
>     created request id 386220
> 
> but it would be much more user-friendly if it also outputted the URL:
> 
>     https://build.opensuse.org/request/show/386220
> 
Depends on the user;) it wouldn't really fit to my workflow, because
I usually continue with processing the reqid on the cmdline (so I could
easily "copy" by just double clicking on it - with your proposed change
I would need to "carefully" select reqid;) ) (but that's a lame excuse
for rejecting this...).
From a technical POV, it should be doable (we can get the url via the
/configuration route).

> so that the submitter can immediately visit it in the browser.
> 
Out of curiosity: why do you want to do this?:)

adrianschroeter:
Just want to support Marcus and say that I use also the double click to get the number. We could make this an optional feature of course ....

aspiers:
Marcus Hüwe wrote:
> Depends on the user;)

True :-)

> it wouldn't really fit to my workflow, because
> I usually continue with processing the reqid on the cmdline (so I could
> easily "copy" by just double clicking on it - with your proposed change
> I would need to "carefully" select reqid;) )

Not really, because I'm not proposing to remove the existing line of
output, merely to add a new one.

> From a technical POV, it should be doable (we can get the url via the
> /configuration route).
> 
> > so that the submitter can immediately visit it in the browser.
> > 
> Out of curiosity: why do you want to do this?:)

One or more of the following:

-   To keep an eye on the builds if for some reason I prefer to do that
    in the browser than via "osc r -vw".
-   To double-check I got everything right (it's easier to visually
    review in the web-rendered UI than "osc linkdiff" or similar)
-   To browse around the source or destination package/project
-   To share the URL with someone else
-   To cross-reference the request from another system, e.g. via
    https://github.com/aspiers/desktop-config/blob/master/lib/browser/userscripts/page-id-helpers/README.md

I'm sure there are other use cases.

aspiers:
[...]

PR #726, opened 2020-01-23
print web url links for creating requests

lethliel:
This will print the direct url to show the created request.
New general bool option 'print_web_links' must be set to enable
this.

Right now this is only for creating requests. More to follow.

fixes https://github.com/openSUSE/osc/issues/194

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -1460,7 +1460,15 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                                        dst_project, dst_package,
                                        opts.message, orev=rev,
 				       src_update=src_update, dst_updatelink=opts.update_link)
+
         print('created request id', result)
+        if conf.config['print_web_links']:
+            root = ET.fromstring(b''.join(show_configuration(apiurl)))
+            node = root.find('obs_url')
+            if node is None or not node.text:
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2020-01-23 04:50:40 -0800, Marco Strigl wrote:
> index a339f918..4e37f50a 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -1460,7 +1460,10 @@ def _check_service(root):
>                                         dst_project, dst_package,
>                                         opts.message, orev=rev,
>  				       src_update=src_update, dst_updatelink=opts.update_link)
> +
>          print('created request id', result)
> +        if conf.config['print_web_links']:
> +            print('%s/request/show/%s' % (apiurl, result))
>
Hmm we should probably use the "obs_url" from the /public/configuration
route. (For instance, if we use the apiurl, the user has to login.)

Btw., in the future, it might make sense to delegate the url construction
to some object/class:

print(WebuiURLMapper(apiurl).request_show(reqid))
or
print(WebuiUrlFor(apiurl).request_show(reqid))
or
print(WebuiUrlFor.request_show(apiurl, reqid))
or
...

(Of course, for this PR, the manual construction is fine:) )

Issue #132, opened 2014-12-22
feature: osc maintained <package> -V return the version of the maintained package


vpereira:
if i want to get all maintained codestreams for a package i do:

`osc maintained <package>`

but then if i want to check the versions, I have to fire a `osc search <package> -V`

would be great to have a `osc maintained <package> -V`
that could answer with something similar:

```
openSUSE:12.3:Update/package (1.2.3)
openSUSE:13.1:Update/package (1.5.6)
openSUSE:13.2:Update/ppackage (x.y.z)
```

or at least a way to reduce the "search" scope just for maintained codestreams (however we would have to bring back the --maintained option). It could be called like:

```
osc search --maintained <package> -V 
```

is that possible or do we have already a way to achieve it?

adrianschroeter:
it is possible to implement that, but it is more then a one liner. Since we need to follow links and ask the backend. Or we start with tracking versions in the DB

ketheriel:
I can implement this next week and use some time in the weekend to check the current API. Is this to implement using the suggested syntax or is there any other way that is prefered?

vpereira:
I still think that its a good syntax

ketheriel:
I am going to take a closer look to the current API and check if we can retrieve package version without having to mess on the obs backend side. 
I don't see issues with implementing 'osc maintained -V <package>'.

PR #727, opened 2020-01-23
print verion in osc maintained if --version is given

lethliel:
Determine the version of the package for maintained project
and print version or '-' if version unknown.

fixes https://github.com/openSUSE/osc/issues/132

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -3324,6 +3324,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
     @cmdln.option('--nodevelproject', action='store_true',
                         help='do not follow a defined devel project ' \
                              '(primary project where a package is developed)')
+    @cmdln.option('--version', action='store_true',
+                        help='print version of maintained package')
     @cmdln.alias('sm')
     @cmdln.alias('maintained')
     def do_mbranch(self, subcmd, opts, *args):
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2020-01-23 05:39:58 -0800, Marco Strigl wrote:
> index a339f918..487418d7 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -3324,6 +3324,8 @@ def do_maintenancerequest(self, subcmd, opts, *args):
>      @cmdln.option('--nodevelproject', action='store_true',
>                          help='do not follow a defined devel project ' \
>                               '(primary project where a package is developed)')
> +    @cmdln.option('-v', '--version', action='store_true',
> +                        help='print version of maintained package')

Hmm what about omitting the short option "-v"? (just to avoid
confusion/ambiguities with a potential "--verbose" option).

> @@ -3381,6 +3383,12 @@ def do_mbranch(self, subcmd, opts, *args):
>          if opts.dryrun:
>              for r in result.findall('package'):
>                  line="%s/%s"%(r.get('project'), r.get('package'))
> +                if opts.version:
> +                    sr = get_source_rev(apiurl, r.get('project'), r.get('package'))
> +                    version = sr.get('version')
> +                    if not version or version == 'unknown':
> +                        version = '-'
> +                    line = line + (' (%s)' % version)

Ideas:
- use "unknown" instead of "-"
- use "(version: x.y.z)" or "(version: unknown)"

But I'm also fine with your current implementation:)

lethliel:
Yes. The -v option is a bit misleading. Changed it.

PR #729, opened 2020-01-24
fix decoding in interactive request mode

lethliel:
In interactive review mode:

If a diff is issued and the request is accepted with 'a -m ok'
the tmpfile with the diff will be read. This tmpfile.read() call
is now decoded properly.


lethliel:
simple fix

PR #731, opened 2020-01-28
Rename SafeConfigParser to ConfigParser

sbahling:
configparser.SafeConfigParser has been changed to configparser.ConfigParser
upstream in Python 3.2. Warnings are saying that the alias will be
removed in future versions.

### DIFF
--- a/osc/OscConfigParser.py
+++ b/osc/OscConfigParser.py
@@ -19,9 +19,11 @@ import sys
 
 if sys.version_info >= ( 3, ):
     import configparser
+    ConfigParser = configparser.ConfigParser
 else:
     #python 2.x
     import ConfigParser as configparser
+    ConfigParser = configparser.SafeConfigParser
 
 import re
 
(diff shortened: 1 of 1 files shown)


sbahling:
I'm not sure if we still maintain Python2 compatibility with latest osc releases, but I guess this change would break that.

marcus-h:
On 2020-01-28 09:30:49 -0800, Scott Bahling wrote:
> I'm not sure if we still maintain Python2 compatibility with latest osc releases, but I guess this change would break that.
>
At the top of the file, we already have the python2/python3 specific imports.
We could do something like this:

diff --git a/osc/OscConfigParser.py b/osc/OscConfigParser.py
index c2aaf74..80aa0eb 100644
--- a/osc/OscConfigParser.py
+++ b/osc/OscConfigParser.py
@@ -19,9 +19,11 @@ import sys

 if sys.version_info >= ( 3, ):
     import configparser
+    ConfigParser = configparser.ConfigParser
 else:
     #python 2.x
     import ConfigParser as configparser
+    ConfigParser = configparser.SafeConfigParser

 import re


and then simply use "ConfigParser" instead of "configparser.ConfigParser".

marcus-h:
Can you squash both commits into one?

marcus-h:
Merged. Thanks a lot!

Btw., I just saw your change (the squash) by accident - in the future, it
would be cool if you could add a short comment to the PR so that we get a
notification:)

PR #732, opened 2020-01-30
- allow "osc r --vertical" for projects

adrianschroeter:
dunno about others, but the default horizontal listing is not really
usable. This is at least true if you have more packages than
repostories.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,6 @@
 0.168.0
   - MR creation honors orev now (bsc#1160446)
+  - allow "osc r --vertical" for projects
 
 0.167.2
   - fix regression in osc chroot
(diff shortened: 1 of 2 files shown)


adrianschroeter:
hm, we could also do horizontal/vertical dependending on the number of packages to repos.

It may be a bit surprising that this is not fixed, but may help in practice ....

Any opinion about that @lethliel @marcus-h ?

marcus-h:
On 2020-01-30 05:47:30 -0800, Adrian Schröter wrote:
> hm, we could also do horizontal/vertical dependending on the number of packages to repos.
>
> It may be a bit surprising that this is not fixed, but may help in practice ....
>
> Any opinion about that @lethliel @marcus-h ?
>
Hmm maybe we could print a hint to use "--vertical"
if #packages >> #repos?

For now, I merged your PR. Thanks!

PR #734, opened 2020-02-07
raise an error when one tries to remove a repository just for a package

adrianschroeter:


marcus-h:
On 2020-02-07 02:34:08 -0800, Adrian Schröter wrote:
> index 44694999..e64f5f51 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -2014,11 +2014,12 @@ def do_deleterequest(self, subcmd, opts, *args):
>          else:
>              raise oscerr.WrongArgs('Please specify at least a project.')
>
> -        if not opts.all and package is None and not opts.repository:
> -            raise oscerr.WrongOptions('No package name has been provided. Use --all option, if you want to request to delete the entire project.')
> -
>          if opts.repository:
>              repository = opts.repository
> +            if package:
> +               raise oscerr.WrongOptions('Repositories can only be removed for the entire project.')
> +        elif not opts.all and package is None:
> +            raise oscerr.WrongOptions('No package name has been provided. Use --all option, if you want to request to delete the entire project.')
>
LGTM. Maybe we should add code for the other "error" cases, too.
IMHO, the following combinations are "illegal":
- package is not None and (opts.all or opts.repository)
- opts.all and (package is not None or opts.repository)
- opts.repository and (opts.all or package is not None)

dmach:
It seems this is identical to a commit that is part of master already: c06b794ddf6e48c8a02db0d10e424fd4b2286e8b and that was merged in:
- https://github.com/openSUSE/osc/pull/898

Closing.

PR #736, opened 2020-02-07
improve speed of bash completion

lethliel:
Command substitution should be done by compgen which is much faster.

builtin compgen -W "${list[*]}" -- "${cur}" | sed -r "s@^${colon_word}@@g"
--> 1 minute 20 seconds

builtin compgen -W '`cat ${projects}`' -- "${cur}" | sed -r "s@^${colon_word}@@g"
--> 4 seconds

### DIFF
--- a/dist/osc.complete
+++ b/dist/osc.complete
@@ -160,15 +160,14 @@ projects ()
     done
     shift $argc
     cur="$1"
-    if test -n "${cur}" ; then
-	list=($(command grep -E "^${cur}" ${projects}))
-    else
-	list=($(command cat ${projects}))
-    fi
     if ((colon)) ; then
 	local colon_word
 	colon_word=${cur%${cur##*:}}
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2020-02-07 04:25:08 -0800, Marco Strigl wrote:
> Command substitution should be done by compgen which is much faster.
>
> builtin compgen -W &quot;${list[*]}&quot; -- &quot;${cur}&quot; | sed -r &quot;s@^${colon_word}@@g&quot;
> --&gt; 1 minute 20 seconds
>
> builtin compgen -W &#39;`cat ${projects}`&#39; -- &quot;${cur}&quot; | sed -r &quot;s@^${colon_word}@@g&quot;
> --&gt; 4 seconds

I cannot really comment on this PR (as I have no clue about the osc.complete
script...). If you think its fine, go ahead and merge it:)

PS: it might make sense to include the speed "comparison" into the commit
message as well.

PR #737, opened 2020-02-09
Allow arguments deduction for getbinaries command

matwey:
When 4-argument type of getbinaries command is used, there may be a case when
REPOSITORY ARCHITECTURE pair is unique. It is quite a case for JeOS image
packages.

Allow deducing repository and architecture when only project and package are
specified.

Signed-off-by: Matwey V. Kornilov <matwey.kornilov@gmail.com>

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -7267,16 +7267,35 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         elif len(args) == 3:
             project, repository, architecture = args
         elif len(args) >= 1 and len(args) <= 2:
-            if is_package_dir(os.getcwd()):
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2020-02-09 02:55:32 -0800, Matwey V. Kornilov wrote:
> When 4-argument type of getbinaries command is used, there may be a case when
> REPOSITORY ARCHITECTURE pair is unique. It is quite a case for JeOS image
> packages.
>
Hmm we could do that but this is a very specific/"exotic" use case (IMHO).
My "fear" is that we have to/should add this behavior to other commands,
which require a repo arch pair, too...

@lethliel/@adrianschroeter: what do you think?

PR #739, opened 2020-02-13
Fixed problems in automated deployment via Travis CI

### LINKED ISSUE #658: Publish `osc` on PyPI?
Hi there,

I was wondering, if we should publish `osc` also as a package on PyPI?

### LINKED ISSUE #708: List the Python dependencies
Hi there,

while working on a derived package I noticed that `osc` actually has some dependencies, e.g. `M2Crypto`. These dependencies should me mentioned in a `requirements.txt` and/or as `install_requires` in `setup.py`.

Having this dependencies defined outside OBS would also be helpful for #658.

crazyscientist:
* Reformatted `.travis.yml`
* Reformatted `setup.py`
* Used `before_*` statements instead of script chains
* Changed `setup.py` to build a proper Python package for PyPI
* fixes #658
* fixes #708

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,18 @@
 language: python
 python:
-- '2.7'
-- '3.6'
+  - '2.7'
+  - '3.6'
+  - '3.7'
 addons:
   apt:
     packages:
-    - diffstat
+      - diffstat
 sudo: false
-script: cd tests; python suite.py
+before_script: cd $TRAVIS_BUILD_DIR/tests
+script: python suite.py
+before_deploy:
+  - cd $TRAVIS_BUILD_DIR/
+  - rm -rf $TRAVIS_BUILD_DIR/dist/*
 deploy:
   provider: pypi
   user: suse
@@ -15,5 +20,5 @@ deploy:
     secure: R4+YNPW2tsiY06hibGvONYn0//1z1QdcY8VmNbYpIRly4eTAbPE9uejKpyuflUkznpEkoqCdFzi5FNFhgat9N+AkIKyX9NTkf0oxaKKbdqBM7H1V8bqLYlAO479262spRyO0ee5fV5v6g81AFjncIV+pGjtQ0Vg/sjVcvGa61bs=
   on:
     tags: true
-    distributions: sdist bdist_wheel
+    distributions: sdist
     repo: openSUSE/osc
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,4 @@
+include NEWS
+include README
+include AUTHORS
+include COPYING
--- a/setup.py
+++ b/setup.py
@@ -3,17 +3,19 @@
 from distutils.core import setup
 import distutils.core
 from distutils.command import build, install_data
+import gzip
 import os.path
-import osc.core
 import sys
 
 import setuptools
 
+import osc.core
 from osc import commandline
-from osc import babysitter
+
 # optional support for py2exe
 try:
     import py2exe
+
     HAVE_PY2EXE = True
 except:
     HAVE_PY2EXE = False
@@ -27,14 +29,13 @@ class build_osc(build.build, object):
     def build_man_page(self):
         """
         """
-        import gzip
         man_path = os.path.join(self.build_base, 'osc.1.gz')
         distutils.log.info('generating %s' % man_path)
         outfile = gzip.open(man_path, 'wt')
         osccli = commandline.Osc(stdout=outfile)
         # FIXME: we cannot call the main method because osc expects an ~/.oscrc
         # file (this would break builds in environments like the obs)
-        #osccli.main(argv = ['osc','man'])
+        # osccli.main(argv = ['osc','man'])
         osccli.optparser = osccli.get_optparser()
         osccli.do_man(None)
         outfile.close()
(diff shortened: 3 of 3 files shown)


crazyscientist:
This fixes #658 :smiley:

marcus-h:
First of all, thanks for the PR! Can you squash some of the
commits? For instance, PATCH 6 reverts parts that were introduced
in PATCH 2, PATCH 3, and PATCH 5. This would simplify the review
a lot:)

Aside: do you really insist on all the whitespace/style changes? Of course,
the new style is (probably) preferred and makes tools like flake8/pep8
happy but in this case it just generates some "patch noise" (but since,
the patch is small, we could cope with that... (although, such style
changes should be done in a separate patch)).

crazyscientist:
@marcus-h Pardon me asking, but I have serious problems understanding your *problem* regarding squashing and reviewing. Would you like to know why?

* GitHub offers a function called "Squash and Merge", which automatically squashes the commits prior to merging with just one click. This is really reliable and comfortable; you should try it :wink: 
* As we are in agreement that PRs should be squashed prior to merging, your approach to look at the single commits seems to be suboptimal. Why not just look at the overall diff?

crazyscientist:
@marcus-h Would you mind having a look at my questions so this PR might get merged someday?

marcus-h:
On 2020-02-14 00:21:06 -0800, Andreas Hasenkopf wrote:
> @marcus-h Pardon me asking, but I have serious problems understanding your *problem* regarding squashing and reviewing. Would you like to know why?
>
> * GitHub offers a function called "Squash and Merge", which automatically squashes the commits prior to merging with just one click. This is really reliable and comfortable; you should try it :wink:

I won't do that for at least two reasons:
- I do not modify the commits, including the commit message, from other people
  (rationale: I could introduce a bug and someone else gets blamed)
- I do not use the github webui

> * As we are in agreement that PRs should be squashed prior to merging, your approach to look at the single commits seems to be suboptimal.

Well, that's how my review process works...
Pseudocode (a bit simplified;) ):

For each commit:
    - Read the commit message
    - Think about it
    - Read the code and check if the code "corresponds" to the commit
      message
    - Think about the code

> Why not just look at the overall diff?
>
I prefer to review (small) logical chunks (that is, commits) instead of
one big diff.

In this case, even if I look at the diff, it is not apparent what you
are fixing and why the changes "fix" "the"/a problem:

- The title of the PR is: "Fixed problems in automated deployment via
  Travis CI"
- The PR's summary contains a list of what you changed
- The interesting part is probably: "Changed setup.py to build a proper
  Python package for PyPI"

Now, if I look at the complete diff for the setup.py, I see lots of
(unrelated) changes: whitespace fixes, import reorganization, a completely
different long_description, explicit requirements etc.
Next, I have to start guessing: what could be a potential problem wrt.
PyPI packaging (you did _not_ even state the problem you are fixing...)
and which of the various changes fix it (and why).
But wait, the title says it fixes something related to travis, hence,
the .travis.yml might have the relevant change... According to the PR
summary, you "[r]eformatted .travis.yml" and "[u]sed before_*
statements instead of script chains". Looking at the diff, you also
removed "bdist_wheel" from the "distributions:".

Heck, I have _no_ clue which problem(s) you are trying to fix... (and my
gut feeling says that even github's "Squash and Merge" feature cannot
really help me in this situation;) *SCNR*)

crazyscientist:
@marcus-h Thank you for the explanation.

Now, let's go through your questions and remarks:

> Now, if I look at the complete diff for the setup.py, I see lots of
(unrelated) changes: whitespace fixes, import reorganization, a completely
different long_description, explicit requirements etc.
* While you are picky about the number of commits and commit messages, I'm quite picky about code style (PEP8, white spaces, order of import statements, ...).
* The old `long_description` was basically just a long summary. If you look at the description of other packages on PyPI, you will notice that it's common practice to use the contents of `README` as `long_description`. So I dared changing this.
* As discussed in the comments of #708 the OBS `.spec` files lists explicit dependencies. Not listing them in the `setup.py` would be bad practice.

> Looking at the diff, you also removed "bdist_wheel" from the "distributions:".
Yes, but not for any special reason. Would you like me to re-add it?

> Heck, I have _no_ clue which problem(s) you are trying to fix...
Well, most of them are cosmetic (Formatting for easier reading). The `before_deploy` is an actual fix of a problem, as **somebody has put files into the `dist` folder, that cannot be published on PyPI**. With this fix deployment will work.

crazyscientist:
@marcus-h Do you have some feedback for me?

marcus-h:
First of all sorry for the late response! I saw your comment but
didn't notice that you also changed the code:/ (github sent no
mail notification...)

On 2020-04-02 06:27:16 -0700, Andreas Hasenkopf wrote:
> @marcus-h Do you have some feedback for me?
>
Apart from the "stuff" where we probably agree to disagree:) only a
very small remark:

> diff --git a/setup.py b/setup.py
> index 91a7eb8..415c1ad 100755
> --- a/setup.py
> +++ b/setup.py

<SNIP>

> +        "Programming Language :: Python :: 3",
> +        "Programming Language :: Python :: 3.4",

python 3.4 is not supported because it does not support bytes
%-interpolation. Can you change that?
Otherwise: LGTM.

crazyscientist:
@marcus-h No problem. I pushed a new commit to remove the classifier (and squashed it for you). Would you mind accepting it now? :crossed_fingers:

marcus-h:
```
  
Date: Fri, 10 Apr 2020 20:07:40 +0200
To: reply+AAIEUTHMOR6Z2ENTKGWZNHF4S26EDEVBNHHCDKGAGM@reply.github.com
Subject: Re: [openSUSE/osc] Fixed problems in automated deployment via Travis
 CI (#739)

On 2020-04-06 00:33:21 -0700, Andreas Hasenkopf wrote:
> @marcus-h No problem. I pushed a new commit to remove the classifier (and squashed it for you).

Thanks!

> Would you mind accepting it now? :crossed_fingers: 
> 

```

PR #740, opened 2020-02-14
fix list of backends for old python-keyring

lethliel:
old python-keyring classes have no name method.
This is used instead:

return self._keyring_backend.__class__.__name__


marcus-h:
Merged. Thanks a lot:)

PR #747, opened 2020-02-19
support request diffing relative to a former request

adrianschroeter:


marcus-h:
On 2020-02-19 08:12:24 -0800, Adrian Schröter wrote:
> index 44694999..2bd71863 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -2090,6 +2090,8 @@ def do_changedevelrequest(self, subcmd, opts, *args):
>
>      @cmdln.option('-d', '--diff', action='store_true',
>                    help='generate a diff')
> +    @cmdln.option('-S', '--superseded-request', metavar='SUPERSEDED_REQUEST',
> +                        help='Create the diff relative to a given former request')

Minor nitpick: the help text sounds more general than it is.

Maybe: "Create the diff relative to a superseded request"?

Btw., is there a reason why we do not allow a diff relative to an "arbitrary"
request? This would allow a more general option name...
Also I'm not really fond of a short option "-S" (I associate(d) "-S" with
"--server-side-source-service-files" but I just realized that it is also used
for "--set-bugowner-request"... *sigh*).

Long story short: LGTM from a technical POV - @lethliel any opinion? If not,
feel free to merge:)

adrianschroeter:
well it works independ of the fact if it is superseded, It just have to have the same target.

I understand your opion regarding -S , but I don't know a better one atm :/

marcus-h:
On 2020-02-20 07:20:22 -0800, Adrian Schröter wrote:
> well it works independ of the fact if it is superseded, It just have to have the same target.
>
Hmm if I'm not completely mistaken the "diff_to_superseded" parameter
should be a superseded reqid (see RequestController.request_command_diff).

> I understand your opion regarding -S , but I don't know a better one atm :/
>
Me neither... An alternative would be to omit the short option.

dmach:
@adrianschroeter could you rebase this PR? Maybe remove the NEWS file from it, we create it on every release from the commit messages anyway.

Issue #748, opened 2020-02-20
name 'oscerr' is not defined


coolo:
```
Default log message was not changed. Press 'c' to continue.
a)bort, c)ontinue, e)dit: Traceback (most recent call last):
  File "/home/coolo/prod/osc/osc/util/helper.py", line 81, in raw_input
    return func(*args)
EOFError: EOF when reading a line

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/coolo/prod/osc/osc-wrapper.py", line 41, in <module>
    r = babysitter.run(osccli)
  File "/home/coolo/prod/osc/osc/babysitter.py", line 64, in run
    return prg.main(argv)
  File "/home/coolo/prod/osc/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/home/coolo/prod/osc/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/home/coolo/prod/osc/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/home/coolo/prod/osc/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/home/coolo/prod/osc/osc/commandline.py", line 2543, in do_request
    opts.message = edit_message()
  File "/home/coolo/prod/osc/osc/core.py", line 4130, in edit_message
    return edit_text(data, delim, suffix='.diff', template=template)
  File "/home/coolo/prod/osc/osc/core.py", line 4151, in edit_text
    ri = raw_input('%s\na)bort, c)ontinue, e)dit: ' % reason)
  File "/home/coolo/prod/osc/osc/util/helper.py", line 84, in raw_input
    raise oscerr.UserAbort()
NameError: name 'oscerr' is not defined

lethliel:
The import of oscerr is missing. Fix incoming.

PR #749, opened 2020-02-20
import oscerr in helper.py

lethliel:
Otherwise oscerr is unknown.

fixes https://github.com/openSUSE/osc/issues/748


marcus-h:
Oops. Merged. Thanks!

PR #752, opened 2020-02-21
Remove compat code for Python < 2.6

dcermak:
Python 2 is dead, Python < 2.6 even more


marcus-h:
Agreed. We don't need to care about python < 2.6 anymore:)
Merged. Thanks!

PR #753, opened 2020-02-22
Fix -n / --name flag on importsrcpkg command.

matthewdva:
osc importsrcpkg -n <pacname> does not work as expected.  If the option is supplied, osc
mistakenly tries to "decode" the pac object.  This patch limits the decode
call when pac is not a string.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -7900,6 +7900,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             title = opts.title
         if opts.name:
             pac = opts.name
+        elif pac is not None:
+            pac = pac.decode()
         if opts.description:
             descr = opts.description
 
@@ -7908,7 +7910,6 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             print('please specify a package name with the \'--name\' option. ' \
                                 'The automatic detection failed', file=sys.stderr)
             sys.exit(1)
-        pac = pac.decode()
         if conf.config['do_package_tracking']:
             createPackageDir(os.path.join(project.dir, pac), project)
         else:


marcus-h:
First of all, thanks for the patch!

On 2020-02-21 17:14:40 -0800, matthewdva wrote:
> index 44694999..c11a1405 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -7908,7 +7908,8 @@ def do_importsrcpkg(self, subcmd, opts, srpm):
>              print('please specify a package name with the \'--name\' option. ' \
>                                  'The automatic detection failed', file=sys.stderr)
>              sys.exit(1)
> -        pac = pac.decode()
> +        if not type(pac) == str:
> +            pac = pac.decode()

Hmm this works but it is a bit "ugly". What about the following the
following patch:

index 4469499..fcd52ca 100644
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -7900,6 +7900,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             title = opts.title
         if opts.name:
             pac = opts.name
+        elif pac is not None:
+            pac = pac.decode()
         if opts.description:
             descr = opts.description

@@ -7908,7 +7910,6 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             print('please specify a package name with the \'--name\' option. ' \
                                 'The automatic detection failed', file=sys.stderr)
             sys.exit(1)
-        pac = pac.decode()
         if conf.config['do_package_tracking']:
             createPackageDir(os.path.join(project.dir, pac), project)
         else:


Feel free to update your PR - afterwards, I'll happily accept it:)

marcus-h:
LGTM. Can you please squash both commits into one?

marcus-h:
Merged. Thanks a lot!

Btw., if you update an osc PR, please leave a comment (so that we get a
mail notification). It was pure luck that I noticed your update:)

Issue #742, opened 2020-02-17
IncompleteRead(0 bytes read)


ignatenkobrain:
This is happening on Fedora Rawhide (33):

```
~/Projects/home:ignatenkobrain:rust 
✦ ❯ osc up
/usr/lib/python3.8/site-packages/osc/core.py:1453: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if r is not 0:
IncompleteRead(0 bytes read)
```

I've tried to debug it a bit and found that the problem is happening around `if options['sslcertck']` in osc/conf.py.

When I removed that part of condition which checks SSL things, everything started to work just fine.

* `osc-0.167.1-281.1.4.fc32.noarch`
* `python3-3.8.2~rc1-1.fc33.x86_64`

lethliel:
Thanks for the report. I will look into it.

lethliel:
Seems to be related to the size of the project (packages count) and if sslcertck is 1: 

This works with sslcertck=1(package count 11):

```
osc api /source/home:mstrigl
<directory count="11">
  <entry name="build"/>
  <entry name="glibmm2"/>
  <entry name="helloSUSE"/>
  <entry name="hellooSC2"/>
  <entry name="hellooSC_docker"/>
  <entry name="mediagoblin"/>
  <entry name="osc"/>
  <entry name="osc-py2"/>
  <entry name="product-builder"/>
  <entry name="test"/>
  <entry name="test-osc"/>
</directory>
```

This works not with sslcertck=1 (package count=274):

```
osc -d api /source/home:ignatenkobrain:rust
GET https://api.opensuse.org/source/home:ignatenkobrain:rust
IncompleteRead(0 bytes read)
```

If I disable sslcertck in the api section it works: 

```
[https://api.opensuse.org]
sslcertck=0
```

```
<directory count="274">
  <entry name="rust-adler32"/>
  <entry name="rust-aho-corasick"/>
  <entry name="rust-ansi_term"/>
  <entry name="rust-approx"/>
  <entry name="rust-arc-swap"/>
  <entry name="rust-arrayref"/>
  <entry name="rust-arrayvec"/>
  <entry name="rust-atty"/>
...

```

lethliel:
The error is raised in http/client.py in function:

```
def _safe_readinto(self, b):
        """Same as _safe_read, but for reading into a buffer."""
        amt = len(b)
        n = self.fp.readinto(b)
        if n < amt:
[...]

PR #755, opened 2020-02-25
core: remove retry limit on log print

pfactum:
If this limit is in place, invoking `buildlog` almost always fails with `IncompleteRead(X bytes read)`. If I remove the retry limit, the `buildlog` just prints the full log as intended. The same applies to `blt`.

Honestly, I do not get why this limit is even in place here. If it is there for some reason, then it either should be increased or another solution should be found because `buildlog` should not fail when it can succeed.

Please review.

Thanks.

Signed-off-by: Oleksandr Natalenko <oleksandr@redhat.com>

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -6132,7 +6132,6 @@ def print_buildlog(apiurl, prj, package, repository, arch, offset=0, strip_time=
         query['last'] = 1
     if lastsucceeded:
         query['lastsucceeded'] = 1
-    retry_count = 0
     while True:
         query['start'] = offset
         start_offset = offset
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2020-02-25 06:41:11 -0800, Oleksandr Natalenko wrote:
> If this limit is in place, invoking `buildlog` almost always fails with `IncompleteRead(X bytes read)`. If I remove the retry limit, the `buildlog` just prints the full log as intended. The same applies to `blt`.
>
> Honestly, I do not get why this limit is even in place here. If it is there for some reason, then it either should be increased or another solution should be found because `buildlog` should not fail when it can succeed.
>
Well usually an IncompleteRead indicates that something is fishy. In
order avoid an "infinite" loop we limit the number of retries (where 3
is arbitrary - we could use the http_retry config option here (see
also the comment in the corresponding merge commit cbd1064)).

Can you post the exact traceback?

A wild guess: you are using python3.8 in combination with m2crypto.
That is, you are probably hitting [1] (see also [2]).

[1] https://github.com/openSUSE/osc/issues/742
[2] https://gitlab.com/m2crypto/m2crypto/-/merge_requests/242

pfactum:
@marcus-h yeah, it seems so, thanks for the pointer. I've rebuilt m2crypto with the suggested patch, and things work fine now.

Lets close this PR then.

PR #758, opened 2020-02-29
Suggest correct python-keyring package version

rindeal:
On Tumbleweed, `zypper in python-keyring` installs python2 version, while `osc` runs on python3.
After this change, user will be pointed to the correct version.


marcus-h:

Issue #761, opened 2020-03-09
Traceback config with two backends and no username


lethliel:
When migrating from an existing oscrc with keyring support the following exception is thrown, when the oscrc includes two APIs:

Conditions:
* osc > 0.166.0 must be used
* no username / password in the API section of oscrc
* osc ls 

I also find it questionable that always the configuration of *all* endpoints is read. 

At the end both credentials_mgr_class entries are generated correct.

```
marco@linux-2ule:~> osc ls home:mstrigl
No user found in section https://api.opensuse.org
Please enter new credentials.
Username: mstrigl
Password: 
1) SecretService Keyring (Backend provided by python-keyring)
2) chainer ChainerBackend (Backend provided by python-keyring)
3) fail Keyring (Backend provided by python-keyring)
4) file EncryptedKeyring (Backend provided by python-keyring)
5) file PlaintextKeyring (Backend provided by python-keyring)
6) kwallet DBusKeyring (Backend provided by python-keyring)
7) Config file credentials manager (Store the credentials in the config file (plain text))
8) Obfuscated Config file credentials manager (Store the credentials in the config file (obfuscated))
9) Transient password store (Do not store the password and always ask for the password)
Select credentials manager: 9
No user found in section https://api.suse.de
Please enter new credentials.
Username: mstrigl
Password: 
1) SecretService Keyring (Backend provided by python-keyring)
2) chainer ChainerBackend (Backend provided by python-keyring)
3) fail Keyring (Backend provided by python-keyring)
4) file EncryptedKeyring (Backend provided by python-keyring)
5) file PlaintextKeyring (Backend provided by python-keyring)
6) kwallet DBusKeyring (Backend provided by python-keyring)
7) Config file credentials manager (Store the credentials in the config file (plain text))
8) Obfuscated Config file credentials manager (Store the credentials in the config file (obfuscated))
9) Transient password store (Do not store the password and always ask for the password)
Select credentials manager: 9
Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.7/site-packages/osc/babysitter.py", line 64, in run
    return prg.main(argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
[...]

PR #762, opened 2020-03-09
active apiurl should be passed to get_config

lethliel:
active_url is determined in commandline.py and could be
options.apiurl given by the user or the default apiurl

This prevents reading all API endpoints when only the active
is needed actually.

fixes https://github.com/openSUSE/osc/issues/761

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -133,6 +133,9 @@ class Osc(cmdln.Cmdln):
     def postoptparse(self, try_again = True):
         """merge commandline options into the config"""
         try:
+            apiurl = conf.DEFAULTS['apiurl']
+            if self.options.apiurl:
+                apiurl = self.options.apiurl
             conf.get_config(override_conffile = self.options.conffile,
                             override_apiurl = self.options.apiurl,
                             override_debug = self.options.debug,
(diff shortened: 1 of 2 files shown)


lethliel:
@marcus-h Please can you have a look. I am not sure if it is the right way to do this.

marcus-h:
On 2020-03-09 06:18:02 -0700, Marco Strigl wrote:
> active_url is determined in commandline.py and could be
> options.apiurl given by the user or the default apiurl
>
Hmm what about the wc?

> This prevents reading all API endpoints when only the active
> is needed actually.
>
Hmm why is this a problem? (It might be "annoying" in case of a
config migration, but well... this is a one time task (IMHO).)

> index 1a223e2e..865a8a7c 100644
> --- a/osc/conf.py
> +++ b/osc/conf.py
> @@ -819,7 +819,8 @@ def get_config(override_conffile=None,
>                 override_post_mortem=None,
>                 override_no_keyring=None,
>                 override_no_gnome_keyring=None,
> -               override_verbose=None):
> +               override_verbose=None,
> +               active_apiurl=None):
>      """do the actual work (see module documentation)"""
>      global config
>
> @@ -886,6 +887,17 @@ def get_config(override_conffile=None,
>          # backward compatiblity
>          scheme, host, path = parse_apisrv_url(config.get('scheme', 'https'), url)
>          apiurl = urljoin(scheme, host, path)
> +        if cp.has_option(url, 'aliases'):
> +            for i in cp.get(url, 'aliases').split(','):
> +                key = i.strip()
> +                if key == '':
> +                    continue
> +                if key in aliases:
> +                    msg = 'duplicate alias entry: \'%s\' is already used for another apiurl' % key
> +                    raise oscerr.ConfigError(msg, conffile)
> +                aliases[key] = url
> +        if active_apiurl and apiurl != active_apiurl and not active_apiurl in aliases:
> +            continue

We would end up with a single entry in the conf.config['api_host_options']
dict. This would "break" commands that work with multiple apiurls (for
instance, "osc copypac --to-apiurl <apiurl2> ...") or the
Project.commitExtPackage codepath.
Also, in cases where it does not "break" (for instance a public route that
requires no authentication), it would bypass the m2crypto code.

PR #764, opened 2020-03-12
Add compactibility with python2 for html.escape

mimi1vx:


lethliel:
Looks good now. 

@marcus-h second opinion?

marcus-h:
First of all, thanks a lot for working on this @mimi1vx!

On 2020-03-12 06:49:41 -0700, Marco Strigl wrote:
> Looks good now.
>
> @marcus-h second opinion?
>
Yes, let's move this into a helper:) (either in util.helper or in the
core module)

try:
    import html
except ImportError:
    import cgi as html


def html_escape(data):
    return html.escape(data, quote=False)


and then simply use "html_escape" to avoid the boilerplate code.
(I'm also fine if we call it "_html_escape" so that it is not part
of the "public" API.)


PS: actually, the commit message is a bit "misleading" because it also
"fixes" the original commit 6dbf103e1030494381c2fbb384f9648a78b68ce6 (html
was sometimes used before importing it).

PPS: s/compactibility/compatibility/ in the commit message.

marcus-h:
Hmm since this affects the users (see boo#1166537), I just pushed a
fix based on this one (see commit 33bbc57b5f1b4cd3aded4a696b095355a8fa4805
("Fix the previously introduced escaping via the html module")).

PR #769, opened 2020-03-16
Fix single service disabled run

cmurphy:
Fix single service disabled run

Without this patch, running an individual service that has parameters
defined in the _service file fails:

  $ osc service run obs_scm
  Please specify valid --scm=... options
  Aborting: service call failed:  /usr/lib/obs/service/obs_scm --outdir [snipped]

This is because although the service is defined in the _service file and
the "scm" parameter is set in it, the service wasn't being found in the
data structure and so the service executable wasn't being called with
the parameters supplied in the _service file. This patch corrects the
issue with the services data structure so that the service data isn't
overridden if it is defined in the _service file. A side effect of this
correction is that instead of overriding the service mode with '', the
mode is set and then does not fall into any permissible callmode
category. We assume that explicit use of "service run <service>" means
the user wants the given service to run now rather than under some
automatic condition, and moreover this was the behavior for services
without parameters before this change, so this patch also changes the
conditional to ignore the mode set in the _service file.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -6938,18 +6938,20 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             osc service COMMAND (inside working copy)
             osc service run [SOURCE_SERVICE]
             osc service runall
-            osc service localrun
-            osc service disabledrun
+            osc service manualrun [SOURCE_SERVICE]
+            osc service localrun [SOURCE_SERVICE]
+            osc service disabledrun [SOURCE_SERVICE]
             osc service remoterun [PROJECT PACKAGE]
             osc service merge [PROJECT PACKAGE]
             osc service wait [PROJECT PACKAGE]
 
             COMMAND can be:
-            run         r  run defined services locally, it takes an optional parameter to run only a
+            run         r  run defined services with modes "trylocal", "localonly", or no mode set locally, may take an optional parameter to run only a
                            specified source service. In case parameters exist for this one in _service file
                            they are used.
             runall      ra run all services independent of the used mode
-            manualrun   mr run all services with mode "manual"
+            manualrun   mr run all services with mode "manual", may take an optional parameter to run only a
+                           specified source service
             remoterun   rr trigger a re-run on the server side
             merge          commits all server side generated files and drops the _service definition
             wait           waits until the service finishes and returns with an error if it failed
--- a/osc/core.py
+++ b/osc/core.py
@@ -422,10 +422,13 @@ class Serviceinfo:
                     os.unlink(ent)
 
         allservices = self.services or []
-        if singleservice and not singleservice in allservices:
+        service_names = [s['name'] for s in allservices]
+        if singleservice and singleservice not in service_names:
             # set array to the manual specified singleservice, if it is not part of _service file
-            data = { 'name' : singleservice, 'command' : [ singleservice ], 'mode' : '' }
+            data = { 'name' : singleservice, 'command' : [ singleservice ], 'mode' : callmode }
             allservices = [data]
+        elif singleservice:
+            allservices = [s for s in allservices if s['name'] == singleservice]
 
         if not allservices:
             # short-circuit to avoid a potential http request in vc_export_env
@@ -449,8 +452,6 @@ class Serviceinfo:
         ret = 0
         for service in allservices:
             if callmode != "all":
-                if singleservice and service['name'] != singleservice:
-                    continue
                 if service['mode'] == "buildtime":
                     continue
                 if service['mode'] == "serveronly" and callmode != "local":


marcus-h:
First of all, thanks for the PR!

On 2020-03-16 14:29:13 -0700, Colleen Murphy wrote:
> index 1ec72af2..bb45d02b 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -420,7 +420,8 @@ def execute(self, dir, callmode = None, singleservice = None, verbose = None):
>                      os.unlink(ent)
>
>          allservices = self.services or []
> -        if singleservice and not singleservice in allservices:
> +        service_names = [s['name'] for s in allservices]
> +        if singleservice and not singleservice in service_names:

Oops... good catch!:)

>              # set array to the manual specified singleservice, if it is not part of _service file
>              data = { 'name' : singleservice, 'command' : [ singleservice ], 'mode' : '' }
>              allservices = [data]
> @@ -453,7 +454,7 @@ def execute(self, dir, callmode = None, singleservice = None, verbose = None):
>                      continue
>                  if service['mode'] == "serveronly" and callmode != "disabled":
>                      continue
> -                if service['mode'] == "disabled" and callmode != "disabled":
> +                if service['mode'] == "disabled" and (callmode != "disabled" and not singleservice):

Hmm I'm not yet convinced about this part: in general, we could do
this. But then, the question is what we should do if someone runs
"osc service run <service_with_mode_buildtime>" or
"osc service run <service_with_mode_serveronly>"
Should we support this, too? If so, then we probably end up with
ignoring the mode and simply execute the single service.

What do you think?

cmurphy:
>>     -                if service['mode'] == "disabled" and callmode != "disabled":
>>     +                if service['mode'] == "disabled" and (callmode != "disabled" and not singleservice):
> 
> Hmm I'm not yet convinced about this part: in general, we could do this. But then, the question is what we should do if someone runs "osc service run <service\_with\_mode\_buildtime>" or "osc service run <service\_with\_mode\_serveronly>" Should we support this, too? If so, then we probably end up with ignoring the mode and simply execute the single service. What do you think?

I guess I need to understand better the use case of the `service run` command. My expectation was that it is meant to run the service, ignoring the mode, so I suppose I would expect "osc service run <service\_with\_mode\_buildtime>" and "osc service run <service\_with\_mode\_serveronly>" to also work.

However, I don't necessarily need that to be the case. What I need is to be able to trigger services with mode="disabled", wiithout using the `service disabledrun` command which is marked as deprecated in the man page. If the advice is to continue using `service disabledrun` for this use case, then I think the documentation should be updated and I would undo this part of the PR.

adrianschroeter:
osc service run $service

was supposed to work independ of _service files only. So you can not expect that it takes arguments from there. We would need to allow to hand them over via CLI instead.

There is indeed no real successor for disabledrun (runall was intended, but it has a different result, because it is not renaming the resulting files).
So, yes, we should remove the "deprecated" marker. (IMHO best would be to have an alias for disabled called "manual" and a manualrun command. This makes it maybe a bit more obvious).

adrianschroeter:
Do you have a use-case for running just a single specified service from a _service file, where it is not enough to specify it via the mode?

marcus-h:
On 2020-03-17 00:24:13 -0700, Adrian Schröter wrote:
> osc service run $service
>
> was supposed to work independ of _service files only. So you can not expect that it takes arguments from there. We would need to allow to hand them over via CLI instead.
>
But only in case the service is not part of the _service (or no _service
file is present at all), right? Otherwise it would "contradict" the
description

"""
run         r  run defined services locally, it takes an optional parameter to run only a
               specified source service. In case parameters exist for this one in _service file
               they are used.
"""

(regardless what the current code is doing)


> There is indeed no real successor for disabledrun (runall was intended, but it has a different result, because it is not renaming the resulting files).
> So, yes, we should remove the "deprecated" marker. (IMHO best would be to have an alias for disabled called "manual" and a manualrun command. This makes it maybe a bit more obvious).
>
Hmm this might be confusing... (one service is with mode="disable" and
another service with mode="manual" => "manualrun" and "disabledrun" would
do the same)

adrianschroeter:
regarding run:

You are right, it is described different atm, so we shouldn't change it. Still we have to deal with the situation:

1) service X does not exist in _service 
 => we do not support parameters here atm
2) service X does exists in _service (one or multiple times, there is not way to specify one atm)
 => there is atm no way to run only a specific one or an manual run with own parameters


regarding "manual":

Yes, I did not "rename" it because it may confuse more then it helps. Maybe we should just clarify it in the help text.

cmurphy:
[**@adrianschroeter**](https://github.com/adrianschroeter) commented on [Mar 17, 2020, 12:42 AM PDT](https://github.com/openSUSE/osc/pull/769#issuecomment-599922811 "2020-03-17T07:42:19Z - Replied by Github Reply Comments"):
> Do you have a use-case for running just a single specified service from a \_service file, where it is not enough to specify it via the mode?

I do not have a use case for just running a single service, my only use case is to avoid writing scripts that depend on a deprecated command.

cmurphy:
Following up on this...I proposed https://github.com/openSUSE/osc/pull/826 to address the suggestion in https://github.com/openSUSE/osc/pull/769#issuecomment-599916468. However, I think this patch is still correct independent of the fate of the "disabledrun" service command. Currently, `osc service run foo` does not work if service "foo" has parameter, this is broken behavior and should be fixed.

cmurphy:
I've changed this patch so that in the single service run case, the "disabled" mode is not treated as special but instead the defined mode is disregarded, see the updated commit message. This is one possibility, other possibilities are to have the single service "run" command respect certain modes but not others (maybe it runs if mode="manual" but not mode="disabled"?) or to have it only run services where there is no mode defined.

marcus-h:
On 2020-07-31 14:00:39 -0700, Colleen Murphy wrote:
> I've changed this patch so that in the single service run case, the "disabled" mode is not treated as special but instead the defined mode is disregarded, see the updated commit message. This is one possibility, other possibilities are to have the single service "run" command respect certain modes but not others (maybe it runs if mode="manual" but not mode="disabled"?) or to have it only run services where there is no mode defined.
>
I know... in your other PR I actively advocated to ignore the mode

... but I thought about it again:) Just let the user (indirectly) specify
the callmode:

$> osc service run foobar -> callmode=None semantics
$> osc service manualrun foobar -> callmode='manual' semantics
$> osc service runall foobar -> callmode='all' semantics (that is, ignore
the mode)

etc. The existing code already supports all these invocations (see the
"singleservice = ..." assignment in commandline.py). It seems the docs
tricked us again (at least I only thought about "osc service run
foobar"). Implementing this is pretty trivial (see below).

What do you think?


> index 469c04e8..4d17638d 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -422,7 +422,8 @@ def execute(self, dir, callmode = None, singleservice = None, verbose = None):
>                      os.unlink(ent)
>
>          allservices = self.services or []
> -        if singleservice and not singleservice in allservices:
> +        service_names = [s['name'] for s in allservices]
> +        if singleservice and not singleservice in service_names:
>              # set array to the manual specified singleservice, if it is not part of _service file
>              data = { 'name' : singleservice, 'command' : [ singleservice ], 'mode' : '' }
>              allservices = [data]

To support the scenario from above, adding something like this should
be sufficient:

elif singleservice:
    allservices = [s for s in allservices if s['name'] == singleservice]

> @@ -451,20 +452,21 @@ def execute(self, dir, callmode = None, singleservice = None, verbose = None):
>              if callmode != "all":
>                  if singleservice and service['name'] != singleservice:
>                      continue

If we do the filtering beforehand, this can be removed. No other changes
are needed:)

> -                if service['mode'] == "buildtime":
> -                    continue
> -                if service['mode'] == "serveronly" and callmode != "local":
> -                    continue
> -                if service['mode'] == "manual" and callmode != "manual":
> -                    continue
> -                if service['mode'] != "manual" and callmode == "manual":
> -                    continue
> -                if service['mode'] == "disabled" and callmode != "disabled":
> -                    continue
> -                if service['mode'] != "disabled" and callmode == "disabled":
> -                    continue
> -                if service['mode'] != "trylocal" and service['mode'] != "localonly" and callmode == "trylocal":
> -                    continue
> +                if not singleservice:
> +                    if service['mode'] == "buildtime":
> +                        continue
> +                    if service['mode'] == "serveronly" and callmode != "local":
> +                        continue
> +                    if service['mode'] == "manual" and callmode != "manual":
> +                        continue
> +                    if service['mode'] != "manual" and callmode == "manual":
> +                        continue
> +                    if service['mode'] == "disabled" and (callmode != "disabled" and not singleservice):
> +                        continue

Hmm a relict from a previous version of this patch?

marcus-h:
On 2020-08-01 00:31:42 +0200, Marcus Hüwe wrote:
> On 2020-07-31 14:00:39 -0700, Colleen Murphy wrote:

<SNIP>

> > index 469c04e8..4d17638d 100644
> > --- a/osc/core.py
> > +++ b/osc/core.py
> > @@ -422,7 +422,8 @@ def execute(self, dir, callmode = None, singleservice = None, verbose = None):
> >                      os.unlink(ent)
> >
> >          allservices = self.services or []
> > -        if singleservice and not singleservice in allservices:
> > +        service_names = [s['name'] for s in allservices]
> > +        if singleservice and not singleservice in service_names:
> >              # set array to the manual specified singleservice, if it is not part of _service file
> >              data = { 'name' : singleservice, 'command' : [ singleservice ], 'mode' : '' }

And we should set mode to callmode:)

> >              allservices = [data]
>
> To support the scenario from above, adding something like this should
> be sufficient:
>
> elif singleservice:
>     allservices = [s for s in allservices if s['name'] == singleservice]
>
> > @@ -451,20 +452,21 @@ def execute(self, dir, callmode = None, singleservice = None, verbose = None):
> >              if callmode != "all":
> >                  if singleservice and service['name'] != singleservice:
> >                      continue
>
> If we do the filtering beforehand, this can be removed. No other changes
> are needed:)

cmurphy:
[**@marcus-h**](https://github.com/marcus-h) commented on [Jul 31, 2020, 3:31 PM PDT](https://github.com/openSUSE/osc/pull/769#issuecomment-667412934 "2020-07-31T22:31:48Z - Replied by Github Reply Comments"):

> I know... in your other PR I actively advocated to ignore the mode
> 
> ... but I thought about it again:) Just let the user (indirectly) specify
> the callmode:
> 
> $> osc service run foobar -> callmode=None semantics
> $> osc service manualrun foobar -> callmode='manual' semantics
> $> osc service runall foobar -> callmode='all' semantics (that is, ignore
> the mode)
> 
> etc. The existing code already supports all these invocations (see the
> "singleservice = ..." assignment in commandline.py). It seems the docs
> tricked us again (at least I only thought about "osc service run
> foobar"). Implementing this is pretty trivial (see below).
> 
> What do you think?

Hmm I'm not really in favor of these semantics because I think of the service command as a filter, not an override. If we implement what you propose then the difference between the behavior of `osc service manualrun` and `osc service manualrun foobar` feels dramatic: the first means "run all services with mode=manual", and the second means "run service foobar no matter what mode is set".

> 
> Hmm a relict from a previous version of this patch?

Oops yes, I had thought I had removed it but had left the removal git-stashed.

marcus-h:
On 2020-08-03 09:28:52 -0700, Colleen Murphy wrote:
> [**@marcus-h**](https://github.com/marcus-h) commented on [Jul 31, 2020, 3:31 PM PDT](https://github.com/openSUSE/osc/pull/769#issuecomment-667412934 "2020-07-31T22:31:48Z - Replied by Github Reply Comments"):
>
> > I know... in your other PR I actively advocated to ignore the mode
> >
> > ... but I thought about it again:) Just let the user (indirectly) specify
> > the callmode:
> >
> > $> osc service run foobar -> callmode=None semantics
> > $> osc service manualrun foobar -> callmode='manual' semantics
> > $> osc service runall foobar -> callmode='all' semantics (that is, ignore
> > the mode)
> >
> > etc. The existing code already supports all these invocations (see the
> > "singleservice = ..." assignment in commandline.py). It seems the docs
> > tricked us again (at least I only thought about "osc service run
> > foobar"). Implementing this is pretty trivial (see below).
> >
> > What do you think?
>
> Hmm I'm not really in favor of these semantics because I think of the service command as a filter, not an override. If we implement what you propose then the difference between the behavior of `osc service manualrun` and `osc service manualrun foobar` feels dramatic: the first means "run all services with mode=manual", and the second means "run service foobar no matter what mode is set".
>
Hmm no that's not what I had in mind:)

osc service manualrun: run all services with mode "manual"
osc service manualrun foobar: run all foobar services with mode "manual"

cmurphy:
[**@marcus-h**](https://github.com/marcus-h) commented on [Aug 3, 2020, 9:59 AM PDT](https://github.com/openSUSE/osc/pull/769#issuecomment-668132100 "2020-08-03T16:59:37Z - Replied by Github Reply Comments"):
> On 2020-08-03 09:28:52 -0700, Colleen Murphy wrote: \*\*\*@\*\*\*.\*\*\*\*\*\]([https://github.com/marcus-h](https://github.com/marcus-h)) commented on \[Jul 31, 2020, 3:31 PM PDT\]([#769 (comment)](https://github.com/openSUSE/osc/pull/769#issuecomment-667412934) "2020-07-31T22:31:48Z - Replied by Github Reply Comments"): 
> 
> Hmm no that's not what I had in mind:) osc service manualrun: run all services with mode "manual" osc service manualrun foobar: run all foobar services with mode "manual"

So if I have a _service file with `<service name="foobar" mode="manual" />` and I run `osc service mr foobar` it would run service "foobar", but if I have `<service name="foobar" />` it would not run?

marcus-h:
On 2020-08-03 10:30:10 -0700, Colleen Murphy wrote:
> [**@marcus-h**](https://github.com/marcus-h) commented on [Aug 3, 2020, 9:59 AM PDT](https://github.com/openSUSE/osc/pull/769#issuecomment-668132100 "2020-08-03T16:59:37Z - Replied by Github Reply Comments"):
> > On 2020-08-03 09:28:52 -0700, Colleen Murphy wrote: \*\*\*@\*\*\*.\*\*\*\*\*\]([https://github.com/marcus-h](https://github.com/marcus-h)) commented on \[Jul 31, 2020, 3:31 PM PDT\]([#769 (comment)](https://github.com/openSUSE/osc/pull/769#issuecomment-667412934) "2020-07-31T22:31:48Z - Replied by Github Reply Comments"):
> >
> > Hmm no that's not what I had in mind:) osc service manualrun: run all services with mode "manual" osc service manualrun foobar: run all foobar services with mode "manual"
>
> So if I have a _service file with `<service name="foobar" mode="manual" />` and I run `osc service mr foobar` it would run service "foobar", but if I have `<service name="foobar" />` it would not run?
>
Yes. For the latter, you would execute "osc service run foobar".
You could also have a _service file that looks like this

<services>
    <service name="foobar">
        <param name="branch">master</param>
    </service>
    <service name="foobar" mode="manual">
        <param name="branch">develop</param>
    </service>
</services>

osc service run foobar: branch is master
osc service mr foobar: branch is develop
osc service runall foobar: run both

cmurphy:
[**@marcus-h**](https://github.com/marcus-h) commented on [Aug 3, 2020, 10:50 AM PDT](https://github.com/openSUSE/osc/pull/769#issuecomment-668157215 "2020-08-03T17:50:31Z - Replied by Github Reply Comments"):

> Yes. For the latter, you would execute "osc service run foobar".
> You could also have a _service file that looks like this
> 
> <services>
> <service name="foobar">
> <param name="branch">master</param>
> </service>
> <service name="foobar" mode="manual">
> <param name="branch">develop</param>
> </service>
> </services>
> 
> osc service run foobar: branch is master
> osc service mr foobar: branch is develop
> osc service runall foobar: run both

When you put it like that I'm surprised that a service is allowed to be defined multiple times at all. That could be potentially hazardous in the runall case. To take the obs_scm service as an example, if you have it defined twice for two different branches, the result would likely be a merge conflict and very confusing changelog entries if using the changesgenerate parameter. Is there a sensible use case for defining a service multiple times with different parameters? I still lean toward trying to make the run command behave more like what is documented rather than changing the behavior of the other commands as well.

marcus-h:
On 2020-08-03 18:23:52 +0000, Colleen Murphy wrote:
> [**@marcus-h**](https://github.com/marcus-h) commented on [Aug 3, 2020, 10:50 AM PDT](https://github.com/openSUSE/osc/pull/769#issuecomment-668157215 "2020-08-03T17:50:31Z - Replied by Github Reply Comments"):
>
> > Yes. For the latter, you would execute "osc service run foobar".
> > You could also have a _service file that looks like this
> >
> > <services>
> > <service name="foobar">
> > <param name="branch">master</param>
> > </service>
> > <service name="foobar" mode="manual">
> > <param name="branch">develop</param>
> > </service>
> > </services>
> >
> > osc service run foobar: branch is master
> > osc service mr foobar: branch is develop
> > osc service runall foobar: run both
>
> When you put it like that I'm surprised that a service is allowed to be defined multiple times at all. That could be potentially hazardous in the runall case. To take the obs_scm service as an example, if you have it defined twice for two different branches, the result would likely be a merge conflict and very confusing changelog entries if using the changesgenerate parameter.

Whether a "runall" makes sense or not depends on the configured services,
IMHO.

> Is there a sensible use case for defining a service multiple times with different parameters?

Yes, for instance

<services>
  <service name="download_url">
    <param name="protocol">https</param>
    <param name="host">example.net</param>
    <param name="path">/release.tar.gz</param>
  </service>
  <service name="download_url" mode="manual">
    <param name="protocol">https</param>
    <param name="host">example.net</param>
    <param name="path">/devel.tar.gz</param>
  </service>
</services>

(in case of "runall", you would end up with two files)

Another example might be the verify_file service etc.

> I still lean toward trying to make the run command behave more like what is documented rather than changing the behavior of the other commands as well.
>
It would just make the (undocumented) "osc service mr foobar" behavior
more sane (IMHO).

The current version of your patch has the "runall" semantics.

cmurphy:
[**@marcus-h**](https://github.com/marcus-h) commented on [Aug 3, 2020, 12:21 PM PDT](https://github.com/openSUSE/osc/pull/769#issuecomment-668197989 "2020-08-03T19:21:19Z - Replied by Github Reply Comments"):


> It would just make the (undocumented) "osc service mr foobar" behavior more sane (IMHO). The current version of your patch has the "runall" semantics.

Good point, I am sold :) Updated

marcus-h:
On 2020-08-03 20:46:00 +0000, Colleen Murphy wrote:
> [**@marcus-h**](https://github.com/marcus-h) commented on [Aug 3, 2020, 12:21 PM PDT](https://github.com/openSUSE/osc/pull/769#issuecomment-668197989 "2020-08-03T19:21:19Z - Replied by Github Reply Comments"):
>
>
> > It would just make the (undocumented) "osc service mr foobar" behavior more sane (IMHO). The current version of your patch has the "runall" semantics.
>
> Good point, I am sold :) Updated
>
Great! The patch looks almost perfect! Below just a very small remark
about the commit message:

> Without this patch, running an individual service that has parameters
> defined in the _service file fails:
>
>   $ osc service run obs_scm
>   Please specify valid --scm=... options
>   Aborting: service call failed:  /usr/lib/obs/service/obs_scm --outdir [snipped]
>
> This is because although the service is defined in the _service file and
> the "scm" parameter is set in it, the service wasn't being found in the
> data structure and so the service executable wasn't being called with
> the parameters supplied in the _service file. This patch corrects the
> issue with the services data structure so that the service data isn't
> overridden if it is defined in the _service file. A side effect of this
> correction is that instead of overriding the service mode with '', the
> mode is set and then does not fall into any permissible callmode

Hmm shouldn't the "not" be deleted? (We set mode to callmode so that
it falls into the permissible callmode category)

> category. We can take this opportunity to define this as the correct

<SNIP>

> index 469c04e8..da52dfa4 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -422,10 +422,13 @@ def execute(self, dir, callmode = None, singleservice = None, verbose = None):
>                      os.unlink(ent)
>
>          allservices = self.services or []
> -        if singleservice and not singleservice in allservices:
> +        service_names = [s['name'] for s in allservices]
> +        if singleservice and not singleservice in service_names:

Minor nitpick: it could be rewritten as

if singleservice and singleservice not in service_names:

(but feel free to keep the current version!)


PS: I really like the detailed commit message:)

cmurphy:
[**@marcus-h**](https://github.com/marcus-h) commented on [Aug 3, 2020, 2:59 PM PDT](https://github.com/openSUSE/osc/pull/769#issuecomment-668262402 "2020-08-03T21:59:26Z - Replied by Github Reply Comments"):
 
>> A side effect of this correction is that instead of overriding the service mode with '', the mode is set and then does not fall into any permissible callmode
> 
> Hmm shouldn't the "not" be deleted? (We set mode to callmode so that it falls into the permissible callmode category)

That's not quite what I meant, I was trying to explain the state of things for services that had a mode set in the _service file if we had only made the fix to the list comprehension on lines 425-426. If it had only been that two-line change, which was focused on just the "run" command with a single service, the mode from the file (e.g. "manual") would no longer match the call mode for the "run" command (None) and so doesn't match any of the conditions. Latest update attempts to elaborate on this.

marcus-h:
On 2020-08-03 15:59:15 -0700, Colleen Murphy wrote:
> [**@marcus-h**](https://github.com/marcus-h) commented on [Aug 3, 2020, 2:59 PM PDT](https://github.com/openSUSE/osc/pull/769#issuecomment-668262402 "2020-08-03T21:59:26Z - Replied by Github Reply Comments"):
>
> >> A side effect of this correction is that instead of overriding the service mode with '', the mode is set and then does not fall into any permissible callmode
> >
> > Hmm shouldn't the "not" be deleted? (We set mode to callmode so that it falls into the permissible callmode category)
>
> That's not quite what I meant, I was trying to explain the state of things for services that had a mode set in the _service file if we had only made the fix to the list comprehension on lines 425-426. If it had only been that two-line change, which was focused on just the "run" command with a single service, the mode from the file (e.g. "manual") would no longer match the call mode for the "run" command (None) and so doesn't match any of the conditions. Latest update attempts to elaborate on this.
>
Ah ok. I like the new version. Unfortunately, I missed a small
detail in the previous version:

> A side effect of this correction for services defined in the _service
> file is that instead of overriding the service mode with '', the mode is
> taken from the _service file. When using the "run" command, this would
> mean that the call mode of None may not be in agreement with the service
> mode defined in the file, e.g. "manual", and so the "run" command would
> no longer cause it to run when it would before. We can take this
> opportunity to define this as the correct behavior - the "run" command
> now only runs services with no mode set - and also ensure that other

Strictly speaking "run"/callmode=None means: trylocal, localonly, no mode


> index 4395fbf8..3bfe00ab 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -6938,18 +6938,20 @@ def do_service(self, subcmd, opts, *args):
>              osc service COMMAND (inside working copy)
>              osc service run [SOURCE_SERVICE]
>              osc service runall
> -            osc service localrun
> -            osc service disabledrun
> +            osc service manualrun [SOURCE_SERVICE]
> +            osc service localrun [SOURCE_SERVICE]
> +            osc service disabledrun [SOURCE_SERVICE]
>              osc service remoterun [PROJECT PACKAGE]
>              osc service merge [PROJECT PACKAGE]
>              osc service wait [PROJECT PACKAGE]
>
>              COMMAND can be:
> -            run         r  run defined services locally, it takes an optional parameter to run only a
> +            run         r  run defined services with no mode set locally, may take an optional parameter to run only a

with modes "trylocal" or "localonly" or no mode

Afterwards, there will be no more "complaints"/remarks (at least
from my side (I promise!:P )).

cmurphy:
[**@marcus-h**](https://github.com/marcus-h) commented on [Aug 3, 2020, 5:04 PM PDT](https://github.com/openSUSE/osc/pull/769#issuecomment-668298799 "2020-08-04T00:04:11Z - Replied by Github Reply Comments"):

> Strictly speaking "run"/callmode=None means: trylocal, localonly, no mode

done

> with modes "trylocal" or "localonly" or no mode 

done

> Afterwards, there will be no more "complaints"/remarks (at least from my side (I promise!:P )).

:)

marcus-h:
Merged. Thanks a lot for the PR (and your patience;) )!

PR #773, opened 2020-03-24
suggest correct python*-m2crypto flavour

jnweiger:
When running with old python2, the suggestion to install package 'python-m2crypto' was correct.
With modern python3 these days, the suggestion should be 'python3-m2crypto' 
Also mention "pip3 install M2Crypto" as an alternative, when no such package is available. E.g. Ubuntu 18.04 and up only have python-m2crypto but no pyhton3-m2crypto

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -526,7 +526,10 @@ def _build_opener(apiurl):
             from M2Crypto import m2urllib2
         except ImportError as e:
             print(e)
-            raise NoSecureSSLError('M2Crypto is needed to access %s in a secure way.\nPlease install python-m2crypto.' % apiurl)
+            if sys.version_info[0] < 3:
+                raise NoSecureSSLError('M2Crypto is needed to access %s in a secure way.\nPlease install python-m2crypto.' % apiurl)
+            else:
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2020-03-24 16:17:02 -0700, Jürgen Weigert wrote:
> index 1a223e2e..8779d270 100644
> --- a/osc/conf.py
> +++ b/osc/conf.py
> @@ -526,7 +526,10 @@ def retry_http_basic_auth(self, host, req, realm):
>              from M2Crypto import m2urllib2
>          except ImportError as e:
>              print(e)
> -            raise NoSecureSSLError('M2Crypto is needed to access %s in a secure way.\nPlease install python-m2crypto.' % apiurl)
> +            if sys.version_info[0] < 3:
> +                raise NoSecureSSLError('M2Crypto is needed to access %s in a secure way.\nPlease install python-m2crypto.' % apiurl)
> +            else:
> +                raise NoSecureSSLError('M2Crypto is needed to access %s in a secure way.\nPlease install python3-m2crypto or try "pip3 install M2Crypto"' % apiurl)

Hmm I'm not sure if we have to mention pip. If the distribution misses the
package, people will find a way to install it... But we could do something
similar to commit fa205be798bfc795f8b60b0f8555dcfe6f5c6653 ("Suggest correct
python-keyring package version").

What do you think?

dmach:

PR #779, opened 2020-04-14
Add ccache argument for oscrc

### LINKED ISSUE #776: osc build has a ccache option that should be in oscrc as well
osc build has a `--ccache` option, which could be in oscrc as well so that it can be used by people as a default rather than just when they remember to add it. It could really help speed up build and development cycles for end users.

Firstyear:
This fixes #776, adding support for ccache to the oscrc configuration to make it easier for people to use in daily workflows.


marcus-h:
Merged. Thanks a lot!

Issue #582, opened 2019-06-18
[help] osc dependson is not the same as osc whatdependson


mcepl:
``osc whatdependson`` is a very useful command to find out how relevant the package is and whether it couldn't be removed. Unfortunately, even after studying ``osc help`` I haven't found that it exists, because it is enlisted only as an alias of ``osc dependson``, which it isn't, these are exact opposites of each other. Please fix ``osc help``.

lethliel:
You are right. The help needs to improved here. Although whatdependson calls the same function only with a reserve flag there needs to be an own help text in osc help.

PR #782, opened 2020-04-21
split dependson/whatdependon in two commands

lethliel:
split the code of do_dependson into two separate commands (just for
the osc help overview)

They are doing the opposite of each other.

Duplicate code was moved to _dependson()

do_whatdependson and do_dependson just call _dependson with an option
reverse set to None or 1.

fixes https://github.com/openSUSE/osc/issues/582

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -5765,16 +5765,15 @@ Please submit there instead, or use --nodevelproject to force direct submission.
 
     # FIXME: the new osc syntax should allow to specify multiple packages
     # FIXME: the command should optionally use buildinfo data to show all dependencies
-    @cmdln.alias('whatdependson')
     def do_dependson(self, subcmd, opts, *args):
-        """${cmd_name}: Show the build dependencies
+        """${cmd_name}: dependson shows the build dependencies inside of a project, valid for a
+        given repository and architecture.
 
-        The command dependson and whatdependson can be used to find out what
-        will be triggered when a certain package changes.
+        The command can be used to find build dependencies (wrt. a given repository and arch)
+        that reside in the same project. To see all build dependencies use the buildinfo command.
+
         This is no guarantee, since the new build might have changed dependencies.
 
-        dependson shows the build dependencies inside of a project, valid for a
-        given repository and architecture.
         NOTE: to see all binary packages, which can trigger a build you need to
               refer the buildinfo, since this command shows only the dependencies
               inside of a project.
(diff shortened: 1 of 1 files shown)


lethliel:
@marcus-h I have absolutely no clue if this even a valid approach to the issue

So please can I have your opinion?

marcus-h:
On 2020-04-21 11:08:11 -0700, Marco Strigl wrote:
> index 923d79b0..da5eaaff 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -5765,12 +5765,12 @@ def do_triggerreason(self, subcmd, opts, *args):
>
>      # FIXME: the new osc syntax should allow to specify multiple packages
>      # FIXME: the command should optionally use buildinfo data to show all dependencies
> -    @cmdln.alias('whatdependson')
>      def do_dependson(self, subcmd, opts, *args):
> -        """${cmd_name}: Show the build dependencies
> +        """${cmd_name}: Show the build dependencies (what the package is relying on)
>
> -        The command dependson and whatdependson can be used to find out what
> -        will be triggered when a certain package changes.
> +        The command dependson can be used to find out what will be triggered when
> +        a certain package changes. In the case of dependson all dependencies the

Hmm I'm not sure if I get this description right. Maybe something like

"The command dependson shows the build dependencies of a certain package."

> +        package relies on are shown.
>          This is no guarantee, since the new build might have changed dependencies.
>
>          dependson shows the build dependencies inside of a project, valid for a

Or just use this ^^^ sentence as a description?

> @@ -5784,17 +5784,48 @@ def do_dependson(self, subcmd, opts, *args):
>
>          usage in package or project directory:
>              osc dependson REPOSITORY ARCH
> -            osc whatdependson REPOSITORY ARCH
>
>          usage:
>              osc dependson PROJECT [PACKAGE] REPOSITORY ARCH
> +
> +        ${cmd_option_list}
> +        """
> +        self._dependson(None, *args)
> +
Suggestion: self._dependson(False, *args)
(IMHO, that's better/more pythonic than using None/1...)

> +
> +    def do_whatdependson(self, subcmd, opts, *args):
> +        """${cmd_name}: Show the build dependencies (what relies on this package)
> +
> +        The command whatdependson can be used to find out what will be triggered when
> +        a certain package changes. In the case of whatdependson all packages are shown
> +        that have a dependency on the given package.
> +
Hmm maybe remove the second sentence? (Or at least the "In the case of
whatdependson" sounds a bit "strange" (because the help text is documenting
the "whatdependson" command).)

> +        This is no guarantee, since the new build might have changed dependencies.
> +
> +        whardependson shows the build dependencies inside of a project, valid for a
> +        given repository and architecture.
> +        NOTE: to see all binary packages, which can trigger a build you need to
> +              refer the buildinfo, since this command shows only the dependencies
> +              inside of a project.
> +
I think we can remove the "NOTE:" part because this only applies to
"dependson".

> +        The arguments REPOSITORY and ARCH can be taken from the first two columns
> +        of the 'osc repos' output.
> +
> +        usage in package or project directory:
> +            osc whatdependson REPOSITORY ARCH
> +
> +        usage:
>              osc whatdependson PROJECT [PACKAGE] REPOSITORY ARCH
>
>          ${cmd_option_list}
>          """
> +        self._dependson(1, *args)
> +
Suggestion: self._dependson(True, *args)

> +
> +    def _dependson(self, reverse, *args):
>          wd = os.curdir
>          args = slash_split(args)
> -        project = packages = repository = arch = reverse = None
> +        project = packages = repository = arch = None
>
>          if len(args) < 2 and (is_package_dir('.') or is_project_dir('.')):
>              self.print_repos()
> @@ -5824,9 +5855,6 @@ def do_dependson(self, subcmd, opts, *args):
>              repository = args[2]
>              arch = args[3]
>
> -        if subcmd == 'whatdependson':
> -            reverse = 1
> -
>          xml = get_dependson(apiurl, project, repository, arch, packages, reverse)
>
>          root = ET.fromstring(xml)
>

Issue #786, opened 2020-04-28
osc bl opensuse_Factory x86_64# weird bug


lethliel:
adding a hash sign to the arch results in a looped director listing: 

```
<directory>
  <entry name="build"/>
  <entry name="ca-certificates-letsencrypt"/>
  <entry name="glibmm2"/>
  <entry name="helloSUSE"/>
  <entry name="hellooSC2"/>
  <entry name="hellooSC_docker"/>
  <entry name="mediagoblin"/>
  <entry name="osc"/>
  <entry name="osc-py2"/>
  <entry name="product-builder"/>
  <entry name="test"/>
  <entry name="test-osc"/>
</directory>
```

PR #787, opened 2020-04-28
quote prj, repository, arch and package 

lethliel:
quote the arguments before creating the apiurl.
This prevents weird behaviour when any has an '#'
at the end.

fixes https://github.com/openSUSE/osc/issues/786

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -5512,7 +5512,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             query = { 'view': 'entry' }
             if opts.last:
                 query['last'] = 1
-            u = makeurl(self.get_api_url(), ['build', project, repository, arch, package, '_log'], query=query)
+            u = makeurl(self.get_api_url(), ['build', quote_plus(project), quote_plus(repository), quote_plus(arch), quote_plus(package), '_log'], query=query)
             f = http_GET(u)
             root = ET.parse(f).getroot()
             offset = int(root.find('entry').get('size'))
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2020-04-28 05:07:14 -0700, Marco Strigl wrote:
> index 46a214a9..f0f32470 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -6132,7 +6132,7 @@ def print_data(data, strip_time=False):
>      while True:
>          query['start'] = offset
>          start_offset = offset
> -        u = makeurl(apiurl, ['build', prj, repository, arch, package, '_log'], query=query)
> +        u = makeurl(apiurl, ['build', prj, repository, quote_plus(arch), package, '_log'], query=query)

We should probably also quote prj, repository and package.

lethliel:
@marcus-h Yes. You are right. I also changed the commit message accordingly.

marcus-h:
On 2020-04-28 05:26:01 -0700, Marco Strigl wrote:
> @marcus-h Yes. You are right. I also changed the commit message accordingly.
>
Thinking about it again: actually, we break the API because 3rd party
users could be smart and pass quoted parameters. That is, we should
actually fix the callers.

Unfortunately, the whole quoting stuff in osc is a bit inconsistent...

What do you think?

lethliel:
@marcus-h 

mhh. I changed the callers now. But IMHO it would make sense to define a central place to do the quoting.. For example in the makeurl method.

marcus-h:
On 2020-04-28 05:49:47 -0700, Marco Strigl wrote:
> @marcus-h
>
> mhh. I changed the callers now. But IMHO it would make sense to define a central place to do the quoting.. For example in the makeurl method.
>
Yes, we should definitely do this. Even if we break the existing API.
That's a _real_ fix that justifies an API breakage:) (IMHO)

For now, I merged the "let the caller do the quoting" workaround.

PR #791, opened 2020-05-12
complete --shell-cmd support

adrianschroeter:
I need this to automate build script tests ...

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -725,6 +725,10 @@ def main(apiurl, opts, argv):
     if opts.shell:
         buildargs.append("--shell")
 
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2020-05-12 07:20:21 +0000, Adrian Schröter wrote:
> -- Commit Summary --
>
>   * bring back --shell-cmd support
>
Hmm we never supported this before (at least according to my memory
and a quick bisect). Are you in the mood for slightly updating the
commit message?:) *duck*
Otherwise, LGTM (given that we already have the option).

lethliel:
Commit was renamed

PR #793, opened 2020-05-19
add option pkg-ccache to be passed to build-cmd

sjamgade:
this option is a /path/to/_ccache.tar. The patch just forwards that
option to the build script.

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -622,6 +622,9 @@ def main(apiurl, opts, argv):
     if opts.ccache or config['ccache']:
         buildargs.append('--ccache')
         xp.append('ccache')
+    if opts.pkg_ccache:
+        buildargs.append('--pkg-ccache=%s' % opts.pkg_ccache)
+        xp.append('ccache')
     if opts.linksources:
         buildargs.append('--linksources')
     if opts.baselibs:
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -6303,6 +6303,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                   help='use N parallel build jobs with icecream')
     @cmdln.option('--ccache', action='store_true',
                   help='use ccache to speed up rebuilds')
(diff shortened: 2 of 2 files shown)


marcus-h:
On 2020-05-19 06:18:11 -0700, Sumit Jamgade wrote:
> index 7d2b8e6a..863ab19c 100644
> --- a/osc/build.py
> +++ b/osc/build.py
> @@ -622,6 +622,9 @@ def main(apiurl, opts, argv):
>      if opts.ccache or config['ccache']:
>          buildargs.append('--ccache')
>          xp.append('ccache')
> +    if opts.pkg_ccache:
> +        buildargs.append('--pkg-ccache %s' % opts.pkg_ccache)
> +        xp.append('ccache')

Hmm this won't work. We should do something like

buildargs.append('--pkg-ccache=%s' % opts.pkg_ccache)

>      if opts.linksources:
>          buildargs.append('--linksources')
>      if opts.baselibs:
> index c7bddc96..3f1ee062 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -6303,6 +6303,8 @@ def parse_repoarchdescr(self, args, noinit = False, alternative_project = None,
>                    help='use N parallel build jobs with icecream')
>      @cmdln.option('--ccache', action='store_true',
>                    help='use ccache to speed up rebuilds')
> +    @cmdln.option('--pkg-ccache', action='store', type='string', metavar='/path/to/_ccache.tar',

action and type can be omitted.

> +                  help='path to existing uncompressed archive ccache. Using this option implies --ccache')

Maybe: "path to an existing uncompressed ccache archive. Using..."?

sjamgade:
I would prefer to keep the metavar. It stresses on the "uncompressed"  part

marcus-h:
On 2020-05-19 07:01:08 -0700, Sumit Jamgade wrote:
> I would prefer to keep the metavar. It stresses on the "uncompressed"  part
>
The metavar part is perfectly fine. I was just referring to the "action"
and "type" args.

lethliel:
Looks good now

PR #795, opened 2020-05-25
osc ls on local working copy in project/package

lethliel:
if the currend working directory is a valid copy of a project or
package use osc ls based on the cwd

@marcus-h Your opinion, please :-)

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -417,17 +417,28 @@ class Osc(cmdln.Cmdln):
         # list sources
         elif not opts.binaries:
             if not args:
-                for prj in meta_get_project_list(apiurl, opts.deleted):
-                    print(prj)
+                is_project = None
+                is_package = None
+                cwd = os.getcwd()
+                if is_project_dir(cwd):
+                    project = store_read_project(cwd)
+                    is_project = True
+                elif is_package_dir(cwd):
+                    project = store_read_project(cwd)
+                    package = store_read_package(cwd)
+                    is_package = True
+                else:
+                    for prj in meta_get_project_list(apiurl, opts.deleted):
+                        print(prj)
 
-            elif len(args) == 1:
+            if len(args) == 1 or is_project:
                 if opts.verbose:
                     if self.options.verbose:
                         print('Sorry, the --verbose option is not implemented for projects.', file=sys.stderr)
                 for pkg in meta_get_packagelist(apiurl, project, deleted = opts.deleted, expand = opts.expand):
                     print(pkg)
 
-            elif len(args) == 2 or len(args) == 3:
+            elif len(args) == 2 or len(args) == 3 or is_package:
                 link_seen = False
                 print_not_found = True
                 rev = opts.revision


marcus-h:
On 2020-05-25 11:44:59 -0700, Marco Strigl wrote:
> index 782d2308..ae709eca 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -417,17 +417,28 @@ def do_list(self, subcmd, opts, *args):
>          # list sources
>          elif not opts.binaries:
>              if not args:
> -                for prj in meta_get_project_list(apiurl, opts.deleted):
> -                    print(prj)
> +                is_project = None
> +                is_package = None

Minor: use False instead of None:)

> +                cwd = os.getcwd()
> +                if is_project_dir(cwd):
> +                    project = store_read_project(cwd)
> +                    is_project = True
> +                elif is_package_dir(cwd):
> +                    project = store_read_project(cwd)
> +                    package = store_read_package(cwd)
> +                    is_package = True
> +                else:
> +                    for prj in meta_get_project_list(apiurl, opts.deleted):
> +                        print(prj)
>
> -            elif len(args) == 1:
> +            if len(args) == 1 or is_project:
>                  if opts.verbose:
>                      if self.options.verbose:
>                          print('Sorry, the --verbose option is not implemented for projects.', file=sys.stderr)
>                  for pkg in meta_get_packagelist(apiurl, project, deleted = opts.deleted, expand = opts.expand):
>                      print(pkg)
>
> -            elif len(args) == 2 or len(args) == 3:
> +            elif len(args) == 2 or len(args) == 3 or is_package:

IMHO, the new behavior is a bit confusing because we add some "ambiguity":/

Examples:
cd prj
osc ls pkg          # does a project listing instead of a package listing
osc ls -v pkg file  # does a project listing instead of showing file details

cd prj/pkg
osc ls file         # does a project listing for the "file" project
osc ls x y          # does a package listing but maybe we should error out
                    # because we are supposed to be "context" sensitive (or
                    # maybe not?:) )


I guess what you had in mind is already implemented as

cd prj
osc ls .

and

cd prj/pkg
osc ls .

(Of course, one could still do "stupid" things like "cd prj/pkg;
osc ls . file" but maybe the "." helps/indicates that this might not
work as expected.)

lethliel:
already implemented as osc ls .

marcus-h:
On 2020-05-25 23:02:05 +0200, Marcus Hüwe wrote:

<SNIP>

> I guess what you had in mind is already implemented as
>
Oh my... it is not...

> cd prj
> osc ls .
>
Expected (IMHO) behavior: package listing
Actual behavior: package listing

> and
>
> cd prj/pkg
> osc ls .
>
Expected (IMHO) behavior: file listing
Actual behavior: package listing

> (Of course, one could still do "stupid" things like "cd prj/pkg;
> osc ls . file" but maybe the "." helps/indicates that this might not
> work as expected.)

Even worse... this does a package listing for the "file" package.

Long story short: the existing "osc ls . <args>" behavior is already
confusing (IMHO). We could "fix" it (but it might "break" some users).
However, we should not add yet another context sensitive behavior (what
you are proposing) because it could break existing users (IMHO).

PR #796, opened 2020-05-25
[WIP] show diff when mr is created

lethliel:
As in do_submitrequest show the diff also when do_maintenancerequest is called.

This needs more work.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -3291,7 +3291,27 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             print('Using target project \'%s\'%s' % (target_project, release_in))
 
         if not opts.message:
-            opts.message = edit_message()
+            try:
+                rdiff = b'old: %s/%s\nnew: %s/%s\n' % (target_project.encode(), source_package.encode(), release_project.encode(), source_package.encode())
+                rdiff += server_diff(apiurl,
+                                release_project, source_package, None,
+                                source_project, source_package, None, True)
+            except:
+                rdiff = b''
+            if rdiff is not None:
+                rdiff = decode_it(rdiff)
+            difflines = []
+            doappend = False
+            changes_re = re.compile(r'^--- .*\.changes ')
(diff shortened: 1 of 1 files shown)


lethliel:
@marcus-h: This is for discussion.

marcus-h:
On 2020-05-25 11:49:46 -0700, Marco Strigl wrote:
> index 782d2308..54dbdb06 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -3291,7 +3291,27 @@ def do_maintenancerequest(self, subcmd, opts, *args):
>              print('Using target project \'%s\'%s' % (target_project, release_in))
>
>          if not opts.message:
> -            opts.message = edit_message()
> +            try:
> +                rdiff = b'old: %s/%s\nnew: %s/%s\n' % (target_project.encode(), source_package.encode(), release_project.encode(), source_package.encode())
> +                rdiff += server_diff(apiurl,
> +                                release_project, source_package, None,
> +                                source_project, source_package, None, True)

Hmm we do not have necessarily have a single source source (source_package is
not defined). This should probably be wrapped in a for-loop that iterates over
source_packages.

> +            except:
> +                rdiff = b''
> +            if rdiff is not None:
> +                rdiff = decode_it(rdiff)

rdiff is never None (and we should always decode_it (see below)).

> +            difflines = []
> +            doappend = False
> +            changes_re = re.compile(r'^--- .*\.changes ')
> +            for line in rdiff.split('\n'):

rdiff might be a bytes instance (see above).

> +                if line.startswith('--- '):
> +                    if changes_re.match(line):
> +                        doappend = True
> +                    else:
> +                        doappend = False
> +                if doappend:
> +                    difflines.append(line)
> +            opts.message = edit_message(footer=rdiff, template='\n'.join(parse_diff_for_commit_message('\n'.join(difflines))))
>

This duplicates some code from do_submitrequest - maybe we can refactor the
common code into a new method (in commandline.py?).

What do you think?

dmach:

PR #797, opened 2020-05-27
fix security issue (bsc#1122675) no / in filename

lethliel:
This checks if the filename of a downloaded file has
been modified (for example by a MITM attack) to contain
slashes. This could mean that the file is compromised
and that the attacker tries to overwrite system files.


lethliel:
already discussed with @marcus-h on IRC.

PR #799, opened 2020-05-29
Remove python2 compat code in myHTTPSHandler

dcermak:


marcus-h:
On 2020-05-29 13:20:52 -0700, Dan Čermák wrote:
> -- Commit Summary --
>
>   * Remove python2 compat code in myHTTPSHandler
>
I'm not sure if we should actively start to remove python2 code...
Admittedly, it has to die at some point but maybe not now (IMHO).

lethliel:
I wouldn't remove it at this point. And if we remove python2 compat code we should remove it everywhere and release a new major release of osc.

@dcermak Thanks for your effort. No offend :-)

dcermak:
Sure, no problem.

PR #800, opened 2020-06-02
Fix and standardize ElementTree imports for Python 3.9

AdamWill:
Importing `cElementTree` has been deprecated since Python 3.3 -
importing `ElementTree` automatically uses the fastest
implementation available - and is finally removed in Python 3.9.
Importing cElementTree directly (not as part of xml) is an even
older relic, it's for Ye Time Before ElementTree Was Added To
Python and it was instead an external module...which was before
Python 2.5. So let's just standardize on importing ElementTree
from xml.etree. Also, let's not repeat this import 12 times in
one file for some reason.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/README
+++ b/README
@@ -24,10 +24,6 @@ Alternatively, you can directly use osc-wrapper.py from the source dir
 (which is easier if you develop on osc).
 
 
-The program needs the cElementTree python module installed. On SUSE, the
-respective package is called python-elementtree (before 10.2: python-xml).
-
-
 
 CONFIGURATION:
 
--- a/osc/build.py
+++ b/osc/build.py
@@ -28,9 +28,12 @@ import osc.conf
 from . import oscerr
 import subprocess
 try:
+    # Works up to Python 3.8, needed for Python < 3.3 (inc 2.7)
     from xml.etree import cElementTree as ET
 except ImportError:
-    import cElementTree as ET
+    # will import a fast implementation from 3.3 onwards, needed
+    # for 3.9+
+    from xml.etree import ElementTree as ET
 
 from .conf import config, cookiejar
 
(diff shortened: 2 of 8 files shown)


AdamWill:
I wrote this assuming we don't have to stay compatible with Python 2.7. If we do, we'll need to instead use a try/except block like this:

    try:
        from xml.etree import cElementTree as ET
    except ImportError:
        from xml.etree import ElementTree as ET

I can change that up if Python 2.7 support still matters, let me know.

Conan-Kudo:
@AdamWill This project still cares about backwards compatibility going back to Python 2.6 (!!!)

marcus-h:
On 2020-06-02 13:30:48 -0700, Adam Williamson wrote:
> I wrote this assuming we don't have to stay compatible with Python 2.7. If we do, we'll need to instead use a try/except block like this:
>
>     try:
>         from xml.etree import cElementTree as ET
>     except ImportError:
>         from xml.etree import ElementTree as ET
>
> I can change that up if Python 2.7 support still matters, let me know.
>
I would prefer to keep the python2.7 support because doing so is pretty
cheap (at the moment). If you think it is too much boilerplate we could
also move this stuff into a separate module. For instance,

in osc/util/xml.py:

try:
    from xml.etree import cElementTree as ET
except ImportError:
    from xml.etree import ElementTree as ET

and then simply

from .util.xml import ET

But that's up to you:)

Thanks a lot for your work so far!

AdamWill:
eh, it's easy enough to just boilerplate it around. I'll update the PR.

marcus-h:
On 2020-06-02 14:53:16 -0700, Adam Williamson wrote:
> eh, it's easy enough to just boilerplate it around. I'll update the PR.
>

PS: in the future, it would be great if you could send a small reply to
the corresponding thread once you are done with the update (otherwise, I
usually do not notice that something changed):)

Thanks again!

PR #802, opened 2020-06-03
Fix the testsuite for python3

marcus-h:
Fix the testsuite for python3. More precisely, make the testsuite bytes aware and
change the way how XML documents are compared (compare trees instead of a
string comparison of the serialized documents). For the details, see the individual
commits.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -7,9 +7,6 @@ import sys
 from xml.etree import cElementTree as ET
 EXPECTED_REQUESTS = []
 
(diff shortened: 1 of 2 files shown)


lethliel:
@marcus-h Thanks for fixing the testsuite.
LGTM

marcus-h:
On 2020-06-04 02:34:44 -0700, Marco Strigl wrote:
> @marcus-h Thanks for fixing the testsuite.
> LGTM
>
Ok, thanks! Merged.

Issue #746, opened 2020-02-19
Very slow local buildlog parsing


AdamMajer:
```
> rpm -q osc
osc-0.167.2-lp152.289.1.noarch

> du /var/tmp/build-root/openSUSE_Tumbleweed-x86_64/.build.log
27908   /var/tmp/build-root/openSUSE_Tumbleweed-x86_64/.build.log

> time /usr/bin/osc lbl > /dev/null
real    0m30.900s
user    0m30.722s
sys     0m0.172s

> time cat /var/tmp/build-root/openSUSE_Tumbleweed-x86_64/.build.log > /dev/null
real    0m0.043s
user    0m0.003s
sys     0m0.005s
```

lethliel:
I think the decoding of the read buildlog is to blame here. 

On a very small buildlog the read times are:

READ:
0.00017642974853515625
DECODE:
1.263442039489746

lethliel:
@AdamMajer Can you tell me which package you were building? Just that I have the same results as you.

AdamMajer:
@lethliel libreoffice ;)

Issue #669, opened 2019-11-04
check in huge shell archives is insanely slow


frispete:
osc 0.166.1 as of TW 20191101 is insanely slow in checking in huge shell archives like the nvidia G05 driver, that comes as a 135 MiB shell archive.

After switching to Python 3, osc's diff handling used to be broken for anything, it cannot decode. Now, it does handle those cases again, but..

Example of checking in a new driver 440.31 of nvidia-gfxG05 to a local OBS (client and server running on decent hardware):
```
$ time oscl ci
Default log message was not changed. Press 'c' to continue.
a)bort, c)ontinue, e)dit: a
aborted.

real    31m10,899s
user    31m1,173s
sys     0m1,694s
```

Compare with osc 0.162.1 (Leap 15.0):
```
$ time oscl ci
Deleting    NVIDIA-Linux-x86_64-440.26.run
Sending    NVIDIA-Linux-x86_64-440.31.run
Transmitting file data .
Committed revision 18.

real    0m29,472s
user    0m10,212s
sys     0m2,851s
```

It took me less than 2 seconds to edit and decline the first operation and to accept the second.

Steps to reproduce:
```
osc branch X11:Drivers:Video/nvidia-gfxG05
osc co home:USERNAME:branches:X11:Drivers:Video/nvidia-gfxG05
osc service dr # fetches NVIDIA-Linux-x86_64-440.31.run
osc ar
osc ci # just cancel this operation
```

marcus-h:
On 2019-11-04 16:47:07 +0000, Hans-Peter Jansen wrote:
> osc 0.166.1 as of TW 20191101 is insanely slow in checking in huge shell archives like the nvidia G05 driver, that comes as a 135 MiB shell archive.
>
> After switching to Python 3, osc's diff handling used to be broken for anything, it cannot decode. Now, it does handle those cases again, but..
>
> Example of checking in a new driver 440.31 of nvidia-gfxG05 to a local OBS (client and server running on decent hardware):
> ```
> $ time oscl ci
> Default log message was not changed. Press 'c' to continue.
> a)bort, c)ontinue, e)dit: a
> aborted.
>
> real    31m10,899s
> user    31m1,173s
> sys     0m1,694s
> ```
>
> Compare with osc 0.162.1 (Leap 15.0):
> ```
> $ time oscl ci
> Deleting    NVIDIA-Linux-x86_64-440.26.run
> Sending    NVIDIA-Linux-x86_64-440.31.run
[...]

PR #803, opened 2020-06-04
Do not use the chardet module in util.helper.decode_it

marcus-h:
In general, decode_it is used to get a str from an arbitrary bytes
instance. For this, decode_it used the chardet module (if present)
to detect the underlying encoding (if the bytes instance corresponds
to a "supported" encoding). The drawback of this detection is that
it can take quite some time in case of a large bytes instance, which
represents no "supported" encoding (see #669 and #746).
Instead of doing a potentially "time consuming" detection, either
assume an utf-8 encoding or a latin-1 encoding. Rationale: it is just
not worth the effort to detect a _potential_ encoding because we have
no clue what the _correct_ encoding is. For instance, consider the
following bytes instance:

b'This character group is not supported: [abc\xc3\xbf]'

It represents a valid utf-8 and latin-1 encoding. What is the "correct"
one? We don't know... Even if you interpret the bytes instance as a
human you cannot give a definite answer (implicit assumption: there is
no additional context available).
That is, if we cannot give a definite answer in case of two potential
encodings, there is no point in bringing even more potential encodings
into play. Hence, do not use the chardet module.

Note: the rationale for trying utf-8 first is that utf-8 is pretty
much in vogue these days and, hence, the chances are "high" that we
guess the "correct" encoding.

Fixes: #669 ("check in huge shell archives is insanely slow")
Fixes: #746 ("Very slow local buildlog parsing")

### DIFF
--- a/osc/util/helper.py
+++ b/osc/util/helper.py
@@ -56,22 +56,19 @@ def decode_list(ilist):
 
 
 def decode_it(obj):
-    """ Decodes the given object if obj is not a string
-        based on the chardet module if possible
-    """
+    """Decode the given object.
 
-    if obj is None or isinstance(obj, str):
+    If the given object has no decode method, the object itself is
+    returned. Otherwise, try to decode the object using utf-8. If this
+    fails due to a UnicodeDecodeError, try to decode the object using
+    latin-1.
+    """
+    if not hasattr(obj, 'decode'):
         return obj
-    else:
-        try:
-            import chardet
-            return obj.decode(chardet.detect(obj)['encoding'])
-        except:
-            try:
-                import locale
-                return obj.decode(locale.getlocale()[1])
-            except:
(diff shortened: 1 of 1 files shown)


lethliel:
Yes. At the time i ported osc to python3 I assumed It would be a good idea to use chardet. 
But only checking for utf-8 and latin-1 makes more sense and is a lot faster.

frispete:
Since the new Linux kernel is out, that most probably will need some NVidia fixes, I will soon see the outcome. 

Giving a heads up, when this code is released for TW would be very nice.

lethliel:
@frispete I will release next week. Latest wednesday. Is that ok for you?

frispete:
Please don't feel pressured by me. Just leave me a note, when released.

Thanks a lot, @lethliel, for your continued support of this extremely useful and valuable gem in our software universe.

Issue #814, opened 2020-06-25
osc log | fails


mmohring:
Since the fixes in osc log, osc log | command fails with a buffer error

mmohring:
Pyhon log of the failure:

$ osc log | less

Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python2.7/site-packages/osc/babysitter.py", line 64, in run
    return prg.main(argv)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 347, in main
    return self.cmd(args)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 370, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 504, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 1235, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python2.7/site-packages/osc/commandline.py", line 6874, in do_log
    run_pager(log)
  File "/usr/lib/python2.7/site-packages/osc/core.py", line 4073, in run_pager
    sys.stdout.buffer.write(message)
  File "/usr/lib/python2.7/site-packages/osc/util/safewriter.py", line 20, in __getattr__
    return getattr(self._writer, name)
AttributeError: 'file' object has no attribute 'buffer'
The last version working was commit e807abbf7c68ce6e04d5b34e6d9d0289942ed97a. I guess has to do with the code fixes for osc log.

marcus-h:
On 2020-06-25 04:54:54 -0700, Martin Mohring wrote:
> $ osc log | less
>
> Traceback (most recent call last):
>   File "/usr/bin/osc", line 41, in <module>
>     r = babysitter.run(osccli)
>   File "/usr/lib/python2.7/site-packages/osc/babysitter.py", line 64, in run
>     return prg.main(argv)
>   File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 347, in main
>     return self.cmd(args)
>   File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 370, in cmd
>     retval = self.onecmd(argv)
>   File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 504, in onecmd
>     return self._dispatch_cmd(handler, argv)
[...]

PR #815, opened 2020-06-25
Fix python2 regression in util.helper.decode_it

marcus-h:
In commit 276d6e2439c8c53c182dbe785b038919e64da9f3 ("Do not use the
chardet module in util.helper.decode_it") util.helper.decode_it was
changed to always decode the passed object if it has a decode method.
Since a python2 str has a decode method, the new code tries to utf-8
decode the passed str. As a result, a unicode object is returned (if
the decoding worked). Since a unicode object is not an instance of
type str, all subsequent isinstance(decoded_obj, str) checks evaluate
to False, which break some codepaths.
In order to fix this, restore the old python2 behavior (that is, if
the passed object is a str, it is not decode it). This change does not
affect the python3 codepaths.

Fixes: #814 ("osc log | fails")

### DIFF
--- a/osc/util/helper.py
+++ b/osc/util/helper.py
@@ -56,14 +56,14 @@ def decode_list(ilist):
 
 
 def decode_it(obj):
-    """Decode the given object.
+    """Decode the given object unless it is a str.
 
-    If the given object has no decode method, the object itself is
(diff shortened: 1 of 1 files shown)


lethliel:
LGTM

Issue #801, opened 2020-06-03
checkin command does not interpret newline characters


glaubitz:
It is possible to add multi-line changelog entries with ```osc vc``` using:

```
# osc vc -m "Initial build\n  + Version 1.2.3"
```

results in:

```
-------------------------------------------------------------------
Wed Jun  3 12:48:40 UTC 2020 - John Paul Adrian Glaubitz <adrian.glaubitz@suse.com>

- Initial build
  + Version 1.2.3
```

The same unfortunately does not work with ```osc checkin```:

```
# osc checkin -m "Initial build\n  + Version 1.2.3"
```

as this results in:

```
----------------------------------------------------------------------------
r2 | glaubitz | 2020-06-03 12:42:19 | 401d8eff06984b98c4db9fc3b6e33cbc | unknown | 

- Initial build\n  + Version 1.2.3
```

lethliel:
Hi,

this is not possible. osc just transfers the string "Initial build\n  + Version 1.2.3". 

For `osc vc -m` it transfers it to build/vc
build/vc echoes the string with `echo -e` into the changes file
From the man page of echo: 

```
-e     enable interpretation of backslash escapes
```

For osc checkin -m the string is given as a query parameter to the API and the API does not interpret newline characters here.

glaubitz:
The API does support multi-line commit messages though. So I'm not sure why it shouldn't accept multiline commit messages on the command line.

lethliel:
Huh? Then I will further investigate. My last information that it does not interpret newline characters

lethliel:
Ah,. It is the backend that does not interpret the newlines

coolo:
Note that the change was reverted. The proper way to have multiline commit logs is using new line characters:
```
coolo@localhost#product-builder>python3.10 ~/prod/osc/osc-wrapper.py ci -m "Hallo^MIhr"
Sending    product-builder.changes
Transmitting file data .
Committed revision 10.
coolo@localhost#product-builder>osc log | head
----------------------------------------------------------------------------
r10 | coolo | 2022-07-08 08:02:06 | 9d6d9d5e097192374b7b4ff3500ab931 | unknown | 

Hallo
Ihr
```

Your assumption that \n in command line arguments is creating new lines is just wrong:

```
coolo@localhost#product-builder>echo "Hallo\nIhr"
Hallo\nIhr
```

osc vc is just strange.

glaubitz:
Why revert it when this little change could have just been implented?

[...]

PR #816, opened 2020-06-25
interpretation of string literals in messages

lethliel:
string literals are not interpreted correctly when using `-m "my\ntwo line message"`

fixes https://github.com/openSUSE/osc/issues/801

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -911,6 +911,10 @@ class Osc(cmdln.Cmdln):
         # If project or package arguments missing, assume to work
         # with project and/or package in current local directory.
         attributepath = []
+
+        if opts.message:
+            opts.message = str(opts.message.encode().decode('unicode_escape'))
+
         if cmd in ['prj', 'prjconf']:
             if len(args) < 1:
                 apiurl = store_read_apiurl(os.curdir)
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2020-06-25 11:50:03 -0700, Marco Strigl wrote:
> string literals are not interpreted correctly when using `osc ci -m &quot;my\ntwo line message&quot;`
>
Hmm there are some places where we do not perform the unicode_escape stuff:
- do_createrequest
- do_requestmaintainership
- do_deleterequest
- do_changedevelrequest
- do_request
- do_createincident

The change in do_submitrequest looks "wrong": in the "is_project_dir" case,
we do not interpret escape characters (while we should). In the "other" case,
we might also interpret escape sequences that are part of a diff (we should
not do that). Just move the "if opts.message: opts.message = ..." code to
the start of the do_submitrequest method.

lethliel:
@marcus-h Yeah. Missed those. 

moved the decoding at the beginning of do_submitrequest.
Also in do_request.

marcus-h:
On 2020-06-26 01:00:03 -0700, Marco Strigl wrote:
> @marcus-h Yeah. Missed those.
>
do_createincident is still missing:)

lethliel:
@marcus-h *args* Now done

marcus-h:
On 2020-06-26 15:51:52 +0000, Marco Strigl wrote:
> @marcus-h *args* Now done
>
Merged. Thanks!

Issue #808, opened 2020-06-12
[RFE] osc reset


mcepl:
(edited log of the IRC conversation with @lethliel to have it recorded here):

That's comething completely different … what I mean is something like ``git reset --hard``: I sent something wrong to my devel project, and then I decide that it was stupid and I would like to get rid off it, but it is already on OBS.

>       So you want to switch to a former revision and submit that to your devel project?

Yes, basically roll back the project to some previous revision or (because I couldn't rewrite the history), to continue the project as if some previous revisions never happened.

>       You will get a osc reset command for that.

My current version of the shell script is:
```sh
#!/bin/bash
set -eu

function usage() {
    echo "usage: osc_reset -r REVNO" >&2
    exit 1
}

function die() {
    echo "$1" >&2
    exit 2
}

[[ "$#" -eq 2 ]] || usage
ls .osc/ >/dev/null 2>&1 || die "Command must be run from OSC package checkout."

REVNO=$2
PRJ=$(cat .osc/_project)
PKG=$(cat .osc/_package)

SR_OUT=$(osc sr -m 'revert' -r 6 "$PRJ" "$PKG" "$PRJ" "$PKG")
REQ_ID=$(echo "$SR_OUT"|awk '/created request id/ { print $4 }')
[[ -n "$REQ_ID" ]] || die "Submit request has not been created properly."

osc request accept -m ok $SR_OUT
```

lethliel:
Thanks for putting this in an issue. 
Will start to work on thins next monday.

marcus-h:
On 2020-06-12 05:23:27 -0700, Matěj Cepl wrote:
> (edited log of the IRC conversation to have it recorded here):
>
> That's comething completely different ... what I mean is
> something like git reset --hard I sent something wrong to my
> devel project and then decide that it is a stupid and I would
> like to get rid off it, but it is already on OBS.
>
> >       So you want to switch to a former revision and submit
> >       that to your devel project?
>
> yes, basically roll back the project to some previous revision or
> (because I couldn't rewrite the history), to continue the project
> as if some previous revisions never happened.
>
Hmm what should happen if your link target evolved in the meantime?

> >       You will get a osc reset command for that.
>
Hmm we maybe we could reuse the existing "osc revert" command.

mcepl:
> Hmm what should happen if your link target evolved in the meantime?

I don’t understand.

> Hmm we maybe we could reuse the existing "``osc revert``" command.

[...]

PR #823, opened 2020-07-07
add revert command to revert a package to a rev

lethliel:
This adds a revert command to restore a specific revision
of a package.

Basically this is just a `osc copypac -r <rev> <proj> <pac> <proj>`

implements https://github.com/openSUSE/osc/issues/808

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -2947,6 +2947,42 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         aggregate_pac(src_project, src_package, dst_project, dst_package, repo_map, opts.disable_publish, opts.nosources)
 
 
+    @cmdln.option('-r', '--revision', metavar='rev',
+                        help='reset to specified revision.')
+    def do_reset(self, subcmd, opts, *args):
+        """${cmd_name}: Reset a package
+
+        This resets a package on the server to a given revision.
+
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2020-07-07 01:23:56 -0700, Marco Strigl wrote:
> This adds a revert command to restore a specific revision
> of a package.
>
Hmm I'm not sure if the command does the "expected" thing [1] (from an
average user's POV). Let's try to clarify the "expected" semantics first
(note: maybe it already does the "expected" thing:) ).

If the command is executed in a package wc, should we also reset the local
wc state (see [2])?

[1] https://github.com/openSUSE/osc/issues/808#issuecomment-654921029
[2] https://github.com/openSUSE/osc/issues/808#issuecomment-643292689

lethliel:
@mcepl Perhaps you can clarify if this is the expected behavior.
At least it is what I understood from your script.

mcepl:
> On 2020-07-07 01:23:56 -0700, Marco Strigl wrote:
> If the command is executed in a package wc, should we also reset the local wc state (see [2])?

I would be completely OK if the command complained that the working directory is in “dirty” state (i.e., there are non-commited changes) and refused to continue. I believe some other subcommands (``sr``?) do exactly that.

dmach:

PR #825, opened 2020-07-21
print detailed error message in interactive review

lethliel:
Print the details of a Bad Request also in interactive review mode.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -7385,6 +7385,13 @@ def request_interactive_review(apiurl, request, initial_cmd='', group=None,
             return True
         except HTTPError as e:
             print('Server returned an error:', e, file=sys.stderr)
+            details = e.hdrs.get('X-Opensuse-Errorcode')
+            if details:
+                print(details, file=sys.stderr)
+            root = ET.fromstring(e.read())
+            summary = root.find('summary')
+            if summary is not None:
+                print(summary.text, file=sys.stderr)
             print('Try -f to force the state change', file=sys.stderr)
         return False
 


marcus-h:
On 2020-07-21 13:56:56 -0700, Marco Strigl wrote:
> index 9bd7b5b1..043a4b1d 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -7385,6 +7385,13 @@ def safe_change_request_state(*args, **kwargs):
>              return True
>          except HTTPError as e:
>              print('Server returned an error:', e, file=sys.stderr)
> +            details = e.hdrs.get('X-Opensuse-Errorcode')
> +            if details:
> +                print(details, file=sys.stderr)
> +            root = ET.fromstring(e.read())
> +            summary = root.find('summary')
> +            if not summary is None:

Minor: if summary is not None: ...

> +                print(summary.text)

Minor: what about printing it to stderr, too?

>              print('Try -f to force the state change', file=sys.stderr)
>          return False

marcus-h:
LGTM.
Can you squash the two commits into one?:)

lethliel:
@marcus-h Forgot to squash. (Note to self: Do note make code changes between cooking dinner for the family, cleaning and get the living room clean again)

marcus-h:
On 2020-07-22 18:21:54 +0000, Marco Strigl wrote:
> @marcus-h Forgot to squash.

Merged. Thanks!

> (Note to self: Do note make code changes between cooking dinner for the family, cleaning and get the living room clean again)
>
Sounds like "fun":D

PR #830, opened 2020-08-12
maintainer search: lookup via package name by default and binary as fallback

adrianschroeter:
This is faster in best case since the binary search does not need
to be executed on the server.

It also finds package names where no binary with that name exists.
(as for some multibuild cases)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,6 @@
+0.171
+  - maintainer search: lookup via package name by default and binary as fallback
+
 0.170
   - fix code for python3.8 and python3.9
   - remove dead code
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -8183,7 +8183,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         """${cmd_name}: Show maintainers according to server side configuration
 
             # Search for official maintained sources in OBS instance
-            osc maintainer BINARY <options>
+            osc maintainer BINARY_OR_PACKAGE_NAME <options>
             osc maintainer -U <user> <options>
             osc maintainer -G <group> <options>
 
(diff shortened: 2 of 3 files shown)


lethliel:
@adrianschroeter This PR has no content.

adrianschroeter:
@lethliel now there is

marcus-h:
On 2020-08-12 00:21:45 -0700, Adrian Schröter wrote:
> index e4738fd2..0fe88901 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -6809,12 +6809,13 @@ def search(apiurl, queries=None, **kwargs):
>          res[urlpath] = ET.parse(f).getroot()
>      return res
>
> -def owner(apiurl, binary, mode="binary", attribute=None, project=None, usefilter=None, devel=None, limit=None):
> +def owner(apiurl, search_string, mode="binary", attribute=None, project=None, usefilter=None, devel=None, limit=None):

Hmm this change makes sense but breaks osc's API/may break 3rd-party
users, which rely on keyword arguments, (if any...).
If we care, we could do something like

def owner(apiurl, search_string=None, mode="binary", attribute=None,
          project=None, usefilter=None, devel=None, limit=None, binary=None):
    # binary is just for API backward compatibility
    if not ((search_string is None) ^ (binary is None)):
        raise ValueError('Either specify search_string or binary')
    elif binary is not None:
        search_string = binary
    ...


(Maybe search_term might be better suited than search_string but that's
"just" naming...)

Otherwise: LGTM (despite the fact that it also "breaks" osc's UI (package
name vs. binary) but I can live with that:) )

This "depends" on [1].

[1] https://github.com/openSUSE/open-build-service/pull/9964

lethliel:
Thanks. Looks good now.

PR #831, opened 2020-08-18
fix content of osc workerinfo --help

JanZerebecki:
Replace usage with better explanation. It was missing that it requires a
prefixed hostarch. Also workername is instead called workerid in the
API.

Usage help was before: osc workerinfo WORKER

Add actual example.

See also the fix for this in OBS API docs:
openSUSE/open-build-service#10024

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -6079,9 +6079,11 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         """${cmd_name}: gets the information to a worker from the server
 
         Examples:
-            osc workerinfo <workername>
+            osc workerinfo x86_64:goat:1
 
-        ${cmd_usage}
+        Usage:
+            osc workerinfo <hostarch>:<workerid>
+
         ${cmd_option_list}
         """
         apiurl = self.get_api_url()


marcus-h:
On 2020-08-18 06:01:51 -0700, Jan Zerebecki wrote:
> Replace usage with better explanation. It was missing that it requires a
> prefixed hostarch. Also workername is instead called workerid in the
> API.
>
Depends a bit on your POV: workername := <hostarch>:<workerid> but your
patch makes it much more obvious!

LGTM (only a small comment/question (see below)).

> index fe8f7e66..31a477db 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -6079,9 +6079,11 @@ def do_workerinfo(self, subcmd, opts, worker):
>          """${cmd_name}: gets the information to a worker from the server
>
>          Examples:
> -            osc workerinfo <workername>
> +            osc workerinfo x86_64:goat01:1

Hmm what about omitting the numbers? Maybe something like "x86_64:goat"
or maybe "x86_64:my_worker".

What do you think?

> +
> +        Usage:
> +            osc workerinfo <hostarch>:<workerid>
>
> -        ${cmd_usage}
>          ${cmd_option_list}
>          """
>          apiurl = self.get_api_url()

JanZerebecki:
Adding the `:1` at the end of the example is important to know that just the arch and node name doesn't work. One needs to add the last number to specify one specific worker on the node. In both OBS installations it is always present and a number. It is part of the workerid.

How about: `x86_64:worker_node:1` ?

marcus-h:
On 2020-08-18 15:05:41 +0000, Jan Zerebecki wrote:
> Adding the `:1` at the end of the example is important to know that just the arch and node name doesn't work. One needs to add the last number to specify one specific worker on the node. In both OBS installations it is always present and a number. It is part of the workerid.
>
That's just the naming scheme that is used, for instance, in the build.o.o
instance. In general, "foobar", "foobar:123" etc. are valid worker ids.

> How about: `x86_64:worker_node:1` ?
>
I'm perfectly fine with this! I also like something "more concrete" like
your "x86_64:goat:1"...
Just take your pick - afterwards, it'll be merged:)

JanZerebecki:
Done

marcus-h:
Merged. Thanks a lot!

PR #833, opened 2020-08-27
Ignore signal.SIGWINCH.

### LINKED ISSUE #584: osc crashes on window resize
I'm "using osc-0.165.1-lp150.258.1.noarch" and observed some
weird behavior:

```
$ osc getbinaries images x86_64
```

During the progress go and resize the window. That ends with

```
   File "/usr/lib64/python3.6/site-packages/M2Crypto/SSL/Connection.py", line
386, in recv_into
   buflen = len(buff_bytes)
   TypeError: object of type 'NoneType' has no len()
```

I'm not pasting the entire stack trace that also shows where
the osc code is affected by this. I guess the reproducer is
easy. If that's not only on my system I think something very
weird is going on, probably related to the progress handling

Using the -q switch disables progress bar and workarounds the problem

Thanks

mcepl:
Fixes #584

@lethliel What’s wrong with something as simple as this?

### DIFF
--- a/osc/babysitter.py
+++ b/osc/babysitter.py
@@ -49,11 +49,16 @@ except ImportError:
 def catchterm(*args):
     raise oscerr.SignalInterrupt
 
+
+# Signals which should terminate the program safely
 for name in 'SIGBREAK', 'SIGHUP', 'SIGTERM':
     num = getattr(signal, name, None)
     if num:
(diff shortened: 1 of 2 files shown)


lethliel:
@mcepl nothing. Thanks for the fix ;-)

PR #834, opened 2020-09-01
catch URLs to github.com and gitlab.com as git repos

adrianschroeter:
yes, ugly, but I am tired to explain people to add a .git suffix
to get the right mode for "osc add"

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -4728,7 +4728,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
 
         # Do some magic here, when adding a url. We want that the server to download the tar ball and to verify it
         for arg in parseargs(args):
-            if arg.endswith('.git') or arg.startswith('git://') or arg.startswith('git@'):
+            if arg.endswith('.git') or arg.startswith('git://') or arg.startswith('git@') or arg.startswith('https://github.com') or arg.startswith('https://gitlab.com'):
                 addGitSource(arg)
             elif arg.startswith('http://') or arg.startswith('https://') or arg.startswith('ftp://'):
                 addDownloadUrlService(arg)


lethliel:
Yes. But I don't see any better way atm

marcus-h:
On 2020-09-01 04:52:59 -0700, Adrian Schröter wrote:
> yes, ugly, but I am tired to explain people to add a .git suffix
> to get the right mode for &quot;osc add&quot;

Hmm let's keep fingers crossed that you don't have to explain why
"osc add https://github.com/openSUSE/osc/archive/0.170.0.tar.gz" does not
work anymore:)
But I agree, it is probably sensible from a user's POV...

adrianschroeter:
On Dienstag, 1. September 2020, 14:49:32 CEST Marcus Hüwe wrote:
> On 2020-09-01 04:52:59 -0700, Adrian Schröter wrote:
> > yes, ugly, but I am tired to explain people to add a .git suffix
> > to get the right mode for &quot;osc add&quot;
> 
> Hmm let's keep fingers crossed that you don't have to explain why
> "osc add https://github.com/openSUSE/osc/archive/0.170.0.tar.gz" does not
> work anymore:)

uh, good hint...

we could extend the hack to be false if "/archive/" substring is 
inside to make it beautiful....

> But I agree, it is probably sensible from a user's POV...
> 
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/openSUSE/osc/pull/834#issuecomment-684827972
> 


-- 

lethliel:
Or check if the suffix is a valid archive?

marcus-h:
On 2020-09-01 13:20:19 +0000, Marco Strigl wrote:
> Or check if the suffix is a valid archive?
>
Hmm neither the suffix check nor the "archive" check would work for
the following gitlab url

https://gitlab.com/m2crypto/m2crypto/-/raw/master/setup.py?inline=false

(setup.py is probably a bad example but think of a spec file)

Another heuristic might be to check if the path of a github/gitlab url
consists of exactly two components (after normalization). If so, it
might be a "git" url.

(Personally, I can live with the status quo:) )

PR #839, opened 2020-09-11
Remove Python 2 compat code from osc

eliroca:
Already creating this to track progress and have a place where I can get feedback


marcus-h:
On 2020-09-11 08:03:02 -0700, Elisei Roca wrote:
> Already creating this to track progress and have a place where I can get feedback

I just had a very quick glance at it: I think most of the changes are
pretty reasonable but I'm not really convinced that we should actively
break python2 support. For instance, keeping stuff like

from __future__ import print_function

or "try: ... except: ImporError ..." does not really hurt (IMHO).

eliroca:
> On 2020-09-11 08:03:02 -0700, Elisei Roca wrote: Already creating this to track progress and have a place where I can get feedback
> I just had a very quick glance at it: I think most of the changes are pretty reasonable but I'm not really convinced that we should actively break python2 support. For instance, keeping stuff like from __future__ import print_function or "try: ... except: ImporError ..." does not really hurt (IMHO).

That is exactly what @lethliel assigned me to do :wink:

eliroca:
@lethliel the pull is ready for review, unless I'm overlooking something. :)

eliroca:
Package with these changes is out and ready to be installed and tested: https://build.opensuse.org/package/show/home:eroca:osc/osc

dmach:
Rather than resolving conflicts in this PR, I've made the changes in #1088 and #1089.
Thanks for the pointer to pyupgrade, it was really helpful!

Issue #201, opened 2016-05-20
'osc repourls' incorrectly assumes all repos are yum compatible


nickbroon:
The `repourls` command blindly assume all repositories of yum-style and contain a .repo file. This is not the case for Debian like repos, etc
As an example: `osc -A https://api.opensuse.org repourls openSUSE:Tools` Yet all the Debian and Ubuntu repositories don't contain .repo files.

It should either not output URL repos of these types, or output Debian APT compatible URLs.

It appears to just have a hard coded template, and iterate over the repositories subsutiting the project/repo names: https://github.com/openSUSE/osc/blob/master/osc/commandline.py#L4221

nickbroon:
This would definitely be very useful for being able to easily add repos to a debian system for automatically build packages.

nickbroon:
Any prospect of this being fixed?

It still outputs incorrect repo information for debian/ubuntu repos.

Correct output for a debian/ubuntu repo would be something like:
```
deb http://download.opensuse.org/repositories/openSUSE:/Tools/xUbuntu_20.04/ /
```

It currently outputs:
```
osc -A https://api.opensuse.org repourls openSUSE:Tools
https://download.opensuse.org/repositories/openSUSE:/Tools/xUbuntu_Next/openSUSE:Tools.repo
https://download.opensuse.org/repositories/openSUSE:/Tools/xUbuntu_20.04/openSUSE:Tools.repo
https://download.opensuse.org/repositories/openSUSE:/Tools/xUbuntu_19.10/openSUSE:Tools.repo
https://download.opensuse.org/repositories/openSUSE:/Tools/xUbuntu_19.04/openSUSE:Tools.repo
https://download.opensuse.org/repositories/openSUSE:/Tools/xUbuntu_18.10/openSUSE:Tools.repo
https://download.opensuse.org/repositories/openSUSE:/Tools/xUbuntu_18.04/openSUSE:Tools.repo
https://download.opensuse.org/repositories/openSUSE:/Tools/xUbuntu_17.04/openSUSE:Tools.repo
https://download.opensuse.org/repositories/openSUSE:/Tools/xUbuntu_16.10/openSUSE:Tools.repo
https://download.opensuse.org/repositories/openSUSE:/Tools/xUbuntu_16.04/openSUSE:Tools.repo
https://download.opensuse.org/repositories/openSUSE:/Tools/xUbuntu_15.10/openSUSE:Tools.repo
[...]

PR #841, opened 2020-09-17
add proper reporuls

lethliel:
for use in Arch, Debian and Ubuntu

Fixes https://github.com/openSUSE/osc/issues/201

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -4384,14 +4384,27 @@ Please submit there instead, or use --nodevelproject to force direct submission.
     def do_repourls(self, subcmd, opts, *args):
         """${cmd_name}: Shows URLs of .repo files
 
-        Shows URLs on which to access the project .repos files (yum-style
-        metadata) on download.opensuse.org.
+        Shows URLs on which to access the project repositories.
 
         usage:
            osc repourls [PROJECT]
 
         ${cmd_option_list}
         """
+        import tempfile
+
+        def _repo_type(apiurl, project,repo):
+            if not os.path.exists('/usr/lib/build/queryconfig'):
+                return None
+            build_config = get_buildconfig(apiurl, project, repo)
+            with tempfile.NamedTemporaryFile() as f:
+                f.write(build_config)
+                f.flush()
+                repo_type = return_external('/usr/lib/build/queryconfig', '--dist',
+                                        f.name, 'repotype').rstrip(b'\n')
+            if not repo_type:
+                return None
+            return decode_it(repo_type)
 
         apiurl = self.get_api_url()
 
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2020-09-17 02:33:08 -0700, Marco Strigl wrote:
> for use in Arch, Debian and Ubuntu
>
> This is not the best fix, but unless we have a API route to get the repo-type for a repository this is the best I can do now.
>
Hmm what about something like this

import tempfile


def do_repourls(...):
    def _repo_type(apiurl, project, repo):
        # hmm or a new function query_config(apiurl, project, repo, setting)
        # in core?
        if not os.path.exists('/usr/lib/build/queryconfig'):
            return None
        build_config = get_buildconfig(apiurl, project, repo)
        with tempfile.NamedTemporaryFile() as f:
            f.write(build_config)
            f.flush()
            repo_type = return_external('/usr/lib/build/queryconfig', '--dist',
                                        f.name, 'repotype')
            if not repo_type:
                # hmm returning None is not good - we should probably do the
                # same heuristic as bs_publish
                return None
            return decode_it(repo_type)
...
    for repo in repos:
        repo_type = _repo_type(apiurl, project, repo)
        if repo_type == 'rpm-md':
            ...
        elif repo_type == 'arch':
            ...
        else:
            # default or error out?


What do you think?

nickbroon:
Yes I think using some form of repo type information would be necessary.
Using the repo name as an approximation will not work in many of the use cases I have as the repos are often just 'standard' for debian, or sometimes a codes name like 'stretch', 'buster'.

lethliel:
@marcus-h 

Yes. This slows down the output a little bit, but I can live with it. 
The guessing of the repotype based on the names would work with OBS but the solution of Marcus is of course better.

marcus-h:
On 2020-09-18 04:33:57 -0700, Marco Strigl wrote:
> Yes. This slows down the output a little bit, but I can live with it.

Yes, definitely:/

> +        def _repo_type(apiurl, project,repo):
> +            if not os.path.exists('/usr/lib/build/queryconfig'):
> +                return None
> +            build_config = get_buildconfig(apiurl, project, repo)
> +            with tempfile.NamedTemporaryFile() as f:
> +                f.write(build_config)
> +                f.flush()
> +                repo_type = return_external('/usr/lib/build/queryconfig', '--dist',
> +                                        f.name, 'repotype')
> +            if repo_type is None:
> +                return None
> +            return decode_it(repo_type)
>
>          apiurl = self.get_api_url()
>
> @@ -4407,10 +4421,19 @@ def do_repourls(self, subcmd, opts, *args):
>          if elm is None or not elm.text:
>              raise oscerr.APIError('download_url configuration element expected')
>
> +        url_deb_tmpl = 'deb ' + elm.text + '/%s/%s/ /'
> +        url_arch_tmpl = 'Server=' + elm.text + '/%s/%s/$arch'
>          url_tmpl = elm.text + '/%s/%s/%s.repo'
>          repos = get_repositories_of_project(apiurl, project)
>          for repo in repos:
> -            print(url_tmpl % (project.replace(':', ':/'), repo, project))
> +            repo_type = (_repo_type(apiurl, project, repo)).rstrip('\n')

Good catch! A minor suggestion: let's move the strip to the _repo_type
function: return_external(...).rstrip('\n'). And then

if not repo_type:
    return None

(in _repo_type)

This "feels" a bit cleaner - but feel free to ignore it. Your current code
works:)

nickbroon:
Should probably also update the help string to remove 'yum-style' and also the 'opensuse.org'.

```
    def do_repourls(self, subcmd, opts, *args):
        """${cmd_name}: Shows URLs of .repo files
        Shows URLs on which to access the project .repos files (yum-style
        metadata) on download.opensuse.org.
        usage:
           osc repourls [PROJECT]
        ${cmd_option_list}
```

lethliel:
all done. Will merge now.

PR #843, opened 2020-09-19
Make python shebang explicit for python3 to allow building for Mageia Cauldron

zilti:


marcus-h:
We could do that but I'm not sure if we get (build) failures for "old"
distributions (like Debian 8/Jessie, Mageia 5 etc.) in the
openSUSE:Tools project. For instance, Jessie ships python34, which osc
does not support. (I could live with that, though...)

@lethliel any opinion?

PS: ideally, the commits should be squashed (and a short rationale in
the commit message would be nice) :)

zilti:
Okay. Noted. I think this pull request might not be necessary, since there seems to be a duplicate already in #640 (switch shebang of wrapper script to python3)

PR #844, opened 2020-09-28
maintenance_request: verify right target for superseding

adrianschroeter:
NOTE: this is more a bugreport then a pull request atm...

We need to check against releaseproject instead of targetproject for
maintenance_incident. Since target_project is almost always the one
maintenance project (eg. openSUSE:Maintenance).

WARNING: this relies on a new xpath query support which is not available
         yet! Also no compat code implemented yet.


adrianschroeter:
the xpath submission is finaly here (again?): https://github.com/openSUSE/open-build-service/pull/13735

dmach:

PR #846, opened 2020-09-30
Move service generated files to .old (fixes #845)

### LINKED ISSUE #845: move service generated files to .old instead of deleting them
Before running services, obs moves all generated files to a `.old` directory:
https://github.com/openSUSE/open-build-service/blob/66eedd78b4ca65439b96be271b04347cfc21f248/src/backend/bs_service#L135

while osc just deletes them:
https://github.com/openSUSE/osc/blob/7b5d10552352295ebdd21193b6dfbf830e7d236a/osc/core.py#L415

So I guess osc should be fixed to be have like the server side in that case.

lnussel:
fixes #845

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -412,14 +412,16 @@ class Serviceinfo:
     def execute(self, dir, callmode = None, singleservice = None, verbose = None):
         import tempfile
 
+        oldcreated = False
         # cleanup existing generated files
         for filename in os.listdir(dir):
             if filename.startswith('_service:') or filename.startswith('_service_'):
-                ent = os.path.join(dir, filename)
-                if os.path.isdir(ent):
-                    shutil.rmtree(ent)
-                else:
-                    os.unlink(ent)
+                if not oldcreated:
+                    if os.path.exists('.old'):
+                        raise oscerr.ServiceRuntimeError('The directory ".old" exists, please remove it')
+                    oldcreated = True
+                    os.mkdir('.old')
+                os.rename(filename, os.path.join('.old', filename))
 
         allservices = self.services or []
         service_names = [s['name'] for s in allservices]
@@ -494,6 +496,9 @@ class Serviceinfo:
             finally:
                 if temp_dir is not None:
                     shutil.rmtree(temp_dir)
+
+        if os.path.exists('.old'):
+            shutil.rmtree('.old')
 
         return 0
 


marcus-h:
On 2020-09-30 01:24:18 -0700, Ludwig Nussel wrote:
> index fd2c1a48..907256cf 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -415,11 +415,9 @@ def execute(self, dir, callmode = None, singleservice = None, verbose = None):
>          # cleanup existing generated files
>          for filename in os.listdir(dir):
>              if filename.startswith('_service:') or filename.startswith('_service_'):
> -                ent = os.path.join(dir, filename)
> -                if os.path.isdir(ent):
> -                    shutil.rmtree(ent)
> -                else:
> -                    os.unlink(ent)
> +                if not os.path.exists('.old'):
> +                    os.mkdir('.old')
> +                os.rename(filename, os.path.join('.old', filename))
>

Hmm the change makes sense but maybe we should error out if ".old" exists?
I'm a bit hesitant to remove a dir, which could have been created by a
user... What do you think?

adrianschroeter:
On Mittwoch, 30. September 2020, 20:10:36 CEST Marcus Hüwe wrote:
> On 2020-09-30 01:24:18 -0700, Ludwig Nussel wrote:
> > index fd2c1a48..907256cf 100644
> > --- a/osc/core.py
> > +++ b/osc/core.py
> > @@ -415,11 +415,9 @@ def execute(self, dir, callmode = None, singleservice = None, verbose = None):
> >          # cleanup existing generated files
> >          for filename in os.listdir(dir):
> >              if filename.startswith('_service:') or filename.startswith('_service_'):
> > -                ent = os.path.join(dir, filename)
> > -                if os.path.isdir(ent):
> > -                    shutil.rmtree(ent)
> > -                else:
> > -                    os.unlink(ent)
> > +                if not os.path.exists('.old'):
> > +                    os.mkdir('.old')
> > +                os.rename(filename, os.path.join('.old', filename))
> >
> 
> Hmm the change makes sense but maybe we should error out if ".old" exists?
> I'm a bit hesitant to remove a dir, which could have been created by a
> user... What do you think?

A different name would be good to avoid conflict with user created dirs, but
we would become incompatible with bs_service ...

There is no other way then just to remove it IMHO...

-- 

lnussel:
I've update the code to throw an exception if the directory was not created by the osc run itself

marcus-h:
On 2020-09-30 22:51:18 -0700, Adrian Schröter wrote:
> On Mittwoch, 30. September 2020, 20:10:36 CEST Marcus Hüwe wrote:

<SNIP>

> > Hmm the change makes sense but maybe we should error out if ".old" exists?
> > I'm a bit hesitant to remove a dir, which could have been created by a
> > user... What do you think?
>
> A different name would be good to avoid conflict with user created dirs, but
> we would become incompatible with bs_service ...
>
Ideally, the services would support an "--old-dir /path/to/old/dir" option.

marcus-h:
Sigh... github ignores some mails again...
```

Date: Thu, 1 Oct 2020 13:46:06 +0200
To: reply+AAIEUTBFAPXNRDYWOAO6AFV5QF7SREVBNHHCU3CO44@reply.github.com
Subject: Re: [openSUSE/osc] Move service generated files to .old (fixes #845)
 (#846)

On 2020-10-01 09:15:20 +0000, Ludwig Nussel wrote:
> I've updated the code to throw an exception if the directory was not created by the osc run itself
> 
Thanks! From a usability POV, we should make sure that we always cleanup
the old dir (for instance, if a service fails). What about the following?

```
diff --git a/osc/core.py b/osc/core.py
index fd2c1a4..f6d1e2c 100644
--- a/osc/core.py
+++ b/osc/core.py
@@ -410,16 +410,27 @@ class Serviceinfo:
         return r

     def execute(self, dir, callmode = None, singleservice = None, verbose = None):
+        old_dir = os.path.join(dir, '.old')
+        if os.path.exists(old_dir) or os.path.islink(old_dir):
+            msg = '"%s" exists, please remove it' % old_dir
+            raise oscerr.OscIOError(None, msg)
+        try:
+            os.mkdir(old_dir)
+            return self._execute(dir, old_dir, callmode, singleservice,
+                                 verbose)
+        finally:
+            if os.path.isdir(old_dir):
+                shutil.rmtree(old_dir)
+
+    def _execute(self, dir, old_dir, callmode=None, singleservice=None,
+                 verbose=None):
         import tempfile

         # cleanup existing generated files
         for filename in os.listdir(dir):
             if filename.startswith('_service:') or filename.startswith('_service_'):
-                ent = os.path.join(dir, filename)
-                if os.path.isdir(ent):
-                    shutil.rmtree(ent)
-                else:
-                    os.unlink(ent)
+                os.rename(os.path.join(dir, filename),
+                          os.path.join(old_dir, filename))

         allservices = self.services or []
         service_names = [s['name'] for s in allservices]
```

```

lnussel:
feel free to throw away my code and put your own :)

marcus-h:
On 2020-10-01 09:15:20 +0000, Ludwig Nussel wrote:
> I've updated the code to throw an exception if the directory was not created by the osc run itself
>
Thanks! From a usability POV, we should make sure that we always cleanup
the old dir (for instance, if a service fails). What about the following?

```
diff --git a/osc/core.py b/osc/core.py
index fd2c1a4..f6d1e2c 100644
--- a/osc/core.py
+++ b/osc/core.py
@@ -410,16 +410,27 @@ class Serviceinfo:
         return r

     def execute(self, dir, callmode = None, singleservice = None, verbose = None):
+        old_dir = os.path.join(dir, '.old')
+        if os.path.exists(old_dir) or os.path.islink(old_dir):
+            msg = '"%s" exists, please remove it' % old_dir
+            raise oscerr.OscIOError(None, msg)
+        try:
+            os.mkdir(old_dir)
+            return self._execute(dir, old_dir, callmode, singleservice,
+                                 verbose)
+        finally:
+            if os.path.isdir(old_dir):
+                shutil.rmtree(old_dir)
+
+    def _execute(self, dir, old_dir, callmode=None, singleservice=None,
+                 verbose=None):
         import tempfile

         # cleanup existing generated files
         for filename in os.listdir(dir):
             if filename.startswith('_service:') or filename.startswith('_service_'):
-                ent = os.path.join(dir, filename)
-                if os.path.isdir(ent):
-                    shutil.rmtree(ent)
-                else:
-                    os.unlink(ent)
+                os.rename(os.path.join(dir, filename),
+                          os.path.join(old_dir, filename))

         allservices = self.services or []
         service_names = [s['name'] for s in allservices]
```

marcus-h:
On 2020-10-01 04:56:14 -0700, Ludwig Nussel wrote:
> feel free to throw away my code and put your own :)
>
I just created https://github.com/openSUSE/osc/pull/847

PS: Sorry for being too stupid to post a patch via github...
PPS: eventually, my original mail got through... (after a considerable
delay:) )

PR #847, opened 2020-10-01
Add .old dir support for source services

marcus-h:
Some services expect "old" service files (that is, files from a
previous service run) to be present in an ".old" dir. Hence, osc
should support that.
Instead of removing all files from a previous service run, move them
to the ".old" dir, run the services, and, finally, remove the ".old"
dir.
Unfortunately, the location of the ".old" dir is hardcoded in the
specific services. That is, we have to be careful if an ".old" dir
exists (in this case, we error out).

Based on [1].

[1] https://github.com/openSUSE/osc/pull/846

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -410,16 +410,27 @@ class Serviceinfo:
         return r
 
     def execute(self, dir, callmode = None, singleservice = None, verbose = None):
+        old_dir = os.path.join(dir, '.old')
+        if os.path.exists(old_dir) or os.path.islink(old_dir):
+            msg = '"%s" exists, please remove it' % old_dir
+            raise oscerr.OscIOError(None, msg)
+        try:
+            os.mkdir(old_dir)
+            return self._execute(dir, old_dir, callmode, singleservice,
+                                 verbose)
+        finally:
+            if os.path.exists(old_dir):
+                shutil.rmtree(old_dir)
+
+    def _execute(self, dir, old_dir, callmode=None, singleservice=None,
+                 verbose=None):
         import tempfile
 
         # cleanup existing generated files
         for filename in os.listdir(dir):
             if filename.startswith('_service:') or filename.startswith('_service_'):
-                ent = os.path.join(dir, filename)
-                if os.path.isdir(ent):
-                    shutil.rmtree(ent)
-                else:
-                    os.unlink(ent)
+                os.rename(os.path.join(dir, filename),
+                          os.path.join(old_dir, filename))
 
         allservices = self.services or []
         service_names = [s['name'] for s in allservices]


lnussel:
lgtm

adrianschroeter:
that means it will break on the second run because .old is already there?

That would not be very userfriendly IMHO.

adrianschroeter:
ah, you remove it always again after each run.... ignore that then

adrianschroeter:
but it will not be easy to have a version with server side version to test again.

Because it will always complain that .old is there.

Would be useful for eg. changelog generators.

lnussel:
Not sure what you are talking about. The code is only executed when running `osc service run`. Eg `osc up -S` which would download the server side stored sources is not affected by that. A `.old` directory cannot be part of a checkout, it doesn't exist on server side.

adrianschroeter:
On Donnerstag, 1. Oktober 2020, 15:32:53 CEST Ludwig Nussel wrote:
> Not sure what you are talking about. The code is only executed when running `osc service run`. Eg `osc up -S` which would download the server side stored sources is not affected by that. A `.old` directory cannot be part of a checkout, it doesn't exist on server side.

and also before each build. And you may run multiple iterations without commiting.

-- 

lnussel:
So far osc just deletes all service generated files. Does keeping them around in .old for just a little bit longer make the situation worse than it already is?

marcus-h:
On 2020-10-01 15:42:24 +0000, Adrian Schröter wrote:
> On Donnerstag, 1. Oktober 2020, 15:32:53 CEST Ludwig Nussel wrote:
> > Not sure what you are talking about. The code is only executed when running `osc service run`. Eg `osc up -S` which would download the server side stored sources is not affected by that. A `.old` directory cannot be part of a checkout, it doesn't exist on server side.
>
> and also before each build. And you may run multiple iterations without commiting.
>
Hmm I don't see the problem with multiple service runs. Example:

Step 1: initially, no _service:* files are present in the package wc
- "osc service run" creates some _service:* files

Step 2: run "osc service run"
- create .old dir (or complain if it exists)
- move all _service:* files into .old
- execute services (here, a service could potentially access the
  old service files)
- remove .old dir

Step 3: run "osc service run"
- goto Step 2

Do I miss something?

The only problem I see is that we might "skip" some files that were
created by a service but have no _service: prefix (depending on the
callmode, no _service:* prefix is added to the filename). This could
be fixed. However, the question is which behavior would a user expect?
(We should probably clarify this before merging this PR:) )

adrianschroeter:
On Donnerstag, 1. Oktober 2020, 22:20:54 CEST Marcus Hüwe wrote:
> On 2020-10-01 15:42:24 +0000, Adrian Schröter wrote:
> > On Donnerstag, 1. Oktober 2020, 15:32:53 CEST Ludwig Nussel wrote:
> > > Not sure what you are talking about. The code is only executed when running `osc service run`. Eg `osc up -S` which would download the server side stored sources is not affected by that. A `.old` directory cannot be part of a checkout, it doesn't exist on server side.
> >
> > and also before each build. And you may run multiple iterations without commiting.
> >
> Hmm I don't see the problem with multiple service runs. Example:
> 
> Step 1: initially, no _service:* files are present in the package wc
> - "osc service run" creates some _service:* files
> 
> Step 2: run "osc service run"
> - create .old dir (or complain if it exists)
> - move all _service:* files into .old
> - execute services (here, a service could potentially access the
>   old service files)
> - remove .old dir
> 
> Step 3: run "osc service run"
> - goto Step 2
> 
> Do I miss something?

that is all fine and please do not see my comments as a blocker here,
since the code is doing an improvement.

I just wanted to point out to situations where the service is
generating content based on the diff to the last run.

And here the version on the server would be more important then
the version on the developer workstation from last attempt.

It is not possible to use/test such services then, if an existing
.old leads always to an error.

But we can handle this via an extra CLI option later on maybe...
 
> The only problem I see is that we might "skip" some files that were
> created by a service but have no _service: prefix (depending on the
> callmode, no _service:* prefix is added to the filename). This could
> be fixed. However, the question is which behavior would a user expect?
> (We should probably clarify this before merging this PR:) )

yes, indeed.

I was more speaking about _service:* files generated on server side.

-- 

lnussel:
so can we merge this?

PR #848, opened 2020-10-05
show repository state and details. It can be broken and the user must…

adrianschroeter:
… be able to see the reason.

That was always a problem, but with the new option for disabled publishing
due to package failures it becomes more critical

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -5788,7 +5788,7 @@ def result_xml_to_dicts(xml):
         rmap['repostate'] = node.get('code')
         rmap['pkg'] = rmap['package'] = rmap['pac'] = ''
         rmap['code'] = node.get('code')
-        rmap['details'] = ''
+        rmap['details'] = node.get('details')
         # the way we currently use this function, there should be
         # always a status element
         snodes = node.findall('status')
@@ -5809,6 +5809,10 @@ def result_xml_to_dicts(xml):
             details = statusnode.find('details')
             if details is not None:
                 smap['details'] = details.text
+            if rmap['code'] == 'broken':
+                # real error just becomes visible in details/verbose
+                smap['code'] = rmap['code']
+                smap['details'] = "repository: " + rmap['details']
             yield smap, is_multi
 
 
@@ -5951,6 +5955,8 @@ def get_prj_results(apiurl, prj, hide_legend=False, csv=False, status_filter=Non
             state = "outdated"
         else:
             state = node.get('state')
+        if node.get('details'):
+            state += ' details: ' + node.get('details')
         tg = (node.get('repository'), node.get('arch'), state)
         targets.append(tg)
         for pacnode in node.findall('status'):


marcus-h:
On 2020-10-05 01:16:04 -0700, Adrian Schröter wrote:
> index fd2c1a48..d9cb3305 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -5770,6 +5770,9 @@ def result_xml_to_dicts(xml):
>              details = statusnode.find('details')
>              if details is not None:
>                  smap['details'] = details.text
> +            if rmap['code'] == 'broken':
> +                smap['code'] = smap['code'] + ' ' if smap['code'] else ''
> +                smap['code'] += ' repository:' + rmap['state']

Hmm this breaks all code that expects smap['code'] to be one of
our "fixed", "predefined" states ("succeeded", "broken", "failed" etc.).

>              yield smap, is_multi
>
>
> @@ -5912,6 +5915,8 @@ def get_prj_results(apiurl, prj, hide_legend=False, csv=False, status_filter=Non
>              state = "outdated"
>          else:
>              state = node.get('state')
> +        if node.get('details'):
> +            state += ' details: ' + node.get('details')

As above - this would break, for instance, the status filter code.

Maybe a stupid question: why do we use "details" here while the code
from above uses "state"?

>          tg = (node.get('repository'), node.get('arch'), state)
>          targets.append(tg)
>          for pacnode in node.findall('status'):
>

marcus-h:
Hmm github eats my mails again...
```

Date: Wed, 15 Dec 2021 23:02:40 +0100
To: reply+AAIEUTBCE5MCKU27TA4XPBV5Q23UJEVBNHHCVGCO4A@reply.github.com
Subject: Re: [openSUSE/osc] show repository state and details. It can be
 broken and the user must… (#848)

On 2021-12-13 04:38:15 -0800, Adrian Schröter wrote:
> index 8732ed11..428fc740 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -5788,7 +5788,7 @@ def result_xml_to_dicts(xml):
>          rmap['repostate'] = node.get('code')
>          rmap['pkg'] = rmap['package'] = rmap['pac'] = ''
>          rmap['code'] = node.get('code')
> -        rmap['details'] = ''
> +        rmap['details'] = node.get('details')

What about storing it in rmap['repodetails']?


>          # the way we currently use this function, there should be
>          # always a status element
>          snodes = node.findall('status')
> @@ -5809,6 +5809,10 @@ def result_xml_to_dicts(xml):
>              details = statusnode.find('details')
>              if details is not None:
>                  smap['details'] = details.text
> +            if rmap['code'] == 'broken':
> +                # real error just becomes visible in details/verbose
> +                smap['code'] = rmap['code']
> +                smap['details'] = "repository: " + rmap['details']

Hmm what about moving this code into get_results (there we can
use the 'repostate' and 'repodetails' keys)?

>              yield smap, is_multi
> 
> 
> @@ -5951,6 +5955,8 @@ def get_prj_results(apiurl, prj, hide_legend=False, csv=False, status_filter=Non
>              state = "outdated"
>          else:
>              state = node.get('state')
> +        if node.get('details'):
> +            state += ' details: ' + node.get('details')

Hmm in my first review I wrote that this breaks the status filtering
code (in get_prj_results) but it seems I was wrong... (I'll have a
closer look at it again - the whole get_prj_results function is a mess...).

>          tg = (node.get('repository'), node.get('arch'), state)
>          targets.append(tg)
>          for pacnode in node.findall('status'):

```

Issue #849, opened 2020-10-05
superseding a "multiple" sr with `-s` submits but produces python Traceback and does not supersed old one:


bnavigator:
I got this error when submitting SR 839609 to supersede 839591:

```
greiner@greinerT450s:~/src/osc/home:bnavigator:branches:devel:languages:python/ > osc up
Updating python-poetry
At revision 0a4280304cd2305c4b05b2dde0bce711.
Updating python-tomlkit
At revision 6235067796326b8eda6f05ae31207ea6.
Updating python-poetry-core
At revision c2b8345365f39e282c684c7bf681271c.
Pending requests for project: home:bnavigator:branches:devel:languages:python (1)
839591  State:new        By:bnavigator   When:2020-10-05T14:24:38
        submit:          home:bnavigator:branches:devel:languages:python/python-poetry@c2fcc0640a24aad42f4a283e51a969c0 -> devel:languages:python
        submit:          home:bnavigator:branches:devel:languages:python/python-poetry-core@c2b8345365f39e282c684c7bf681271c -> devel:languages:python
        submit:          home:bnavigator:branches:devel:languages:python/python-tomlkit@379d26e5d529b95d8997e978070dc9bc -> devel:languages:python
        Descr: update poetry and it's deps: requires new poetry-core
 

greiner@greinerT450s:~/src/osc/home:bnavigator:branches:devel:languages:python/ > osc sr devel:languages:python -m "update poetry and it's deps: requires new poetry-core"  -s 839591
Request(s) created:  839609 Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.8/site-packages/osc/babysitter.py", line 64, in run
    return prg.main(argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 347, in main
    return self.cmd(args)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 370, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 504, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 1235, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.8/site-packages/osc/commandline.py", line 1333, in do_submitrequest
    'superseded by %s' % result, result)
[...]

PR #850, opened 2020-10-05
use sr_ids[0] for superseding

lethliel:
result is not assigned in the case of many packages per request

fixes https://github.com/openSUSE/osc/issues/849

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -1329,7 +1329,7 @@ class Osc(cmdln.Cmdln):
             if len(myreqs) > 0:
                 for req in myreqs:
                     change_request_state(apiurl, str(req), 'superseded',
-                                             'superseded by %s' % result, result)
+                                             'superseded by %s' % sr_ids[0], sr_ids[0])
 
             sys.exit('Successfully finished')
 


marcus-h:
On 2020-10-05 17:55:38 +0000, Marco Strigl wrote:
> index 5ebcaf6e..370e8518 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -1329,7 +1329,7 @@ def _check_service(root):
>              if len(myreqs) > 0:

Hmm we should probably change this to something like

if len(myreqs) > 0 and sr_ids:

or more pythonic:

if myreqs and sr_ids:

because sr_ids could be empty (in the --separate-requests case).

>                  for req in myreqs:
>                      change_request_state(apiurl, str(req), 'superseded',
> -                                             'superseded by %s' % result, result)
> +                                             'superseded by %s' % sr_ids[0], sr_ids[0])
>
Hmm I'm not sure if sr_ids[0] is really correct in the --separate-requests
case. Maybe we need some special handling for the --separate-requests case
in the future...

bnavigator:
The block directly above doesn't look right either:

https://github.com/openSUSE/osc/blob/8085aa6012dceb20235335ce67f1ead689da8971/osc/commandline.py#L1315-L1327

Shouldn't it print the content of `value` for the prompt? `myreqs` only contains the value of a given `-s`. And `value` is added to `myreqs` as item of a list. Is this always a single item or are there multiple possible? Either the text ("are") should be changed or the list extension.

bnavigator:
> Hmm I'm not sure if sr_ids[0] is really correct in the --separate-requests case. Maybe we need some special handling for the --separate-requests case in the future...

In that case you need a detection for package to package mapping and handling for multiple `-s` parameters. Right now there is no automatic detection at all for a multiple request.

aaptel:
I ran into this bug today again. Can we merge this and open another issue to deal with that edge case?

PR #851, opened 2020-10-07
Add default timeout to http calls

andrii-suse:


marcus-h:
On 2020-10-07 03:59:31 -0700, Andrii Nikitin wrote:
> -- Commit Summary --
>
>   * Add default timeout to http calls
>
Hmm...
Why? Why 60? What are you trying to achieve?

andrii-suse:
By default http_request has unlimited timeout, which means that short network outage can leave a request hanging indefinitely.
I believe I did experience exactly this today (though I didn't prove source of hanging, just did conclude the reason empirically, based on previous experience with python).
So any value for timeout will be better than default.
Next step could be adding corresponding parameter to oscrc.

marcus-h:
On 2020-10-07 12:58:45 -0700, Andrii Nikitin wrote:
> By default http_request has unlimited timeout, which means that short network outage can leave a request hanging indefinitely.
> I believe I did experience exactly this today (though I didn't prove source of hanging, just did conclude the reason empirically, based on previous experience with python).
> So any value for timeout will be better than default.

Hmm this depends on the POV. If you run, for instance, "osc rbl ...",
which produces no output for more than 60 seconds, you will end up in
a timeout. Nevertheless, a timeout might be convenient for 3rd party
users (who use osc as a library).

> Next step could be adding corresponding parameter to oscrc.
>
That's something I could live with:) However, there is a small issue
with the current patch (see below).

> index fd2c1a48..bdae6b88 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -3326,7 +3326,7 @@ def makeurl(baseurl, l, query=[]):
>      return urlunsplit((scheme, netloc, '/'.join([path] + list(l)), query, ''))
>
>
> -def http_request(method, url, headers={}, data=None, file=None):
> +def http_request(method, url, headers={}, data=None, file=None, timeout=60):
>      """wrapper around urllib2.urlopen for error handling,
>      and to support additional (PUT, DELETE) methods"""
>      def create_memoryview(obj):
> @@ -3396,7 +3396,7 @@ def create_memoryview(obj):
>      try:
>          if isinstance(data, str):
>              data = bytes(data, "utf-8")
> -        fd = urlopen(req, data=data)
> +        fd = urlopen(req, data=data, timeout=timeout)

The passed timeout is currently neither considered by m2crypto nor by
our m2crypto specific subclasses:/ That's (probably) why we removed it
several years ago (see commit d04c36189e1bdbbe99fb877030a05b67019a8a9b,
which is a perfect example for a miserable commit message...).

andrii-suse:
Indefinite timeout doesn't depend on POV - it is plain wrong on practice (at least I don't see any reason why it can be reasonable).
But I agree that this PR is not ready and since I don't plan to fix it properly at the moment: I am ok if the PR is just closed.

marcus-h:
On 2020-10-12 02:49:44 -0700, Andrii Nikitin wrote:
> Indefinite timeout doesn't depend on POV - it is plain wrong on practice (at least I don't see any reason why it can be reasonable).

Well, for every configured timeout, one could construct a spec file such
that "osc rbl prj pkg repo arch" runs into a timeout while a build is
still running with this specific spec file.
We could use the SO_KEEPALIVE socket option. Since this sends a "probe"
on the TCP level, we would detect your "broken network" use case after
some (considerable/configurable) amount of time.

PR #856, opened 2020-10-14
Use verifymd5 to determine freshness

bmwiedemann:
Use `verifymd5` to determine freshness
to speed up `osc up` for large repo checkouts by a factor of 10 or more

The verifymd5 value is not (yet) stored
so we quickly compute it as needed

Fixes https://github.com/openSUSE/open-build-service/issues/10250

In my test, it is now possible to finish `osc up` of `openSUSE:Factory` across the Atlantic (165ms RTT) in just 93 seconds.

Note: Needs testing with OBS services

could also profit from a review by @mlschroe or @adrianschroeter

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -1244,6 +1244,13 @@ class Package:
                 self.in_conflict.remove(fname)
                 self.write_conflictlist()
 
+    def _compute_verifymd5(self):
+        md5str = ''
+        for f in self.filelist:
+            if f.name != '_link':
+                md5str += f.md5+'  '+f.name+"\n"
+        return hashlib.md5(md5str.encode('utf-8')).hexdigest()
+
     def info(self):
         source_url = makeurl(self.apiurl, ['source', self.prjname, self.name])
         r = info_templ % (self.prjname, self.name, self.absdir, self.apiurl, source_url, self.srcmd5, self.rev, self.linkinfo)
@@ -2254,6 +2261,8 @@ rev: %s
     def update_needed(self, sinfo):
         # this method might return a false-positive (that is a True is returned,
         # even though no update is needed) (for details, see comments below)
+        if sinfo.get('verifymd5') == self._compute_verifymd5():
+            return False
         if self.islink():
             if self.isexpanded():
                 # check if both revs point to the same expanded sources


marcus-h:
First of all, thanks for patch!

On 2020-10-14 12:25:23 -0700, Bernhard M. Wiedemann wrote:
> Use `verifymd5` to determine freshness
> to speed up `osc up` for large repo checkouts by a factor of 10 or more
>
> The verifymd5 value is not (yet) stored
> so we quickly compute it as needed
>
> Fixes https://github.com/openSUSE/open-build-service/issues/10250
>
> In my test, it is now possible to finish `osc up` of `openSUSE:Factory` across the Atlantic (165ms RTT) in just 93 seconds.
>
> Note: Needs testing with OBS services
>
Actually, you tested (one of) the OBS service scenarios: openSUSE:Factory
is a project without _link files but some packages have a _service file
(where all services have mode=disabled). For all these packages,
Package.update_needed always returns True (since the sourceinfo does
not provide information about the lsrcmd5). Moreover, since for all
these packages, the services are disabled, the verifymd5 from the
sourceinfo matches the locally computed verifymd5.

Just to ensure that we are both on the same track: can you please
try the following?

Without your patch:

$> cd openSUSE:Factory
$> osc up -S                        # will probably take same time
$> time osc -H up -S 2>without_patch.txt

Now, with your patch

$> cd openSUSE:Factory
$> osc up                           # will probably take same time
$> time osc -H up 2>with_patch.txt

The times should not differ too much. Additionally, there should
be almost no diff if you run "diff -u with_patch.txt without_patch.txt"
(except for the initial sourceinfo GETs or for packages which really
changed in between).

> index fd2c1a48..e8d4082a 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -1244,6 +1244,13 @@ def wc_repair(self, apiurl=None):
>                  self.in_conflict.remove(fname)
>                  self.write_conflictlist()
>
> +    def getverifymd5(self):

Hmm maybe call this _calculate_verifymd5 or _compute_verifymd5 (at least
it should start with an underscore to indicate that it is not part of the
"public" API (see below)).

> +        md5str = ''
> +        for f in self.filelist:
> +            if f.name != '_link':

We should not skip a _link file. If a _link file exists locally, the
verifymd5s should mismatch.

> +                md5str += f.md5+'  '+f.name+"\n"
> +        return hashlib.md5(md5str.encode('utf-8')).hexdigest()
> +

That is, we could simply do something like this

m = hashlib.md5()
for f in self.filelist:
    m.update(('%s  %s\n' % (f.md5, f.name)).encode('utf-8')
return m.hexdigest()


However, there is IMHO one major drawback with this approach: the
computation of the verifymd5 is IMHO an implementation detail of the
backend. We should avoid that (if possible).

That's why I would like to try to revive [1], [2].

Would this be OK for you?


[1] https://lists.opensuse.org/opensuse-buildservice/2014-06/msg00067.html
[2] https://lists.opensuse.org/opensuse-buildservice/2014-07/msg00079.html

bmwiedemann:
I tested the timing with my smaller `home:bmwiedemann:reproducible` repo:
```
real    0m22.387s
user    0m0.663s
sys     0m0.033s


real    0m6.137s
user    0m0.398s
sys     0m0.032s
```

> The times should not differ too much

It seems, it does differ a lot, because unpatched osc does 2-3 API requests per unchanged package.
```bash
wc with*_patch.txt
  66   66 2554 without_patch.txt
  15   15  828 with_patch.txt
```

If you can get a similar effect in a cleaner way (e.g. with your OBS patch), I'm fine with that. I just really like to be able to finish an `osc up` in a minute and not 2 days.

dmach:
I've tested this code several times and it always worked and it usually improved performance a little bit.
Let's merge it and see what happens :)

marcus-h:
On 2022-04-07 08:49:06 +0000, Daniel Mach wrote:
> I've tested this code several times and it always worked and it usually improved performance a little bit.
> Let's merge it and see what happens :)
>
Can we please stop merging broken/incomplete stuff?:/

$> osc branch openSUSE:Tools osc home:Marcus_H:verifymd5_test
...
$> osc co -u home:Marcus_H:verifymd5_test
...
$> cd home:Marcus_H:verifymd5_test
$> ls osc/_link
osc/_link
$> osc api -X DELETE /source/home:Marcus_H:verifymd5_test/osc/_link
...
$> osc up -u
...
$> ls osc/_link
osc/_link
$>

This is wrong. Can we please revert it?

dmach:
> Can we please stop merging broken/incomplete stuff?:/

The pull-request was sitting here for 18 months with no feedback that it doesn't work or breaks anything.
I did some smoke-testing and it worked for me, unfortunately I did not catch it.
We definitely need to increase our test coverage and run the tests on every pull request.

marcus-h:
Hmm github seems to ignore my mail(s) again...
```
On 2022-04-11 08:15:41 +0000, Daniel Mach wrote:
> > Can we please stop merging broken/incomplete stuff?:/
>
> The pull-request was sitting here for 18 months with no feedback that it doesn't work or breaks anything.

It is mentioned in [1] but it might be a bit buried in my wrong assumptions:)

Another effect of this patch is that after an "osc up -S", for instance, in a
project like openSUSE:Factory, "osc info" might report the "wrong" srcmd5, which
"could" be confusing (this can happen, for instance, if there is a _service file
where all services are disabled (and there is no _link file), then the file sets
of the expanded and unexpanded rev are the same but the revision identifiers (the
md5) differ). Whether this is a real issue is debatable, though.

> I did some smoke-testing and it worked for me, unfortunately I did not catch it.
> We definitely need to increase our test coverage and run the tests on every pull request.

[1] https://github.com/openSUSE/osc/pull/856#issuecomment-1094690076
```

dmach:
I spoke to @adrianschroeter and @mlschroe yesterday and they confirmed that reverting is the only option.
Code from this PR optimizes only certain use case, leaving others broken.

Issue #865, opened 2020-11-19
Traceback in osc/util/cpio.py line 128: TypeError: Can't mix strings and bytes in path components


mcepl:
Running a regular `osc build` with osc-0.171.0-310.1.noarch I got this backtrace:
```
…
fetching packages for 'home:mcepl:branches:OBS_Maintained:python-pip': | Elapsed Time: 0:00:00                                                                                            
Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.8/site-packages/osc/babysitter.py", line 69, in run
    return prg.main(argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 347, in main
    return self.cmd(args)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 370, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 504, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 1235, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.8/site-packages/osc/commandline.py", line 6658, in do_build
    return osc.build.main(self.get_api_url(), opts, args)
  File "/usr/lib/python3.8/site-packages/osc/build.py", line 1005, in main
    fetcher.run(bi)
  File "/usr/lib/python3.8/site-packages/osc/fetch.py", line 253, in run
    self.__fetch_cpio(buildinfo.apiurl)
  File "/usr/lib/python3.8/site-packages/osc/fetch.py", line 125, in __fetch_cpio
    self.__download_cpio_archive(apiurl, project, repo, arch, package, **pkgs)
  File "/usr/lib/python3.8/site-packages/osc/fetch.py", line 74, in __download_cpio_archive
    archive.copyin_file(hdr.filename)
  File "/usr/lib/python3.8/site-packages/osc/util/cpio.py", line 187, in copyin_file
    self._copyin_file(hdr, dest, fn)
  File "/usr/lib/python3.8/site-packages/osc/util/cpio.py", line 128, in _copyin_file
    fn = os.path.join(dest, fn)
  File "/usr/lib64/python3.8/posixpath.py", line 90, in join
    genericpath._check_arg_types('join', a, *p)
[...]

PR #866, opened 2020-11-20
Fix a potential TypeError in CpioRead.copyin and CpioRead.copyin_file

marcus-h:
If no "dest" argument is specified when calling CpioRead.copyin or
CpioRead.copyin_file, a TypeError occurs in CpioRead._copyin_file
because os.getcwd(), which returns a str, is used as dest and, hence,
the subsequent os.path.join(...) fails (because it tries to join a
str and a bytes).
In order to avoid this, encode the result of os.getcwd().

Note that the existing

archive.copyin_file(hdr.filename,
                    os.path.dirname(tmpfile),
                    os.path.basename(tmpfile))

was OK because CpioRead._copyin_file os.path.join()s "dest" and
"new_fn", which are both str. It is just changed to stress that
CpioRead is a bytes-only API.

Fixes: #865 ("Traceback in osc/util/cpio.py line 128: TypeError:
Can't mix strings and bytes in path components")

### DIFF
--- a/osc/fetch.py
+++ b/osc/fetch.py
@@ -90,8 +90,8 @@ class Fetcher:
                     try:
                         fd, tmpfile = tempfile.mkstemp(prefix='osc_build_file')
                         archive.copyin_file(hdr.filename,
-                                            os.path.dirname(tmpfile),
-                                            os.path.basename(tmpfile))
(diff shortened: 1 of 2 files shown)


lethliel:
LGTM

lethliel:
And thanks.. :-)

PR #885, opened 2020-12-28
CI pipeline: Skip cleanup prior to deployment

### LINKED ISSUE #884: osc pypi artifact is not up to date
👋  looks like the latest release in here is `v0.171.0`, but the pypi artifact is still with `v0.167.2`.

crazyscientist:
Deployment via Travis CI failed due an error during cleanup (fixes #884):

`Could not restore untracked files from stash entry`

Disabling the cleanup fixes this problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ before_deploy:
   - rm -rf $TRAVIS_BUILD_DIR/dist/*
 deploy:
   provider: pypi
+  skip_existing: true
+  skip_cleanup: true
   user: suse
   password:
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2020-12-28 01:26:33 -0800, Andreas Hasenkopf wrote:
> Deployment via Travis CI failed due an error during cleanup (#884):
>
> `Could not restore untracked files from stash entry`
>
> Disabling the cleanup fixes this problem.

Cool! I just merged it even if I have no clue about travis or its automatic
pypi deployment (because the current state does not seem to work anyway).

Thanks a lot for the patch! Much appreciated!

PR #887, opened 2021-01-05
Add 'stage' options to 'osc build'

e4t:
The stage options can be used to start or stop the build with
a specific stage or build only a single stage.
For example for RPM builds this is useful to avoid a potentially
lengthy %build stage to test changes in the %install or %files
stage only.
Of course this works only if the previous stages have been built
during a prior call and have not been deleted, yet.
The stages are named according to the stage names of the underlying
packaging system (for rpm these are -bp, -bc, -bi, -bb, -bs and
-bl).
To build all stages up to the specified stage: --stage-pre=<stage>,
to build only the specified stage: --stage-single=<stage>,
to build all stages starting from the specified stage:
--stage-pre=<stage>.
Whether this option is supported depends on the underlying build
and packaging system. Presently, it is only supported for RPM
builds.
Additional changes to 'obs-build' are required as well.

Signed-off-by: Egbert Eich <eich@suse.com>

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -687,6 +687,18 @@ def main(apiurl, opts, argv):
         pac = pac + ":" + opts.multibuild_package
     if opts.wipe:
         buildargs.append("--wipe")
+    if (opts.stage_single and (opts.stage_pre or opts.stage_post)) \
+       or (opts.stage_pre and opts.stage_post):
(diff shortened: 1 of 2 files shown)


adrianschroeter:
this one needs to wait for https://github.com/openSUSE/obs-build/pull/647

adrianschroeter:
Alternative optoins got implemented in the build script. Added these now in https://github.com/openSUSE/osc/pull/899 for osc.

PR #891, opened 2021-01-18
support zstd compressed arch linux packages

adrianschroeter:
fixes current local build problems

### DIFF
--- a/osc/util/archquery.py
+++ b/osc/util/archquery.py
@@ -22,11 +22,15 @@ class ArchQuery(packagequery.PackageQuery, packagequery.PackageQueryResult):
     def read(self, all_tags=True, self_provides=True, *extra_tags):
         # all_tags and *extra_tags are currently ignored
         f = open(self.__path, 'rb')
-        #self.magic = f.read(5)
-        #if self.magic == '\375\067zXZ':
-        #    self.pkgsuffix = 'pkg.tar.xz'
+        magic = f.read(4)
+        f.close()
         fn = open('/dev/null', 'wb')
-        pipe = subprocess.Popen(['tar', '-O', '-xf', self.__path, '.PKGINFO'], stdout=subprocess.PIPE, stderr=fn).stdout
+        pipe = None
+        if magic == b'(\xb5/\xfd':
+            decompress = subprocess.Popen(['zstdcat', self.__path], stdout=subprocess.PIPE, stderr=fn)
+            pipe = subprocess.Popen(['tar', '-O', '-xf', '-', '.PKGINFO'], stdin=decompress.stdout, stdout=subprocess.PIPE, stderr=fn).stdout
+        else:
+            pipe = subprocess.Popen(['tar', '-O', '-xf', self.__path, '.PKGINFO'], stdout=subprocess.PIPE, stderr=fn).stdout
         for line in pipe.readlines():
             line = line.rstrip().split(b' = ', 2)
             if len(line) == 2:


marcus-h:
On 2021-01-18 09:19:30 +0000, Adrian Schröter wrote:
> Subject: [PATCH] support zstd compressed arch linux packages
>
> fixes current local build problems

Just to be sure I understand it correctly: this patch is only
needed if the local tar does not know about zstd, right?

> ---
>  osc/util/archquery.py | 12 ++++++++----
>  1 file changed, 8 insertions(+), 4 deletions(-)
>
> diff --git a/osc/util/archquery.py b/osc/util/archquery.py
> index e8debe8b..d05285a9 100644
> --- a/osc/util/archquery.py
> +++ b/osc/util/archquery.py
> @@ -22,11 +22,15 @@ def __init__(self, fh):
>      def read(self, all_tags=True, self_provides=True, *extra_tags):
>          # all_tags and *extra_tags are currently ignored
>          f = open(self.__path, 'rb')
> -        #self.magic = f.read(5)
> -        #if self.magic == '\375\067zXZ':
> -        #    self.pkgsuffix = 'pkg.tar.xz'
> +        magic = f.read(4)
> +        f.close()
>          fn = open('/dev/null', 'wb')
> -        pipe = subprocess.Popen(['tar', '-O', '-xf', self.__path, '.PKGINFO'], stdout=subprocess.PIPE, stderr=fn).stdout
> +        pipe = None
> +        if magic == b'(\xb5/\xfd':
> +            decompress = subprocess.Popen(['zstdcat', self.__path], stdout=subprocess.PIPE, stderr=fn)
> +            pipe = subprocess.Popen(['tar', '-O', '-xf', '-', '.PKGINFO'], stdin=decompress.stdout, stdout=subprocess.PIPE, stderr=fn).stdout

Hmm what about using tar's -I/--use-compress-program option?

> +        else:
> +            pipe = subprocess.Popen(['tar', '-O', '-xf', self.__path, '.PKGINFO'], stdout=subprocess.PIPE, stderr=fn).stdout

Maybe something like

with open(self.__path, 'rb') as f:
    magic = f.read(4)
cmd = ['tar', '-O', '-xf', self.__path, '.PKGINFO']
if magic == b'(\xb5/\xfd':
    cmd[2:2] = ['-I', 'zstd']
pipe = ...

(untested)

dmach:
@adrianschroeter how about to use @marcus-h 's solution? It seems to be a bit more elegant.

osc/util/archquery.py
```diff
@@ -22,11 +22,15 @@ def __init__(self, fh):
...
-        #if self.magic == '\375\067zXZ':
-        #    self.pkgsuffix = 'pkg.tar.xz'
+        magic = f.read(4)
+        f.close()
         fn = open('/dev/null', 'wb')
-        pipe = subprocess.Popen(['tar', '-O', '-xf', self.__path, '.PKGINFO'], stdout=subprocess.PIPE, stderr=fn).stdout
+        pipe = None
+        if magic == b'(\xb5/\xfd':
+            decompress = subprocess.Popen(['zstdcat', self.__path], stdout=subprocess.PIPE, stderr=fn)
+            pipe = subprocess.Popen(['tar', '-O', '-xf', '-', '.PKGINFO'], stdin=decompress.stdout, stdout=subprocess.PIPE, stderr=fn).stdout
```
dirkmueller, line 31: tar -I zstd

adrianschroeter:
it depends not really of tar on arch linux, but instead of the tar used by the osc user. If he does it on an older distro, he will still run into it.

But I do not feel strong in any direction here. We can also assume that most distros have a tar supporting zstd meanwhile.

dmach:
Ok, let's assume all modern distros have `zstd` support built in already.
If it turns to be a problem again, we'll reopen this or write a new patch.

PR #897, opened 2021-02-05
add showlinked command

adrianschroeter:
thought we have it since > 10 years ....

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,5 @@
 0.173.0
-  -
+  - add showlinked command to show all references of packages linking to a given one
 
 0.172.0
   - support --lastsucceeded/--last-succeeded in "osc buildlog", "osc
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -739,6 +739,33 @@ class Osc(cmdln.Cmdln):
         set_devel_project(apiurl, project, package, devprj, devpkg)
 
 
+    def do_showlinked(self, subcmd, opts, *args):
+        """${cmd_name}: Show all packages linking to a given one
+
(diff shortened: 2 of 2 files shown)


marcus-h:
On 2021-02-05 12:20:46 +0000, Adrian Schröter wrote:
> +    def do_showlinked(self, subcmd, opts, *args):
> +        """${cmd_name}: Show all packages linking to a given one
> +
> +        Examples:
> +            osc showlinked [PROJECT PACKAGE]

Hmm an alternative to a new command might be an additional option to
"osc ls". Maybe something like "osc ls prj pkg --linked". But to be
honest I have no clue what's better from a user's POV...

> +        ${cmd_option_list}
> +        """
> +
> +        args = slash_split(args)
> +        apiurl = self.get_api_url()
> +        localdir = os.getcwd()
> +        project = package = None
> +        if not args:
> +            if is_package_dir(localdir):
> +                project = store_read_project(localdir)
> +                package = store_read_package(localdir)
> +        elif len(args) == 2:
> +            project = args[0]
> +            package = args[1]
> +
> +        if project == None or package == None:
> +            raise oscerr.WrongArgs('Either specify project and package or call it from a package working copy')
> +

Minor: could be simplified to

if len(args) == 2:
    project, package = args
elsif is_package_dir(localdir):
    project = store_read_project(localdir)
    package = store_read_package(localdir)
else:
    raise oscerr.WrongArgs('Either specify project and package or call it from a package working copy')

(feel free to ignore.)

Otherwise, LGTM.

marcus-h:
Merged. Thanks!

Issue #903, opened 2021-03-21
AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'getchildren'


z3ntu:
I got this exception when running `osc add https://github.com/openrazer/openrazer/releases/download/v3.0.0/openrazer-3.0.0.tar.xz` on my project.

osc 0.172.0

```
Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.9/site-packages/osc/babysitter.py", line 69, in run
    return prg.main(argv)
  File "/usr/lib/python3.9/site-packages/osc/cmdln.py", line 347, in main
    return self.cmd(args)
  File "/usr/lib/python3.9/site-packages/osc/cmdln.py", line 370, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.9/site-packages/osc/cmdln.py", line 504, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.9/site-packages/osc/cmdln.py", line 1235, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.9/site-packages/osc/commandline.py", line 4756, in do_add
    addGitSource(arg)
  File "/usr/lib/python3.9/site-packages/osc/core.py", line 7153, in addGitSource
    stripETxml( services )
  File "/usr/lib/python3.9/site-packages/osc/core.py", line 7142, in stripETxml
    for child in node.getchildren():
AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'getchildren'
```

marcus-h:
On 2021-03-21 10:32:37 -0700, Luca Weiss wrote:
> I got this exception when running `osc add https://github.com/openrazer/openrazer/releases/download/v3.0.0/openrazer-3.0.0.tar.xz` on my project.
>

<SNIP>

>   File "/usr/lib/python3.9/site-packages/osc/core.py", line 7142, in stripETxml
>     for child in node.getchildren():
> AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'getchildren'

Should be fixed in PR#904 [1].

[1] https://github.com/openSUSE/osc/pull/904

PR #904, opened 2021-03-21
Do not use the deprecated Element.getchildren anymore

marcus-h:
Element.getchildren is deprecated and not available on python39
anymore. Instead, iterate over the element itself (which iterates
over the element's children).

Fixes: #903 ("AttributeError: 'xml.etree.ElementTree.Element' object
has no attribute 'getchildren'")


z3ntu:
Seems to work, I'm hitting a different problem now though: https://github.com/openSUSE/osc/issues/905

PR #907, opened 2021-03-25
remove "need root" autodetection

adrianschroeter:
This kind of guessing can not really work here and leads to failing
builds when using KVM. (eg. when using a preinstallimage)

Removing the code, since we have a now a way to allow the user to
specify building as user via su-wrapper config

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,7 @@
     running the install section to see the result of it (use --stage=i=).
     Check the help for more details.
   - allow to run build script as non-root, by setting su-wrapper empty
+    => osc is not guessing anymore if user builds are wanted
 
 0.172.0
   - support --lastsucceeded/--last-succeeded in "osc buildlog", "osc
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2021-03-25 01:05:44 -0700, Adrian Schröter wrote:
> This kind of guessing can not really work here and leads to failing
> builds when using KVM. (eg. when using a preinstallimage)
>
Hmm what's different from a non preinstallimage build?

Merged. Thanks!

adrianschroeter:
On Donnerstag, 25. März 2021, 11:59:39 CEST Marcus Hüwe wrote:
> On 2021-03-25 01:05:44 -0700, Adrian Schröter wrote:
> > This kind of guessing can not really work here and leads to failing
> > builds when using KVM. (eg. when using a preinstallimage)
> >
> Hmm what's different from a non preinstallimage build?

the rpm packages can just get extracted as user. The ownership and %post
script actions get fixed during the installation inside of the VM of
the same packages.

But preinstallimage content gets not installed again, so the system would be
broken when extracting as user.

-- 

Adrian Schroeter ***@***.***>

PR #908, opened 2021-03-25
do not use preinstallimages when building as non-root

adrianschroeter:
it can not work and build script will direct complain in future

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -547,10 +547,12 @@ def calculate_build_root(apihost, prj, pac, repo, arch):
                             % {'repo': repo, 'arch': arch, 'project': prj, 'package': pac, 'apihost': apihost}
     return buildroot
 
-def run_build(opts, *args):
-    cmd = [config['build-cmd']]
-    cmd += args
+def build_as_user():
+    if os.environ.get('OSC_SU_WRAPPER', config['su-wrapper']).split():
+        return False
+    return True
 
+def su_wrapper(cmd):
     sucmd = os.environ.get('OSC_SU_WRAPPER', config['su-wrapper']).split()
     if sucmd:
         if sucmd[0] == 'su':
@@ -559,6 +561,14 @@ def run_build(opts, *args):
             cmd = sucmd + ['-s', cmd[0], 'root', '--'] + cmd[1:]
         else:
             cmd = sucmd + cmd
+    return cmd
+
+def run_build(opts, *args):
+    cmd = [config['build-cmd']]
+    cmd += args
+
+    cmd = su_wrapper(cmd)
+
     if not opts.userootforbuild:
         cmd.append('--norootforbuild')
     return run_external(cmd[0], *cmd[1:])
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2021-03-25 06:50:21 -0700, Adrian Schröter wrote:
> index 55a69a4a..95166f58 100644
> --- a/osc/build.py
> +++ b/osc/build.py
> @@ -547,18 +547,28 @@ def calculate_build_root(apihost, prj, pac, repo, arch):
>                              % {'repo': repo, 'arch': arch, 'project': prj, 'package': pac, 'apihost': apihost}
>      return buildroot
>
> -def run_build(opts, *args):
> -    cmd = [config['build-cmd']]
> -    cmd += args
> +def build_as_user():
> +    if os.environ.get('OSC_SU_WRAPPER', config['su-wrapper']).split():
> +        return False
> +    return True
>
> +def su_wrapper(cmd):
>      sucmd = os.environ.get('OSC_SU_WRAPPER', config['su-wrapper']).split()

Now we always take OSC_SU_WRAPPER into account (but I don't think
it will harm - actually, it makes the behavior more consistent...)

(Maybe mark both functions as "private" (_build_as_user, _su_wrapper)?)

>      if sucmd:
>          if sucmd[0] == 'su':
>              if sucmd[-1] == '-c':
>                  sucmd.pop()
> -            cmd = sucmd + ['-s', cmd[0], 'root', '--'] + cmd[1:]
> +            cmd = sucmd + ['-s', cmd[0], 'root', '--' ] + cmd[1:]

Minor: remove space before ']' (feel free to ignore)

> +def run_build(opts, *args):
> +    cmd = [config['build-cmd']]
> +    cmd += args
> +
> +    cmd = su_wrapper(cmd)
> +
>      if not opts.userootforbuild:
>          cmd.append('--norootforbuild')
>      return run_external(cmd[0], *cmd[1:])
> @@ -1007,6 +1017,9 @@ def main(apiurl, opts, argv):
>      imagefile = ''
>      imagesource = ''
>      imagebins = []
> +    if build_as_user:

The function call is missing.

> +       # preinstallimage extraction will fail
> +       bi.preinstallimage = None

What's special in the preinstallimage codepath compared to the "normal"
preinstall pkg codepath? For instance, both use "bsdtar ... --chroot ...",
which only works if the current user has the appropriate capabilities.

mlschroe:
Normal preinstalls are not an issue because packages will be re-installed later on so we can get away with not setting the correct user and not creating device nodes and the like. That's not true for preinstall images, though.

marcus-h:
On 2021-03-25 14:59:51 +0000, Michael Schroeder wrote:
> Normal preinstalls are not an issue because packages will be re-installed later on so we can get away with not setting the correct user and not creating device nodes and the like.

Ah ok, but this is currently not yet implemented in init_buildsystem,
right?
(In theory an rpm could also ship device nodes.)

marcus-h:
On 2021-03-25 17:37:20 +0000, Marcus Hüwe wrote:
> (In theory an rpm could also ship device nodes.)
>
Scratch that... as you said missing device nodes are (usually) no
problem during preinstall.

Issue #417, opened 2018-05-14
unable to build using osc build Error: build ('/usr/bin/build') command not found 


hackertron:
Hello i am trying to build a package ( rclone ) for s390x. i have installed osc using zypper install osc . 

when i  execute osc build i get following error 
`Error: build ('/usr/bin/build') command not found
Install the build package from http://download.opensuse.org/repositories/openSUSE:/Tools/`

I did visisted the link and downloaded every RPM in that directory : http://download.opensuse.org/repositories/openSUSE:/Tools/SLE_12_SP3/s390x/

I am using s390x which is built on SLES12 SP3. I have also tried same thing on my OpenSUSE LEAP it works absolutely fine there .

adrianschroeter:
the build package is in noarch/ directory.

PR #909, opened 2021-04-08
Add support for sccache.

Firstyear:
Sccache is an alternate build caching system to ccache/icecream. It
supports C, C++ and Rust. It can optionally have distributed or remote
caches via redis, s3 object stores, memcached, azure storage or
google cloud storage.

This can help to significantly improve the performance of Rust rebuilds.

For example, Kanidm changes from 400s to 122s on a rebuild, and rust-lang
rebuilds improve from 7200s to 4770s. With some changes to the rust
packages especially this will be possible to speed up over version
changes as well.

See also: obs-build PR https://github.com/openSUSE/obs-build/pull/680

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -649,6 +649,16 @@ def main(apiurl, opts, argv):
     if opts.pkg_ccache:
         buildargs.append('--pkg-ccache=%s' % opts.pkg_ccache)
         xp.append('ccache')
+    if opts.sccache_uri or config['sccache_uri'] or opts.sccache or config['sccache']:
+        if opts.pkg_ccache or opts.ccache or config['ccache']:
+            raise oscerr.WrongArgs('Error: sccache and ccache can not be enabled at the same time')
+        if opts.sccache_uri:
+            buildargs.append('--sccache-uri=%s' % opts.sccache_uri)
+        elif config['sccache_uri']:
+            buildargs.append('--sccache-uri=%s' % config['sccache_uri'])
+        else:
+            buildargs.append('--sccache')
+        xp.append('sccache')
     if opts.linksources:
         buildargs.append('--linksources')
     if opts.baselibs:
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -6463,6 +6463,10 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                   help='use ccache to speed up rebuilds')
     @cmdln.option('--pkg-ccache', metavar='/path/to/_ccache.tar',
                   help='path to an existing uncompressed archive ccache. Using this option implies --ccache')
+    @cmdln.option('--sccache', action='store_true',
+                  help='use sccache to speed up rebuilds. Conflicts with --cache')
+    @cmdln.option('--sccache-uri', metavar='redis://127.0.0.1:6389',
+                  help='Optional remote URI for sccache storage. Implies --sccache.')
     @cmdln.option('--with', metavar='X', dest='_with', action='append',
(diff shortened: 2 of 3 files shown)


Firstyear:
@adrianschroeter Relates openSUSE/obs-build#680 #417

marcus-h:
Thanks a lot for the PR! Some minor comments below.

On 2021-04-08 03:11:05 -0700, Firstyear wrote:
> index 55a69a4a..95cfdd6c 100644
> --- a/osc/build.py
> +++ b/osc/build.py
> @@ -639,6 +639,16 @@ def main(apiurl, opts, argv):
>      if opts.pkg_ccache:
>          buildargs.append('--pkg-ccache=%s' % opts.pkg_ccache)
>          xp.append('ccache')
> +    if opts.sccache_uri or config['sccache_uri'] or opts.sccache or config['sccache']:
> +        if opts.ccache or config['ccache']:
> +            raise oscerr.WrongArgs('Error: sccache and ccache can not be enabled at the same time')

Shouldn't we also error out in case of the "--pkg-ccache" option?

> +        if opts.sccache_uri:
> +            buildargs.append('--sccache-uri=%s' % opts.sccache_uri)
> +        elif config['sccache_uri']:
> +            buildargs.append('--sccache-uri=%s' % config['sccache_uri'])
> +        else:
> +            buildargs.append('--sccache')
> +        xp.append('sccache')
>      if opts.linksources:
>          buildargs.append('--linksources')
>      if opts.baselibs:
> index e7903b6f..2ab9ddc6 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -6463,6 +6463,10 @@ def parse_repoarchdescr(self, args, noinit = False, alternative_project = None,
>                    help='use ccache to speed up rebuilds')
>      @cmdln.option('--pkg-ccache', metavar='/path/to/_ccache.tar',
>                    help='path to an existing uncompressed archive ccache. Using this option implies --ccache')
> +    @cmdln.option('--sccache', action='store_true',
> +                  help='use sccache to speed up rebuilds. Conflicts with --cache')
> +    @cmdln.option('--sccache-uri', metavar='redis://127.0.0.1:6389', default=None,
> +                  help='Optional remote URI for sccache storage. Implies --sccache.')

Minor: no need ot specify default=None.

>      @cmdln.option('--with', metavar='X', dest='_with', action='append',
>                    help='enable feature X for build')
>      @cmdln.option('--without', metavar='X', action='append',
> index 5b5506ab..a2def6ee 100644
> --- a/osc/conf.py
> +++ b/osc/conf.py
> @@ -121,6 +121,8 @@ def _get_processors():
>              'builtin_signature_check': '1',     # by default use builtin check for verify pkgs
>              'icecream': '0',
>              'ccache': '0',
> +            'sccache': '0',
> +            'sccache_uri': None,
>
Hmm in case of an unset option, we should use the empty string '' instead
of None.

Otherwise, LGTM!

Firstyear:
> Shouldn't we also error out in case of the "--pkg-ccache" option?

Yep! Good catch, I have updated this.

> Minor: no need ot specify default=None.

Updated

> Hmm in case of an unset option, we should use the empty string '' instead of None. Otherwise, LGTM!

And also done! Thank you for the review! I have squashed and rebased to master as well :)

Firstyear:
Odd, I'm not seeing those conflicts locally. Any ideas what I've done wrong here?

marcus-h:

Firstyear:
Thanks for merging it!

Issue #304, opened 2017-07-03
osc ci - cannot handle more than 2 GB file uploads


mmohring:
if you add a very big file downloaded_files.tar.gz, more than 2 GB to a package working copy with:

$ osc add downloaded_files.tar.gz
$ osc ci -m"added: very big file"

it results in an error trace:

sending    downloaded_files.tar.gz
Transmitting file data .Traceback (most recent call last):
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python2.7/site-packages/osc/babysitter.py", line 61, in run
    return prg.main(argv)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 501, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 1232, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python2.7/site-packages/osc/commandline.py", line 4568, in do_commit
    self._commit(subcmd, opts, args)
  File "/usr/lib/python2.7/site-packages/osc/commandline.py", line 4652, in _commit
    prj.commit(packages, msg=msg, files=files, skip_local_service_run=skip_local_service_run, verbose=opts.verbose, can_branch=can_branch, force=opts.force)
  File "/usr/lib/python2.7/site-packages/osc/core.py", line 973, in commit
    p.commit(msg, verbose=verbose, skip_local_service_run=skip_local_service_run, can_branch=can_branch, force=force)
  File "/usr/lib/python2.7/site-packages/osc/core.py", line 1493, in commit
    self.put_source_file(filename, tdir)
  File "/usr/lib/python2.7/site-packages/osc/core.py", line 1341, in put_source_file
    http_PUT(u, file = tfilename)
  File "/usr/lib/python2.7/site-packages/osc/core.py", line 3312, in http_PUT
    def http_PUT(*args, **kwargs):    return http_request('PUT', *args, **kwargs)
  File "/usr/lib/python2.7/site-packages/osc/core.py", line 3299, in http_request
    fd = urlopen(req, data=data)
[...]

Issue #202, opened 2016-05-28
osc user authentification seems to be broken with last commit


kkeil:
with current master for example  this does not longer work
`osc -c ~/testobsadmin.rc meta prjconf  Test10 -F newprjconf
Sending meta data...
Server returned an error: HTTP Error 500: Internal Server Error

Request: https://api.linux-pingi.de/source/Test10/_config
Headers:
content-language: en
accept-ranges: bytes
vary: accept-language,accept-charset
server: Apache
connection: close
cache-control: public
date: Sat, 28 May 2016 10:42:35 GMT
content-type: text/html; charset=utf-8
`
Same command with 0.154.0  works, a quick bisect found 
[c53a7681](https://github.com/openSUSE/osc/commit/c53a7681ef289181b03e1077efbfd246bf1247f6) as reason.

marcus-h:
Hmm I cannot reproduce this:

marcus@linux:~> osc meta prjconf home:Marcus_H > prjconf
marcus@linux:~> echo >> prjconf 
marcus@linux:~> osc meta prjconf home:Marcus_H -F prjconf
Sending meta data...
Done.
marcus@linux:~>

Can you post the output of
"osc -d -c ~/testobsadmin.rc meta prjconf  Test10 -F newprjconf"?

mmohring:
I can reproduce this:

`> osc -A https://api.myobsapi.com meta prj -F OBS:Tools:Unstable.xml OBS:Tools:Unstable`

returns:

`Sending meta data...
Server returned an error: HTTP Error 500: Internal Server Error

Request: https://api.openlifecyclemanager.com/source/OBS%3ATools%3AUnstable/_meta
Headers:
content-language: en
accept-ranges: bytes
vary: accept-language,accept-charset
server: Apache
connection: close
cache-control: public
date: Sun, 29 May 2016 18:04:51 GMT
content-type: text/html; charset=utf-8
`

while the git commit 8466e49d59a9b6375943807a55564d7071689b0c works.

mmohring:
I checked with git commit c53a7681ef289181b03e1077efbfd246bf1247f6 , which returns this error.

mmohring:
Still happens with git commit 502dbaa030a13d48184eafec5d1fe3c3378af1fb the current git master.

marcus-h:
On 2016-05-29 11:15:45 -0700, Martin Mohring wrote:
> I can reproduce this:
> 
[...]

PR #910, opened 2021-04-09
Support an arbitrary sized file in core.http_request

marcus-h:
```
The old code only supports a file whose size is less then or equal
to INT_MAX (due to a reasonable(!) limit in M2Crypto). The actual
issue is in core.http_request which mmap(...)s the file, wraps it
into a memoryview/buffer and then passes the memoryview/buffer to
urlopen. Eventually, the whole memoryview/buffer is read into memory
(see m2_PyObject_GetBufferInt). If the file is too large (> INT_MAX),
m2_PyObject_GetBufferInt raises a ValueError (which is perfectly
fine!).
Reading a whole file into memory is completely insane. In order to
avoid this, we now simply pass a file-like object to urlopen (more
precisely, the file-like object is associated with the Request
instance that is passed to urlopen). The advantange is that the
file-like object is processed in chunks of 8192 bytes (see
http.client.HTTPConnection) (that is, only 8192 bytes are read into
memory (instead of the whole file)).

There are two pitfalls when passing a file-like object to urlopen:
* By default, a chunked Transfer-Encoding is applied. It seems that
  some servers (like api.o.o) do like this (PUTing a file with
  a chunked Transfer-Encoding to api.o.o results in status 400). In
  order to avoid a chunked Transfer-Encoding, we explicitly set a
  Content-Length header (we also do this in the non-file case (just
  for the sake of completeness)).
* If the request fails with status 401, it is retried with an
  appropriate Authorization header. When retrying the request, the
  file's offset has to be repositioned to the beginning of the file
  (otherwise, a 0-length body is sent which most likely does not
  match the Content-Length header).

Note: core.http_request's "data" and "file" parameters are now mutually
exclusive because specifying both makes no sense (only one of them
is considered) and it simplifies the implementation a bit.
```

Fixes: #202 ("osc user authentification seems to be broken with last
commit")
Fixes: #304 ("osc ci - cannot handle more than 2 GB file uploads")

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -478,6 +478,25 @@ def get_apiurl_usr(apiurl):
 def _build_opener(apiurl):
     from osc.core import __version__
     global config
+
+    class OscHTTPBasicAuthHandler(HTTPBasicAuthHandler, object):
+        # python2: inherit from object in order to make it a new-style class
+        # (HTTPBasicAuthHandler is not a new-style class)
+        def _rewind_request(self, req):
+            if hasattr(req.data, 'seek'):
+                # if the request is issued again (this time with an
+                # Authorization header), the file's offset has to be
+                # repositioned to the beginning of the file (otherwise,
+                # a 0-length body is sent which most likely does not match
+                # the Content-Length header (if present))
+                req.data.seek(0)
+
+        def retry_http_basic_auth(self, host, req, realm):
+            self._rewind_request(req)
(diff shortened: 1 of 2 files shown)


asdil12 (approved):
I was able to upload a 2.1GB source file using this patch.

marcus-h:
On 2021-04-14 05:37:05 -0700, Dominik Heidler wrote:
> I was able to upload a 2.1GB source file using this patch.
>
Thanks for testing! I just merged the PR.

PR #913, opened 2021-04-28
Adjust the description size of 'osc service run'

pmgdeb:


marcus-h:
Merged. Thanks a lot!

PR #914, opened 2021-04-28
support _keyinfo route with signkey command by default

adrianschroeter:


marcus-h:
Merged. Thanks! (We should probably also use the _keyinfo route
in the fetch module.)

PR #918, opened 2021-05-14
support cross compile using a sysroot

adrianschroeter:
We need to hand over the sysroot rpms in rpmlist with a prefix. And we
need to take care of the "crossarch" definition in buildinfo

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,7 @@
     Check the help for more details.
   - allow to run build script as non-root, by setting su-wrapper empty
     => osc is not guessing anymore if user builds are wanted
+  - add support for cross arch local build using a sysroot
 
 0.172.0
   - support --lastsucceeded/--last-succeeded in "osc buildlog", "osc
--- a/osc/build.py
+++ b/osc/build.py
@@ -126,11 +126,20 @@ class Buildinfo:
             # snapcraft also supports rpm
             self.pacsuffix = 'deb'
 
+        # The architectures become a bit mad ...
+        # buildarch: The architecture of the build result      (host arch in GNU definition)
+        # hostarch:  The architecture of the build environment (build arch in GNU defintion)
+        # crossarch: Same as hostarch, but indicating that a sysroot with an incompatible architecture exists
         self.buildarch = root.find('arch').text
+        if root.find('crossarch') != None:
+            self.crossarch = root.find('crossarch').text
+        else:
+            self.crossarch = None
         if root.find('hostarch') != None:
             self.hostarch = root.find('hostarch').text
         else:
             self.hostarch = None
+
         if root.find('release') != None:
             self.release = root.find('release').text
         else:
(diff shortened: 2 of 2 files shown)


adrianschroeter:
in case you want to test:

osc co home:adrianSuSE:crosstest/kernel-default

marcus-h:
I just had a very quick glance at it:

On 2021-05-14 07:04:24 -0700, Adrian Schröter wrote:
> index 42e07ea0..f43c40c4 100644
> --- a/osc/build.py
> +++ b/osc/build.py
> @@ -79,6 +79,8 @@
>  if hostarch == 'parisc':
>      hostarch = 'hppa'
>
> +crossarch = None
> +

Unused.

>  class Buildinfo:
>      """represent the contents of a buildinfo file"""
>
> @@ -126,11 +128,20 @@ def __init__(self, filename, apiurl, buildtype = 'spec', localpkgs = [], binaryt
>              # snapcraft also supports rpm
>              self.pacsuffix = 'deb'
>
> +        # The architectures become a bit mad ...

Definitely:)

> +        # buildarch: The architecture of the build result      (host arch in GNU definition)
> +        # hostarch:  The architecture of the build environment (build arch in GNU defintion)

I haven't checked the GNU definitions but buildarch <-> hostarch is
intended, right?

> +        # crossarch: Same as hostarch, but indicating that a sysroot with an incompatible architecture exists
>          self.buildarch = root.find('arch').text
> +        if root.find('crossarch') != None:

Minor: "if root.find('crossarch') is not None:" (but feel free to
keep it as is because the surrounding code also has the "!= None"
check...).

> +            self.crossarch = root.find('crossarch').text
> +        else:
> +            self.crossarch = None

Otherwise, LGTM.

adrianschroeter:
On Montag, 17. Mai 2021, 23:12:27 CEST Marcus Hüwe wrote:
> I just had a very quick glance at it:
> 
> On 2021-05-14 07:04:24 -0700, Adrian Schröter wrote:
> > diff --git a/osc/build.py b/osc/build.py
> > index 42e07ea0..f43c40c4 100644
> > --- a/osc/build.py
> > +++ b/osc/build.py
> > @@ -79,6 +79,8 @@
> >  if hostarch == 'parisc':
> >      hostarch = 'hppa'
> >
> > +crossarch = None
> > +
> 
> Unused.

removed

> >  class Buildinfo:
> >      """represent the contents of a buildinfo file"""
> >
> > @@ -126,11 +128,20 @@ def __init__(self, filename, apiurl, buildtype = 'spec', localpkgs = [], binaryt
> >              # snapcraft also supports rpm
> >              self.pacsuffix = 'deb'
> >
> > +        # The architectures become a bit mad ...
> 
> Definitely:)
> 
> > +        # buildarch: The architecture of the build result      (host arch in GNU definition)
> > +        # hostarch:  The architecture of the build environment (build arch in GNU defintion)
> 
> I haven't checked the GNU definitions but buildarch <-> hostarch is
> intended, right?

yes... our definition, GNU autotools, cmake, kernel, rpm macros ... all differ from each other :/

I will create some docu for that ...

> > +        # crossarch: Same as hostarch, but indicating that a sysroot with an incompatible architecture exists
> >          self.buildarch = root.find('arch').text
> > +        if root.find('crossarch') != None:
> 
> Minor: "if root.find('crossarch') is not None:" (but feel free to
> keep it as is because the surrounding code also has the "!= None"
> check...).

hm yes, there would be plenty of places to change. We could do so independend of this.

-- 

Adrian Schroeter ***@***.***>

Issue #920, opened 2021-06-04
osc commit should fail gracefully in case of nonexistent  filename


deneb-alpha:
Hello folks, 
I was committing a patchinfo change I made locally and while doing it I passed the wrong file path.

the command I executed was:
`osc commit -m "merged 242209 and 242247" _patchinfo`
 but I was in the wrong local subfolder and in that location a file `_patchinfo` was nonexistent.

# current behaviour:
 
osc is tracing:
```
Traceback (most recent call last):
  File "/usr/bin/osc", line 45, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.6/site-packages/osc/babysitter.py", line 69, in run
    return prg.main(argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 347, in main
    return self.cmd(args)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 370, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 504, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.6/site-packages/osc/cmdln.py", line 1235, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 4901, in do_commit
    self._commit(subcmd, opts, args)
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 5009, in _commit
    prj.commit(packages, msg=msg, files=prj_files, skip_local_service_run=skip_local_service_run, verbose=opts.verbose, can_branch=can_branch, force=opts.force)
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 1038, in commit
    self.commitExtPackage(pac, msg, todo, verbose=verbose, skip_local_service_run=skip_local_service_run)
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 1111, in commitExtPackage
    if os_path_samefile(os.path.join(self.dir, pac), os.getcwd()):
  File "/usr/lib/python3.6/site-packages/osc/core.py", line 270, in os_path_samefile
    return os.path.samefile(path1, path2)
  File "/usr/lib64/python3.6/genericpath.py", line 96, in samefile
    s1 = os.stat(f1)
FileNotFoundError: [Errno 2] No such file or directory: './_patchinfo'
```
# expected behaviour:

osc should provide an error message without a trace.

PR #921, opened 2021-06-04
Handle "cd prj; osc ci non_existent_pkg" more gracefully

marcus-h:
When trying to commit a non-existent package via Project.commit it
is treated as an external package (because a non-existent package
has no "state" inside the project). That is, Project.commitExtPackage
is called, which fails with a FileNotFoundError in case of a
non-existent package (and the traceback is printed to the user).
In order to fix this, treat a non-existent package as broken package.
That is, simply print an info message and do not error out with a
traceback (note: the commit is _not_ aborted).

Fixes: #920 ("osc commit should fail gracefully in case of
nonexistent  filename")


lethliel:
LGTM

PR #923, opened 2021-06-28
fix local product builds using obsrepositories:/ directives

adrianschroeter:
We need an absolute path since osc calls build in user home directory.

Also extend the file name to hint to osc (and not any service
configuration)


lethliel:
LGTM

PR #928, opened 2021-07-10
osc co/up: highlight pending requests' header

jengelh:
Found myself to mentally miss the request list of 1 pending rq and so
I am pitching the idea of subtle boldening for experimentation.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -7441,7 +7441,9 @@ def print_request_list(apiurl, project, package = None, states = ('new', 'review
     if not conf.config['check_for_request_on_action'] and not force:
         return
     requests = get_request_list(apiurl, project, package, req_state=states)
-    msg = 'Pending requests for %s: %s (%s)'
+    msg = '\nPending requests for %s: %s (%s)'
+    if sys.stdout.isatty():
+        msg = '\033[1m{}\033[0m'.format(msg)
     if package is None and len(requests):
         print(msg % ('project', project, len(requests)))
     elif len(requests):


marcus-h:
Sorry for the delay...

On 2021-07-10 02:43:05 -0700, Jan Engelhardt wrote:
> Found myself to mentally miss the request list of 1 pending rq and so
> I am pitching the idea of subtle boldening for experimentation.

Hmm I don't have a strong opinion here... The question is if we should
introduce a config option for disabling/enabling (I don't care about
its default) this behavior.

Any opinions?

dmach:
I'm ok with such change if it's done differently.
I don't like 2 occurrences of  the same text and using escape codes directly.

How about to create output.py that would encapsulate all generic code working with the terminal
and move the stuff there in form of methods and constants.

output.py:
```
RESET = '\033[0m'
BOLD = '\033[1m'

def bold(text):
    if not sys.stdout.isatty():
        return text
    return BOLD + text + RESET
```

core.py:
```
msg = '\n' + output.bold('Pending requests for %s: %s (%s)')
```

osc/core.py
```diff
@@ -7383,7 +7383,10 @@ def print_request_list(apiurl, project, package = None, states = ('new', 'review
     if not conf.config['check_for_request_on_action'] and not force:
         return
     requests = get_request_list(apiurl, project, package, req_state=states)
-    msg = 'Pending requests for %s: %s (%s)'
+    if not sys.stdout.isatty():
+        msg = '\nPending requests for %s: %s (%s)'
+    else
```
dmach, line 7388: ... and there's missing colon after `else` -> syntax error.

dmach: @jengelh ping?

dmach:
Thank you. Merging.

Issue #936, opened 2021-08-03
'osc build --local-package ...' fails with 'not an osc package working copy'


e4t:
When running `osc build --local-package ..` on a package that does not exist on OBS, the error message 
`Error: "/<directory>" is not an osc package working copy.` 
is generated. This occurs when build.main() tries to run source services, which are part of the core.Package
infrastructure which requires much of the OBS package infrastructure to be in place.
Since the purpose of `--local-package` is that this doesn't have to be the case, it is probably best to not attempt this.

osc version used: 0.174.0-lp152.321

e4t:
Pull request #937 fixes this issue.

PR #937, opened 2021-08-03
[build] Do not try to run source services if --local-package is set

e4t:
When using the option --local-package to build a package that doesn't
exist on the on the server, yet, the error message:
 "Error: "<directory>" is not an osc package working copy."
is generated.
This occurs when build.main() attempts to run source services which
is probably not a good idea as these are part of the core.Package
infrastructure which cannot be initialized for such packages.

It is probably best to skip the source services. See Issue#936.

Signed-off-by: Egbert Eich <eich@suse.com>

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -790,6 +790,9 @@ def main(apiurl, opts, argv):
     if opts.noinit:
         buildargs.append('--noinit')
 
+    if not is_package_dir('.'):
+        opts.noservice = True
+
     # check for source services
     if not opts.offline and not opts.noservice:
         p = osc.core.Package(os.curdir)


e4t:
> we should not skip services in this case, but fix the issue that it can not get executed.

I've looked into this. Since running services is a functionality of the core.Package class it needs to initialize this. This attempts to test a version of the package directory meta-data, of course this meta data is not there. While in our case it may not be required, It will be in some other case where the Package class is used. 
But that's not all:  commenting out this check for fixes, a similar error message is generated later when the Project class is being initialized - despite passing the --alternative-project=<project> arg.
Fixing this would've required me to understand why runservice is part of the Package class in the first place and what it requires from it. This is far beyond the time I'm able to invest on this.

The --local-package option is clearly not doing what it is expected to. I believe the issue can be worked around by adding the --no-service option - something this patch does implicitly. Also, this used to work as I'm using it as part of a script which I used regularly years ago. So, that it doesn't is in fact a regression that should be fixed.
I agree, in the general case it would be good to be able to run service files even without requiring a project - my special use case doesn't require this. It seems that some services cannot live without a configured OBS project/package underneath. This is probably the biggest challenge.

marcus-h:
On 2021-08-03 10:18:25 +0000, Egbert Eich wrote:
> > we should not skip services in this case, but fix the issue that it can not get executed.
>
> I've looked into this. Since running services is a functionality of the core.Package class it needs to initialize this. This attempts to test a version of the package directory meta-data, of course this meta data is not there.

Hmm "of course" might be a "conceptual" issue here: originally, the
--local-package option was introduced for the case where a local
package (including the package "meta data") exists but the package
does not exist on the server.

Original use case:
$> cd prj
$> osc mkpac my_new_pkg
$> cd my_new_pkg
$> <create source files>
$> osc build --local-package

In this case, all services can be executed.

What you have in mind is probably the following use case:
$> cd "$(mktemp -d)"
$> <create source files>
$> osc build --local-package --alternative-project <prj>

(that is, performing a local build in a non-package working copy
(aka a directory without "meta data")). Right?

This used to work some years ago because service execution errors
were silently ignored, which is wrong (this was fixed in commit
c39c3b8cae826723c23e5115e5c051438d7aa8ba ("Cleanup the source
services execution code in do_build")).

Long story short: in order to get the "old" behavior, you have to
use the --noservice option, that is

$> osc build --local-package --alternative-project <prj> --noservice

Does this make sense?

> While in our case it may not be required, It will be in some other case where the Package class is used.
> But that's not all:  commenting out this check for fixes, a similar error message is generated later when the Project class is being initialized - despite passing the --alternative-project=<project> arg.
> Fixing this would've required me to understand why runservice is part of the Package class in the first place and what it requires from it. This is far beyond the time I'm able to invest on this.
>
In essence, Package.run_source_services just passes the project,
package, and apiurl to the ServiceInfo instance.

> The --local-package option is clearly not doing what it is expected to. I believe the issue can be worked around by adding the --no-service option - something this patch does implicitly. Also, this used to work as I'm using it as part of a script which I used regularly years ago. So, that it doesn't is in fact a regression that should be fixed.

See above. Btw., this PR is identical to PR #573 [1].

> I agree, in the general case it would be good to be able to run service files even without requiring a project - my special use case doesn't require this. It seems that some services cannot live without a configured OBS project/package underneath. This is probably the biggest challenge.
>
We just need a heuristic to derive the project, package, and apiurl.
- apiurl: use the default apiurl
- project: use the one specified by the --alternative-project option?
- package:
  * basename of cwd? (does this make sense if cwd is a tmpdir created
    via mktemp -d?)
  * build description name? (does this make sense if, for instance, a
    spec file is named like this: my_pkg-openSUSE_Factory.spec?)


[1] https://github.com/openSUSE/osc/pull/573

e4t:
@marcus-h 
'osc mkpac' is definitely not an option for me, moreover, the documentation for --local-package does not draw a connection to it either. 
From looking at #573 mine doesn't seem to be the only use case for using 'osc build' without having a proper package inside a proper project. 
Trying to make assumptions about the package will not work there,  while I do provide the build description (spec-file) on the command line.
I do like your suggestion in #573, however:
```
if not is_package_dir('.'):
    opts.noservice = True
```

e4t:
I've updated my patch including Marcus' suggestion, now.

marcus-h:
On 2021-08-07 13:02:47 -0700, Egbert Eich wrote:
> I've updated my patch including Marcus' suggestion, now.
>
Hmm hmm. I'm a bit hesitant...
- my naive POV: if a user runs "osc build" in a non package wc, they
  should be aware that services won't work. Hence, they should simply
  specify the --noservice option.
- from a user's POV: it is (probably?) not apparent that the services
  cannot be executed in a non package wc build. Hence, osc should
  automatically do the "right" thing and skip the services. Moreover,
  skipping the services in case of a non package wc build was the
  "old default" (so the current behavior can be considered as a
  regression).

Now, I'm somewhat convinced that merging this PR is the "right"
thing (at least from a user's POV:) ).

@***@***.***: any objections?


PS: could you slightly "tweak" the summary line of your patch?

> [build] Do run source services when building from an OSC
> package directory

That's what we already do:) Maybe use something like

"Do not run source services when building in a non package wc"

Maybe you could also add a

Fixes: #936 ("'osc build --local-package ...' fails with 'not an
osc package working copy'")

line?


PPS: thanks for your PR and thanks for bringing this up again!

marcus-h:
On 2021-08-07 16:47:22 -0700, Marcus Hüwe wrote:
> @***@***.***: any objections?
>
Oh my... let's try it this way
@adrianschroeter
@lethliel
:)

e4t:
Let me give my reasoning:

1. As you said already - the current behavior was not the behavior that we had for the longest time. Thus it can be considered a regression.
2.  I personally do not expect source services to run at all when I run osc build. If I need source services (and most of my packages need them) or want to test them, I will run them explicitly.

About the patch title: the new patch is unrelated to the  '--local-package' option, that you may need it in case you are running from a 'non-package wc' is now just a coincidence. This is why I changed to title.

marcus-h:
On 2021-08-07 23:14:55 -0700, Egbert Eich wrote:
> About the patch title: the new patch is unrelated to the  '--local-package' option, that you may need it in case you are running from a 'non-package wc' is now just a coincidence. This is why I changed to title.
>
Hmm your current summary line

> [build] Do run source services when building from an OSC
> package directory

does not reflect what the patch is doing. This statement is
also true without this patch:) That's why I suggested something
like

"Do not run source services when building in a non package wc"

(or "Do not try to run source ...").

e4t:
@marcus-h - indeed, you're right. I meant to say one thing and wrote the opposite :(

I hope I got it right now:
"Don't run source services when not building within an OSC package working copy"

marcus-h:
On 2021-08-08 09:20:10 -0700, Egbert Eich wrote:
> @marcus-h - indeed, you're right. I meant to say one thing and wrote the opposite :(
>
> I hope I got it right now:
> "Don't run source services when not building within an OSC package working copy"
>
Great! Thanks a lot!

If no one raises any objections within the next week/few days, I'll happily
merge it:)

e4t:
@marcus-h - Thanks! 
... And sorry for the confusion.

marcus-h:
So far there were no objections => merged.

Again, thanks a lot for the PR!

PR #940, opened 2021-08-12
Ensure creation of `~/.osc_cookiejar` adheres to XDG Base Directory Specification

hyperupcall:
In a similar theme to #349, this ensures the file is created according to the XDG Base Directory specification. Just like that change, it uses the XDG Base Directly specification on windows (usually I prefer to use the spec only on Unix-like platlforms). In this case, `XDG_CACHE_HOME` seemed to be a better fit compared to `XDG_STATE_HOME`

A subtle difference compared to #349 is that `XDG_STATE_HOME` is not used if it is defined, but empty (in the other change, the variable is used if it is empty). This is more correct, according to the spec. 

<details>
<summary>Test suite results (passes)</summary>

```txt
➤ python suite.py
..................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 194 tests in 1.253s

OK
```

</details>

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -94,6 +94,19 @@ def _get_processors():
     except ValueError as e:
         return 1
 
+
+def _identify_osccookiejar():
+    if os.path.isfile(os.path.join(os.path.expanduser("~"), '.osc_cookiejar')):
+        # For backwards compatibility, use the old location if it exists
+        return '~/.osc_cookiejar'
+
+    if os.getenv('XDG_STATE_HOME', '') != '':
+        osc_state_dir = os.path.join(os.getenv('XDG_STATE_HOME'), 'osc')
+    else:
+        osc_state_dir = os.path.join(os.path.expanduser("~"), '.local', 'state', 'osc')
+
+    return os.path.join(osc_state_dir, 'cookiejar')
+
 DEFAULTS = {'apiurl': 'https://api.opensuse.org',
             'user': None,
             'pass': None,
@@ -136,7 +149,7 @@ DEFAULTS = {'apiurl': 'https://api.opensuse.org',
             'post_mortem': '0',
             'use_keyring': '0',
             'gnome_keyring': '0',
-            'cookiejar': '~/.osc_cookiejar',
+            'cookiejar': _identify_osccookiejar(),
             # fallback for osc build option --no-verify
             'no_verify': '0',
             # enable project tracking by default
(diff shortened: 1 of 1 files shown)


hyperupcall:
> Why?:)
Naively, I would say that we should store it in $XDG_STATE_HOME
because:
> - it should persist between (application) restarts
> - it is not important enough to store it in $XDG_DATA_HOME

Ahhh yeah, let's do `$XDG_STATE_HOME`

> What are your arguments for $XDG_CACHE_HOME?

I originally went with it because I second guessed myself with `$XDG_STATE_HOME` (which is why that typo exists in the second paragraph :P)

> Good catch! (You are talking about $XDG_CACHE_HOME, right?)

Yeah

> Are you in the mood to create a new PR that fixes this for
$XDG_CONFIG_HOME?:)

Yup!

I'll soon update my branch with the code improvements you mentioned :)

marcus-h:
On 2021-08-13 14:10:49 -0700, Edwin Kofler wrote:
> I'll soon update my branch with the code improvements you mentioned :)
>
Thanks a bunch! The patches look good to me!

Can you squash both commits into one?

hyperupcall:
Awesome! - now squashed into a single commit

marcus-h:
On 2021-08-14 17:26:35 -0700, Edwin Kofler wrote:
> index 245dce76..33522dad 100644
> --- a/osc/conf.py
> +++ b/osc/conf.py
> @@ -94,6 +94,19 @@ def _get_processors():
>      except ValueError as e:
>          return 1
>
> +
> +def _identify_osccookiejar():
> +    if os.path.isfile(os.path.join(os.path.expanduser("~"), '.osc_cookiejar')):
> +        # For backwards compatibility, use the old location if it exists
> +        return '~/.osc_cookiejar'
> +
> +    if os.getenv('XDG_STATE_HOME', '') != '':
> +        osc_state_dir = os.path.join(os.getenv('XDG_STATE_HOME'), 'osc')
> +    else:
> +        osc_state_dir = os.path.join(os.path.expanduser("~"), '.cache', 'osc')
> +
This should be

osc_state_dir = os.path.join(os.path.expanduser("~"), '.local', 'state', 'osc')

> +    return os.path.join(osc_state_dir, 'cookiejar')
> +

hyperupcall:
Oops! Sorry about that

marcus-h:
Merged. Thanks a lot!

PR #950, opened 2021-09-10
Add support to manually override download url locations.

Firstyear:
This adds support base on each apiurl to customise the download location. This can be combined with a local proxy cache or other to allow caching of packages outside of osc's environment.

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -221,7 +221,8 @@ boolean_opts = ['debug', 'do_package_tracking', 'http_debug', 'post_mortem', 'tr
     'status_mtime_heuristic', 'print_web_links', 'ccache', 'sccache', 'build-shell-after-fail']
 integer_opts = ['build-jobs']
 
-api_host_options = ['user', 'pass', 'passx', 'aliases', 'http_headers', 'realname', 'email', 'sslcertck', 'cafile', 'capath', 'trusted_prj']
+api_host_options = ['user', 'pass', 'passx', 'aliases', 'http_headers', 'realname', 'email', 'sslcertck', 'cafile', 'capath', 'trusted_prj',
+    'downloadurl']
 
 new_conf_template = """
 [general]
@@ -1023,6 +1024,16 @@ def get_config(override_conffile=None,
             api_host_options[apiurl]['trusted_prj'] = cp.get(url, 'trusted_prj').split(' ')
         else:
             api_host_options[apiurl]['trusted_prj'] = []
+
+        # ⚠️  This option is experimental and may be removed at any time in the future!
+        # This allows overriding the download url for an OBS instance to specify a closer mirror
+        # or proxy system, which can greatly improve download performance, latency and more.
+        # For example, this can use https://github.com/Firstyear/opensuse-proxy-cache in a local
+        # geo to improve performance.
+        if cp.has_option(url, 'downloadurl'):
+            api_host_options[apiurl]['downloadurl'] = cp.get(url, 'downloadurl')
+        else:
+            api_host_options[apiurl]['downloadurl'] = None
 
     # add the auth data we collected to the config dict
     config['api_host_options'] = api_host_options
--- a/osc/build.py
+++ b/osc/build.py
@@ -144,7 +144,15 @@ class Buildinfo:
             self.release = root.find('release').text
         else:
             self.release = None
-        self.downloadurl = root.get('downloadurl')
+        if config['api_host_options'][apiurl]['downloadurl']:
+            self.enable_cpio = False
+            self.downloadurl = config['api_host_options'][apiurl]['downloadurl'] + "/repositories"
+            if config['http_debug']:
+                print("⚠️   setting dl_url to %s" % config['api_host_options'][apiurl]['downloadurl'])
+        else:
+            self.enable_cpio = True
+            self.downloadurl = root.get('downloadurl')
+
         self.debuginfo = 0
         if root.find('debuginfo') != None:
             try:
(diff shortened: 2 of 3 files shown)


Firstyear:
@adrianschroeter Do you mind taking a look at this?

marcus-h:
On 2021-09-09 22:52:26 -0700, Firstyear wrote:
> This adds support base on each apiurl to customise the download location. This can be combined with a local proxy cache or other to allow caching of packages outside of osc&#39;s environment.

Hmm what is your concrete use case for this?

Firstyear:
@marcus-h Downloads from dl.o.o are extremely slow in Australia. As a result, I have a local mirror cache setup that I want to be able to use. We also have a mirrorcache instance in the region that we want to use directly. Being able to configure this would be useful for myself and other SUSE developers in the region, or in other regions with poor internet performance.

marcus-h:
On 2021-09-27 16:28:50 -0700, Firstyear wrote:
> @marcus-h Downloads from dl.o.o are extremely slow in Australia. As a result, I have a local mirror cache setup that I want to be able to use. We also have a mirrorcache instance in the region that we want to use directly. Being able to configure this would be useful for myself and other SUSE developers in the region, or in other regions with poor internet performance.
>
Hmm can't you just setup a proxy that does the forwarding or simply
redirect d.o.o to your own instance via /etc/hosts?

Firstyear:
> Hmm can't you just setup a proxy that does the forwarding 

No, zypper has behaviours that pathologically break any attempt to use a generic http cache. Also TLS prevents caching.

> or simply redirect d.o.o to your own instance via /etc/hosts?

TLS prevents this also.

This is why I wrote https://github.com/Firstyear/opensuse-proxy-cache which is exactly what I'm trying to use here to resolve my issue, but I need osc to be able to be directed to it.

Firstyear:
Review reminder, this would help us in Australia a lot.

Firstyear:
@marcus-h @adrianschroeter Can you please review this? This is an important change that will affect all of us in Australia and remote locations, as we can then prefer faster local mirrors. The difference is package download rate is 300kb/s as currently is, vs 12MB/s from local mirrors. which is a different of about 40x.

simotek:
I'd also find this really useful.

marcus-h:
On 2021-11-21 22:13:36 -0800, Firstyear wrote:
> @marcus-h @adrianschroeter Can you please review this? This is an important change that will affect all of us in Australia and remote locations, as we can then prefer faster local mirrors. The difference is package download rate is 300kb/s as currently is, vs 12MB/s from local mirrors. which is a different of about 40x.
>
Hmm to be honest, I do not want to introduce a new config option for
such a specific problem/hack because we would have to maintain it
"forever"...
Instead, I would prefer to solve it by introducing the notion of a
cache manager: in essence, a cache manager takes care of fetching
the binaries and storing them locally (if desired). Having this,
we could implement/provide several cache managers, for instance,
- a cache manager could take care of purging "old" binaries from
  disk
- another cache manager could behave like the current osc cache+fetch
  code
- yet another cache manager could utilize a different URL for fetching
  the binaries (this would be your case)
- a chained cache manager (composition of other cache managers)
- etc.

For configuring a cache manager, a user just specifies the concrete
python class of a cache manager in their oscrc (basically, in the
same vein how we specify the "credentials_mgr_class" (per apiurl)
config option).

Would something like this be acceptable for you?

Firstyear:
Sure, that would be completely reasonable to me if you wanted to architect it like that. That's a lot more work and more invasive though, and I'm not sure I'm familiar enough and have the time to submit that. Is this something you are already planning to work on?

tserong:
> I'd also find this really useful.

Likewise

marcus-h:
On 2021-11-22 13:29:22 -0800, Firstyear wrote:
> Sure, that would be completely reasonable to me if you wanted to architect it like that. That's a lot more work and more invasive though, and I'm not sure I'm familiar enough and have the time to submit that. Is this something you are already planning to work on?
>
Yep, it is on my TODO (I just need to find time to do it...).

Firstyear:
> Yep, it is on my TODO (I just need to find time to do it...).

Okay, so here's the issue. We have a PR here that works, and does what multiple people need for quality of life. It's a feature you are open to, but in a different model. But you say also you don't have time. So either we wait for you to do it, which could take a long time, or we can merge this. So I'd like to ask, "when" would you have this done by?

marcus-h:
On 2021-11-24 19:32:29 -0800, Firstyear wrote:
> > Yep, it is on my TODO (I just need to find time to do it...).
>
> Okay, so here's the issue. We have a PR here that works, and does what multiple people need for quality of life. It's a feature you are open to, but in a different model. But you say also you don't have time. So either we wait for you to do it, which could take a long time, or we can merge this. So I'd like to ask, "when" would you have this done by?
>
Unfortunately, I cannot give you a serious ETA... My plan is to have a
look at it at the end of this year. Since this is urgent for you (which
I can fully understand!), I could probably live with merging this IF we
do NOT announce it as an official feature and treat it more like a hack
which is likely to be removed in a subsequent release (without adding
backward compatible code etc.).

Would this be OK for you?
@adrianschroeter, @lethliel: what do you think?

Firstyear:
Sounds fine to me, we could even rename the config parameter to call it unstable or something like that so people know it may be removed.

osc/conf.py
```diff
@@ -1007,6 +1007,11 @@ def get_config(override_conffile=None,
         else:
             api_host_options[apiurl]['trusted_prj'] = []
 
+        if cp.has_option(url, 'downloadurl'):
+            api_host_options[apiurl]['downloadurl'] = cp.get(url, 'downloadurl')
+        else:
+            api_host_options[apiurl]['downloadurl'] = None
+
```
dmach, line 1014: Could you add a comment here saying that the option is unstable/experimental and may change or disappear without prior notice? It might be also good to describe here what the option does.

dmach (changes requested):

osc/conf.py
Firstyear: Changed :)

Firstyear:
Is there anything else blocking this from being merged?

lethliel:
Since this is fine for everyone I will merge this. As @marcus-h wrote this can be removed in any future PR.

Firstyear:
Thank you! Many of the Australians are already excited to use this :)

PR #951, opened 2021-09-14
detachbranch: remove _link when link target got removed

adrianschroeter:
not 100% sure if the self.baserev check may lead to problems elsewhere ...

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -2855,7 +2855,14 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             li = Linkinfo()
             li.read(root.find('linkinfo'))
             if li.islink() and li.haserror():
-                raise oscerr.LinkExpandError(project, package, li.error)
+                try:
+                    show_package_meta(apiurl, li.project, li.package)
+                except HTTPError as e:
+                    if e.code == 404:
+                        print("Link target got removed, dropping link. WARNING: latest submissions in link target might be lost!")
+                        delete_files(apiurl, project, package, ['_link'])
+                    else:
+                        raise oscerr.LinkExpandError(project, package, li.error)
             elif not li.islink():
                 print('package \'%s/%s\' is no link' % (project, package), file=sys.stderr)
             else:
--- a/osc/core.py
+++ b/osc/core.py
@@ -540,7 +540,7 @@ class Linkinfo:
 
     def islink(self):
         """returns True if the linkinfo is not empty, otherwise False"""
-        if self.xsrcmd5 or self.lsrcmd5:
+        if self.xsrcmd5 or self.lsrcmd5 or self.error is not None:
             return True
         return False
 


marcus-h:
On 2021-09-14 12:12:38 +0000, Adrian Schröter wrote:
> index b49feda8..35ab2d75 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -2855,7 +2855,14 @@ def do_detachbranch(self, subcmd, opts, *args):
>              li = Linkinfo()
>              li.read(root.find('linkinfo'))
>              if li.islink() and li.haserror():
> -                raise oscerr.LinkExpandError(project, package, li.error)
> +                try:
> +                  linked_meta = show_package_meta(apiurl, li.project, li.package)

Minor: the assignment can be dropped.

> +                except HTTPError as e:
> +                  if e.code == 404:
> +                     print("Link target got removed, dropping link. WARNING: latest submissions in link target might be lost!")

Hmm at first I thought the warning part does not make sense but
I think I got it now:)

An alternative would be to expand to the last working rev (if
existing).

> +                     delete_files(apiurl, project, package, ['_link'])
> +                  else:
> +                     raise oscerr.LinkExpandError(project, package, li.error)

"Strange" indentation in the try-except (3 instead of 4).

>              elif not li.islink():
>                  print('package \'%s/%s\' is no link' % (project, package), file=sys.stderr)
>              else:
> index 4a34099d..4e384102 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -540,7 +540,7 @@ def read(self, linkinfo_node):
>
>      def islink(self):
>          """returns True if the linkinfo is not empty, otherwise False"""
> -        if self.xsrcmd5 or self.lsrcmd5:
> +        if self.xsrcmd5 or self.lsrcmd5 or self.baserev:

Hmm what about using "or self.error" (or "or self.error is not None")?
This attribute should always be present in case of a link error (the
baserev is not guaranteed to exist).

>              return True
>          return False

adrianschroeter:
good proposal with the self.error ....

marcus-h:
Can you also fix the indentation?

adrianschroeter:
On Dienstag, 14. September 2021, 18:54:21 CEST Marcus Hüwe wrote:
> Can you also fix the indentation?

I only see 2 and 4 chars indention..

-- 

Adrian Schroeter ***@***.***>

marcus-h:
On 2021-09-15 06:54:30 +0000, Adrian Schröter wrote:
> On Dienstag, 14. September 2021, 18:54:21 CEST Marcus Hüwe wrote:
> > Can you also fix the indentation?
>
> I only see 2 and 4 chars indention..
>
Oh indeed, no idea why I counted 3:)
Can you change the 2 to 4? (indention one of the few things that is
(more or less) consistent in the whole osc codebase;) )

marcus-h:
Merged. Thanks a lot!

Issue #960, opened 2021-10-17
git checkout fails on Windows


jayvdb:
The following occurs when doing a clone of this repo on Windows
```
Cloning into 'osc'...
remote: Enumerating objects: 17478, done.
remote: Counting objects: 100% (275/275), done.
remote: Compressing objects: 100% (166/166), done.
Receiving objects: 100% (17478/17478), 11.49 MiB | 1.14 MiB/s, done. 17203

Resolving deltas: 100% (11875/11875), done.
error: invalid path 'tests/prdiff_fixtures/home:user:branches:some:project/.osc/_apiurl'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```

https://stackoverflow.com/questions/5991805/how-do-i-clone-files-with-colons-in-the-filename gives a good overview of this problem, which doesnt appear to be solvable with any git config.

The checkout can be made almost complete by using `git restore --staged --source=HEAD .`
Then `git status` shows

```
On branch master
Your branch is up to date with 'origin/master'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        deleted:    tests/prdiff_fixtures/home:user:branches:some:project/.osc/_apiurl
        deleted:    tests/prdiff_fixtures/home:user:branches:some:project/.osc/_packages
        deleted:    tests/prdiff_fixtures/home:user:branches:some:project/.osc/_project
        deleted:    tests/prdiff_fixtures/home:user:branches:some:project/common-two
        deleted:    tests/prdiff_fixtures/home:user:branches:some:project/directory
        deleted:    tests/prdiff_fixtures/new:prj/common-two
        deleted:    tests/prdiff_fixtures/new:prj/directory
        deleted:    tests/prdiff_fixtures/old:prj/common-two
        deleted:    tests/prdiff_fixtures/old:prj/directory
        deleted:    tests/prdiff_fixtures/some:project/.osc/_apiurl
        deleted:    tests/prdiff_fixtures/some:project/.osc/_packages
        deleted:    tests/prdiff_fixtures/some:project/.osc/_project
        deleted:    tests/prdiff_fixtures/some:project/common-two
        deleted:    tests/prdiff_fixtures/some:project/directory
        deleted:    tests/update_fixtures/osctest/services/_service:exists
        deleted:    tests/update_fixtures/testUpdateServiceFilesAddDelete__service:bar
[...]

PR #962, opened 2021-10-18
store_write_string: Delete preexisting file on Win

jayvdb:
os.rename on Windows always raises an exception
if the target already exists.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -6600,6 +6600,8 @@ def store_write_string(dir, file, string, subdir=''):
             string = decode_it(string)
         f.write(string)
         f.close()
+        if sys.platform[:3] == 'win' and os.path.exists(fname):
+            os.remove(fname)
         os.rename(fname + '.new', fname)
     except:
         if os.path.exists(fname + '.new'):


marcus-h:
On 2021-10-17 17:31:11 -0700, John Vandenberg wrote:
> index 8732ed11..dbb03158 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -6600,6 +6600,8 @@ def store_write_string(dir, file, string, subdir=''):
>              string = decode_it(string)
>          f.write(string)
>          f.close()
> +        if sys.platform[:3] == 'win' and os.path.exists(fname):
> +            os.remove(fname)

Hmm we could do that but then the operation is not "atomic" anymore,
which could result in a "broken" project/package working copy (which
could be repaired, though).

Is there no atomic alternative on windows?

>          os.rename(fname + '.new', fname)
>      except:
>          if os.path.exists(fname + '.new'):

jayvdb:
Not with [os.rename](https://docs.python.org/3/library/os.html#os.rename).

Reading https://stackoverflow.com/questions/167414/is-an-atomic-file-rename-with-overwrite-possible-on-windows , it looks like there are some atomic alternatives on recent Windows.

https://docs.python.org/3/library/os.html#os.replace might be the alternative.

dmach:
Could you try to remove the platform specific condition and use `os.replace()` as you suggested?

From Python docs:
> If you want cross-platform overwriting of the destination, use replace().

jayvdb:
I dont have a Windows computer to test any longer.
And until https://github.com/openSUSE/osc/issues/960 is solved, using CI isnt an option either.

dmach:
Closing as we dropped Windows from supported operating systems (https://github.com/openSUSE/osc/pull/995)

PR #966, opened 2021-10-21
Improve logic for conffile mode handling

Firstyear:
It was noted during the use of osc inside of cargo_audit scans with the security team, that osc when run in nsjail on a read-only bind mount always attempts to chmod the osc config file. This improves the logic so that the chmod is only attempted if required, and if it fails we issue a warning instead of outright failing the execution. 

Fixes: https://github.com/openSUSE/obs-service-cargo_audit/issues/4

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -38,6 +38,7 @@ The configuration dictionary could look like this:
 
 import bz2
 import base64
+import errno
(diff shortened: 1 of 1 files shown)


Firstyear:
Added check for errno EROFS.

PR #975, opened 2021-11-30
extract config section from _buildenv if -r is set

lethliel:
use buildconfig stored in _buildenv to be sure the same config is used to build the package locally as it was used to build on the server when the original package was created.

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -20,7 +20,7 @@ except ImportError:
 
 from tempfile import NamedTemporaryFile, mkdtemp
 from osc.fetch import *
-from osc.core import get_buildinfo, store_read_apiurl, store_read_project, store_read_package, meta_exists, quote_plus, get_buildconfig, is_package_dir, dgst
+from osc.core import get_buildinfo, store_read_apiurl, store_read_project, store_read_package, meta_exists, quote_plus, get_buildconfig, is_package_dir, dgst, get_repr_buildconfig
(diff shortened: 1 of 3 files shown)


marcus-h:
On 2021-11-30 02:47:20 -0800, Marco Strigl wrote:
> use buildconfig stored in _buildenv to be sure the same config is used to build the package locally as it was used to build on the server when the original package was created.

Hmm currently the "reproducible" code path is triggered if a
_buildenv.<repo>.<arch> or _buildenv is present in the CWD. That is,
we probably do not need the "--reproduce" option.

As you point out, the current "reproducible" code path is flawed
because the "wrong" buildconfig is used. However, retrieving the
buildconfig from the latest _buildenv file, which is present on
the server, is also "wrong" (IMHO) because the local _buildenv and
the server-side _buildenv could mismatch. Hence, we should use the
buildconfig that is stored in the local _buildenv, IMHO.

It seems that the backend is also using the "wrong" buildconfig
when generating the buildinfo from a POSTed buildenv file. Instead
of using the one that is provided in the buildenv file, the latest
one is used. This could result, for instance, in a different set
of preinstall packages, which could result in a different build
result (compared to the "original" build result)...
(Note: I just had a very quick glance at the backend code, so I
might be wrong here.)

What do you think?

lethliel:
This has already been solved in the buildservice and can be closed.

PR #978, opened 2021-12-06
flake8: Initial configuration

dmach:
Set max line length to 120.
Keep the all the other settings on their default values.

If `max-line-length = 120` is too much, let me know which value you prefer.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,3 @@
+[flake8]
+exclude = .git,__pycache__
+max-line-length = 120


marcus-h:
On 2021-12-06 12:00:41 +0000, Daniel Mach wrote:
> Set max line length to 120.
> Keep the all the other settings on their default values.
>
> If `max-line-length = 120` is too much, let me know which value you prefer.
>
I'm also fine with 79 chars but given the existing osc codebase
120 chars might be a good start:)

Does this change have any impact on the existing codebase?

dmach:
> Does this change have any impact on the existing codebase?

It affects flake8 results. I hope it's going to mute some PEP 8 Speaks errors.
The checkers we've enabled show errors/warnings only for the code that has changed in a pull-request.

marcus-h:
On 2021-12-07 23:53:51 -0800, Daniel Mach wrote:
> > Does this change have any impact on the existing codebase?
>
> It affects flake8 results. I hope it's going to mute some PEP 8 Speaks errors.
> The checkers we've enabled show errors/warnings only for the code that has changed in a pull-request.
>
Ah cool! Merged. Thanks! Let's see how this works in practice:)

Issue #390, opened 2018-03-07
symlinked $HOME/.oscrc gets replaced with a ordinary file


HeikoSchlittermann:
For privacy reasons I had `~/.oscrc` linked from NFS mounted home directory to a directory on local storage. When _osc_ added something to my `~/.oscrc`, it replaced the symlink with an ordinary file, thus exposing my obs login credentials to the machines mounting the NFS share.

logological:
It should follow the symlink and write to the real file.  A lot of people (myself included) manage their dotfiles across machines using a Git repository and GNU Stow to create a symlink farm.  Having `osc` remove the symlink breaks this setup.

HeikoSchlittermann:
Thans for fixing this!

marcus-h:
On 2021-12-08 09:36:33 +0000, Heiko Schlittermann wrote:
> Thanks for fixing this!
>
Kudos go to @dmach:)

PR #979, opened 2021-12-06
conf: Preserve oscrc symlink

dmach:
When a user creates a symlink pointing from ~/.config/osc/oscrc to a different location,
don't overwrite the symlink but follow it when writing configuration on disk.

Fixes https://github.com/openSUSE/osc/issues/390

Could someone review this from security perspective?
I don't think that this could cause any harm (e.g. a user setting the path so root overwrites system files etc.), but I may have missed something.

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -695,8 +695,20 @@ def write_config(fname, cp):
     if os.path.exists(fname) and not os.path.isfile(fname):
         # only write to a regular file
         return
+
+    # config file is behind a symlink
+    # resolve the symlink and continue writing the config as usual
+    if os.path.islink(fname):
+        fname = os.readlink(fname)
+
+    # create directories to the config file (if they don't exist already)
     if not os.path.exists(os.path.dirname(fname)):
-        os.makedirs(os.path.dirname(fname), mode=0o700)
+        try:
+            os.makedirs(os.path.dirname(fname), mode=0o700)
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                raise
+
     with open(fname + '.new', 'w') as f:
         cp.write(f, comments=True)
     try:


marcus-h:
On 2021-12-06 05:30:05 -0800, Daniel Mach wrote:
> When a user creates a symlink pointing from ~/.config/osc/oscrc to a different location,
> don&#39;t overwrite the symlink but follow it when writing configuration on disk.
>
> Fixes https://github.com/openSUSE/osc/issues/390
>
> Could someone review this from security perspective?
> I don&#39;t think that this could cause any harm (e.g. a user setting the path so root overwrites system files etc.), but I may have missed something.
>
Since the symlink is configured by the user, it should be no problem,
IMHO.

> index fa12ef60..8c06bc46 100644
> --- a/osc/conf.py
> +++ b/osc/conf.py
> @@ -695,6 +695,26 @@ def write_config(fname, cp):
>      if os.path.exists(fname) and not os.path.isfile(fname):
>          # only write to a regular file
>          return
> +
> +    # config file is behind a symlink
> +    if os.path.islink(fname):
> +        # resolve the symlink
> +        fname = os.readlink(fname)
> +
> +        # create directories to the config file (if they don't exist already)
> +        try:
> +            os.makedirs(os.path.dirname(fname), mode=0o700)
> +        except OSError as e:
> +            if e.errno != errno.EEXIST:
> +                raise
> +
> +        # write the config directly (without an intermediate file)
> +        # because we don't want to make any assumptions about the directory the symlink points to
> +        with open(fname, 'w') as f:
> +            cp.write(f, comments=True)
> +
Hmm writing the config file should be "atomic", IMHO. Otherwise,
we could end up with a broken config file.

What about

if os.path.islink(fname):
    fname = os.readlink(fname)
# continue with the old code
...

If you are worried about the *.new file, we could use a temporary file
(or just don't care (for instance, "osc -c /path/to/config ..." could
also write a *.new file in the /path/to dir)).

> +        return
> +
>      if not os.path.exists(os.path.dirname(fname)):
>          os.makedirs(os.path.dirname(fname), mode=0o700)
>      with open(fname + '.new', 'w') as f:
>

dmach:
I simplified the code as you suggested.
I also kept my version of calling makedirs(), because I consider it better (thread-safe, never throws an exception in this place when 2 programs run in parallel).

marcus-h:
On 2021-12-07 08:35:17 +0000, Daniel Mach wrote:
> I simplified the code as you suggested.
> I also kept my version of calling makedirs(), because I consider it better (thread-safe, never throws an exception in this place when 2 programs run in parallel).
>
Ok, I'm fine with that. Maybe we should only call os.makedirs(...) if the
directory does not exist?

if not os.path.exists(os.path.dirname(fname)):
    try:
        os.makedirs(...)
    ...


At least I don't see a reason why we should call it unconditionally:)

dmach:
> At least I don't see a reason why we should call it unconditionally:)

Fixed.

marcus-h:
Merged. Thanks a lot!

Issue #965, opened 2021-10-20
unsupported format character 'B' (0x42) at index 66


mcepl:
Could it be because I am building from `devel:libraries:c_c++` (notice non-alphanumeric characters `+`)?

```
libgit2@stitny$ oscb --clean
Building libgit2.spec for openSUSE_Factory/x86_64
Run source service: /usr/lib/obs/service/format_spec_file --outdir /home/matej/build/home:mcepl:branches:devel:libraries:c_c++/libgit2/tmphkl0g84r.format_spec_file.service
Run source service: /usr/lib/obs/service/source_validator --outdir /home/matej/build/home:mcepl:branches:devel:libraries:c_c++/libgit2/tmptc2onr7b.source_validator.service
- package has baselibs.conf: (new or modified)
--------------------------------------------------------------
libgit2-1_3
--------------------------------------------------------------
###ASK /home/matej/build/home:mcepl:branches:devel:libraries:c_c++/libgit2/baselibs.conf
Getting buildconfig from server and store to /home/matej/build/home:mcepl:branches:devel:libraries:c_c++/libgit2/.osc/_buildconfig-openSUSE_Factory-x86_64
Getting buildinfo from server and store to /home/matej/build/home:mcepl:branches:devel:libraries:c_c++/libgit2/.osc/_buildinfo-openSUSE_Factory-x86_64.xml
Updating cache of required packages

The build root needs packages from project 'devel:libraries:c_c++'.
Note that malicious packages can compromise the build result or even your system.
Would you like to ...
0 - quit (default)
1 - always trust packages from 'devel:libraries:c_c++'
2 - trust packages just this time
? 2
Traceback (most recent call last):
  File "/home/matej/.local/bin/osc", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/matej/archiv/knihovna/repos/osc/osc-wrapper.py", line 45, in <module>
    r = babysitter.run(osccli)
  File "/home/matej/archiv/knihovna/repos/osc/osc/babysitter.py", line 69, in run
    return prg.main(argv)
  File "/home/matej/archiv/knihovna/repos/osc/osc/cmdln.py", line 347, in main
    return self.cmd(args)
  File "/home/matej/archiv/knihovna/repos/osc/osc/cmdln.py", line 370, in cmd
    retval = self.onecmd(argv)
  File "/home/matej/archiv/knihovna/repos/osc/osc/cmdln.py", line 504, in onecmd
[...]

PR #980, opened 2021-12-07
Escape % character in binary download URLs

marcus-h:
Without escaping the % character, the download URL could be subject
to string formatting (depending on the subsequent characters). For
instance, if the url attribute's value of a buildinfo's path element
contains the substring "c_c%2B%2B", the "%2B" is interpreted as a
format string (see issue #965), which is wrong ("B" is not a valid
format character at all). In order to avoid this, escape all
% characters in the download urls.
Note: escaping the % characters in the download url itself is OK
because we only intend to "format" the path.

Note: we do not escape the % characters for urls from the config
file (implicit assumption: the user already correctly escaped the
urls (whether this assumption is sensible or not is debatable, of
course)).

Fixes: #965 ("unsupported format character 'B' (0x42) at index 66")

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -180,7 +180,8 @@ class Buildinfo:
             self.pathes.append(node.get('project')+"/"+node.get('repository'))
             # a hash providing the matching URL for specific repos for newer OBS instances
             if node.get('url'):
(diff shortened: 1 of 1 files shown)


lethliel:
LGTM

Issue #972, opened 2021-11-16
Resize of terminal while osc getbinaries makes osc crash


zaitor:
Resize of (gnome)-terminal window, while a osc getbinaries is in progress, kills osc/download

```
user@world:~> osc getbinaries --debug --destdir=/some/random/writeable/dir/ openSUSE:Factory:Staging:J/kernel-default/standard/x86_64
_buildenv: 100%|#####################################################################| Time: 0:00:00
_statistics: 100%|###################################################################| Time: 0:00:00
kernel-default-5.15.2-1.2.x86_64.rpm: 100%|##########################################| Time: 0:00:19
Traceback (most recent call last):2.x86_64.rpm:   2%|                                | ETA:  0:03:22
  File "/usr/bin/osc", line 45, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.8/site-packages/osc/babysitter.py", line 69, in run
    return prg.main(argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 347, in main
    return self.cmd(args)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 370, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 504, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 1235, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.8/site-packages/osc/commandline.py", line 7519, in do_getbinaries
    get_binary_file(apiurl,
  File "/usr/lib/python3.8/site-packages/osc/core.py", line 4714, in get_binary_file
    download(u, target_filename, progress_obj, target_mtime)
  File "/usr/lib/python3.8/site-packages/osc/core.py", line 4672, in download
    for buf in streamfile(url, http_GET, BUFSIZE, progress_obj=progress_obj):
  File "/usr/lib/python3.8/site-packages/osc/core.py", line 6123, in streamfile
    data = xread(bufsize)
  File "/usr/lib64/python3.8/tempfile.py", line 473, in func_wrapper
    return func(*args, **kwargs)
  File "/usr/lib64/python3.8/socket.py", line 669, in readinto
    return self._sock.recv_into(b)
  File "/usr/lib64/python3.8/http/client.py", line 503, in readinto
    n = self.fp.readinto(b)
[...]

PR #987, opened 2021-12-27
Fix crash on terminal resize during download

dmach:
Resolves https://github.com/openSUSE/osc/issues/972

### DIFF
--- a/osc/babysitter.py
+++ b/osc/babysitter.py
@@ -57,9 +57,6 @@ for name in 'SIGBREAK', 'SIGHUP', 'SIGTERM':
     if num:
         signal.signal(num, catchterm)
 
-# Signals which should be ignored
-for sig in (signal.SIGWINCH,):
-    signal.signal(sig, signal.SIG_IGN)
 
 def run(prg, argv=None):
     try:
--- a/osc/meter.py
+++ b/osc/meter.py
@@ -3,6 +3,9 @@
 # and distributed under the terms of the GNU General Public Licence,
 # either version 2, or (at your option) any later version.
 
+
+import signal
+
 try:
     import progressbar as pb
     have_pb_module = True
@@ -23,6 +26,11 @@ class PBTextMeter(object):
                 # a ZeroDivisionException
                 widgets.insert(1, pb.Percentage())
             self.bar = pb.ProgressBar(widgets=widgets, maxval=size)
+        # When a signal handler is set, it resets SA_RESTART flag
+        # - see signal.siginterrupt() python docs.
+        # ProgressBar's constructor sets signal handler for SIGWINCH.
+        # So let's make sure that it doesn't interrupt syscalls in osc.
+        signal.siginterrupt(signal.SIGWINCH, False)
         self.bar.start()
 
     def update(self, amount_read):


marcus-h:
On 2021-12-27 06:22:09 -0800, Daniel Mach wrote:
> diff --git a/osc/meter.py b/osc/meter.py
> index 3372a1a1..fa7a99a0 100644
> --- a/osc/meter.py
> +++ b/osc/meter.py

> @@ -10,19 +16,49 @@
>      have_pb_module = False
>
>
> +if have_pb_module:
> +    class ProgressBar(pb.ProgressBar):
> +        """
> +        Modified ProgressBar that sets term_width argument to avoid setting SIGWINCH signal handler.
> +
> +        Background:
> +        We're doing this because the download routines are unable to restart interrupted syscalls.
> +        See signal(7) for more details:
> +        Interruption of system calls and library functions by signal handlers

Hmm it is rather due to a "bug" in m2crypto: m2.ssl_read returns None
if the SSL connection has a specific state and a read(...) syscall was
interrupted, which is indirectly issued by SSL_read(...). It is just
that M2Crypto.SSL.Connection.recv_into does not handle None.

(In our specific blocking IO case, the desired behavior would probably
be to call m2.ssl_read again (until it returns someting != None).)

> +        """
> +
> +        def __init__(self, *args, **kwargs):
> +            # set term_width to avoid auto-detection in progressbar's code
> +            # that sets SIGWINCH signal handler
> +            kwargs["term_width"] = self._get_term_width()
> +            super(ProgressBar, self).__init__(*args, **kwargs)
> +
> +        def _get_term_width(self):
> +            try:
> +                _, width = array.array('h', fcntl.ioctl(sys.stderr, termios.TIOCGWINSZ, '\0' * 8))[:2]

Hmm the winsize struct consists of unsigned short members. Strictly
speaking, we should use "H" instead of "h". We could also use
struct.calcsize to compute the "correct" size of the winsize struct

size = struct.calcsize('4H')
winsize = fcntl.ioctl(sys.stderr, termios.TIOCGWINSZ, '\0' * size)
width = struct.unpack('4H', winsize)[1]

(just an idea...)

> +            except Exception:

Hmm what about only catching OSError?

> +                width = self._env_size()
> +            return width
> +
> +        def update(self, *args, **kwargs):
> +            # recompute term_width in case the term width has changed
> +            self.term_width = self._get_term_width()

Hmm we probably end up with "a lot" of unnecessary ioctl syscalls:/ but
I guess that's OK...

Otherwise, nice patch!:)

dmach:
@marcus-h all right, after reading your comment I've started digging in how signals work in Python and realized, that there could be a way simpler patch. Please check the new version and let me know if you have any concerns.

marcus-h:
On 2022-01-03 02:52:24 -0800, Daniel Mach wrote:
> @marcus-h all right, after reading your comment I've started digging in how signals work in Python and realized, that there could be a way simpler patch. Please check the new version and let me know if you have any concerns.
>
Cool! LGTM.

Minor: while you are at it, what about removing the (useless)
SIGWINCH stuff from the babysitter? (Feel free to ignore:) )

dmach:
I've removed the useless SIGWINCH stuff in another commit in this PR.
Just in case, I've manually tested that osc works in quiet mode as expected (a different code path than the default SIGWINCH handling).

marcus-h:
On 2022-01-07 12:55:02 +0000, Daniel Mach wrote:
> I've removed the useless SIGWINCH stuff in another commit in this PR.
> Just in case, I've manually tested that osc works in quiet mode as expected (a different code path than the default SIGWINCH handling).
>
Merged. Thanks a lot!

PR #991, opened 2022-01-19
Replace unmaintained cmdln.py with a custom code

dmach:
Cmdln.py is unmaintained for years and uses deprecated optparse.
Let's replace it with a simpler custom code. Also remove code
that generates man page, we'll replace it with a 3rd party tool.

:information_source: The commandline code will probably evolve to something else (class per command?) over time, but this is meant as a drop-in replacement for the current cmdln.py.

- [x] Fix TODOs in the code.
- [x] Properly test this change as **anything** can break.
- [x] Generate the osc man page (used argparse-manpage; see the included spec file)
- [X] Implement `osc help` and `osc help <command>`


osc/commandline.py
```diff
@@ -7892,7 +7746,8 @@ def do_my(self, subcmd, opts, *args):
     @cmdln.alias('se')
     @cmdln.alias('bse')
     def do_search(self, subcmd, opts, *args):
-        """${cmd_name}: Search for a project and/or package.
+        """
+        Search for a project and/or package
```
abitrolly, line 7750: Maybe commit docstings separately to keep diff small, as they don't seem related.

dmach: I understand that it's difficult to review, but the change is massive anyway.
I'll probably keep it as it is.

Issue #964, opened 2021-10-18
download uses os.fchmod, not available on Windows


jayvdb:
Similar to https://github.com/openSUSE/osc/issues/948 , [`os.fchmod`](https://docs.python.org/3/library/os.html#os.fchmod) is not available on Windows.

jayvdb:
Also 

```py
Traceback (most recent call last):
  File "C:/Users/vandjohn/Downloads/WPy64-3771/python-3.7.7.amd64/Scripts/osc-wrapper.py", line 10, in <module>
    from osc import commandline, babysitter
  File "C:\Users\vandjohn\Downloads\WPy64-3771\python-3.7.7.amd64\lib\site-packages\osc\babysitter.py", line 60, in <module>
    for sig in (signal.SIGWINCH,):
AttributeError: module 'signal' has no attribute 'SIGWINCH'
```

dmach:
When python 2.7 support is dropped, we can switch to `os.chmod()`

From python docs:
```
os.fchmod(fd, mode)
 As of Python 3.3, this is equivalent to os.chmod(fd, mode)
```

`signal.SIGWINCH` will probably need some special handling

Issue #963, opened 2021-10-18
get_source_file fails on Windows


jayvdb:
A lot of tests fail in `get_source_file` on Windows with `TypeError: a bytes-like object is required, not 'str'`

Probably this is because on Windows, `locale.getpreferredencoding()` is `cp1252`

```
======================================================================
ERROR [0.000s]: testDiffRemoteExistingLocalNotExisting (test_difffiles.TestDiffFiles)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\vandjohn\gh\osc\tests\common.py", line 160, in wrapped_test_method
    test_method(*args)
  File "C:\Users\vandjohn\gh\osc\tests\common.py", line 160, in wrapped_test_method
    test_method(*args)
  File "C:\Users\vandjohn\gh\osc\tests\common.py", line 160, in wrapped_test_method
    test_method(*args)
  File "C:\Users\vandjohn\gh\osc\tests\test_difffiles.py", line 254, in testDiffRemoteExistingLocalNotExisting
    self.__check_diff(p, exp, 3)
  File "C:\Users\vandjohn\gh\osc\tests\test_difffiles.py", line 301, in __check_diff
    for i in p.get_diff(revision):
  File "C:\Users\vandjohn\gh\osc\osc\core.py", line 2086, in get_diff
    get_source_file(self.apiurl, self.prjname, self.name, f.name, tmpfile, revision)
  File "C:\Users\vandjohn\gh\osc\osc\core.py", line 4733, in get_source_file
    u = makeurl(apiurl, ['source', prj, package, pathname2url(filename.encode(locale.getpreferredencoding(), 'replace'))], query=query)
  File "C:\Users\vandjohn\Downloads\WPy64-3771\python-3.7.7.amd64\lib\nturl2path.py", line 53, in pathname2url
    if not ':' in p:
TypeError: a bytes-like object is required, not 'str'

======================================================================
ERROR [0.365s]: testDiffRemoteMissingLocalDeleted (test_difffiles.TestDiffFiles)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\vandjohn\gh\osc\tests\common.py", line 160, in wrapped_test_method
    test_method(*args)
  File "C:\Users\vandjohn\gh\osc\tests\test_difffiles.py", line 297, in testDiffRemoteMissingLocalDeleted
    self.__check_diff(p, '', 3)
[...]

PR #961, opened 2021-10-18
assertEqualMultiline: Ignore Windows EOL

jayvdb:
This allows many tests to pass on Windows.

PR #995, opened 2022-01-26
Drop Windows from supported operating systems

dmach:
The ':' character is used as a separator in Open Build Service
and constantly appears in directory names after running osc commands.
Windows do not support ':' as a valid character on file system.
This breaks not only osc but also basic commands such
as 'git clone' on a project that contains colons in paths.
    
That's why we decided to make osc unsupported on Windows.

After merging this, I'd close the following issues and PRs:
https://github.com/openSUSE/osc/issues/960
https://github.com/openSUSE/osc/issues/963
https://github.com/openSUSE/osc/issues/964
https://github.com/openSUSE/osc/pull/961
https://github.com/openSUSE/osc/pull/962

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -4653,8 +4653,6 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             prj_dir = opts.output_dir if opts.output_dir else project
             if not opts.output_dir and conf.config['checkout_no_colon']:
                 prj_dir = prj_dir.replace(':', '/')
-            if sys.platform[:3] == 'win':
-                prj_dir = prj_dir.replace(':', ';')
             if os.path.exists(prj_dir):
                 sys.exit('osc: project directory \'%s\' already exists' % prj_dir)
 
--- a/osc/core.py
+++ b/osc/core.py
@@ -4082,8 +4082,6 @@ def _get_linux_distro():
 
 def get_default_editor():
     system = platform.system()
-    if system == 'Windows':
-        return 'notepad'
     if system == 'Linux':
         dist = _get_linux_distro()
         if dist == 'debian':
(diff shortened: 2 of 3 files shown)


marcus-h:
On 2022-01-26 06:17:32 -0800, Daniel Mach wrote:
> diff --git a/setup.py b/setup.py
> index 118e8a5d..5f0abdb1 100755
> --- a/setup.py
> +++ b/setup.py
> @@ -113,7 +113,7 @@ def finalize_options(self):
>      long_description_content_type="text/plain",
>      author='openSUSE project',
>      ***@***.***',
> -    license='GPL',
> +    license='GPLv2+',

Minor: what about using an SPDX license identifier like "GPL-2.0-or-later"?
(Unless this is frowned upon in setup.py)

dmach:
> Minor: what about using an SPDX license identifier like "GPL-2.0-or-later"?
I'd keep it as it is for now, because it's consistent with the license in `classifiers`.

We probably could omit the license completely in favor of classifiers:
https://docs.python.org/3/distutils/setupscript.html#additional-meta-data - Note#5.
But I think it's better to be explicit.

dmach:
Update: I've realized that setup.py still contained windows specific code which I have removed, amended to the last commit and force-pushed.

marcus-h:
On 2022-01-27 06:16:43 -0800, Daniel Mach wrote:
> Update: I've realized that setup.py still contained windows specific code which I have removed, amended to the last commit and force-pushed.
>
Hmm if we are "actively" dropping windows support, we should probably
remove it from the commandline and core modules, too. Namely,
- core.get_default_editor
- core.get_default_pager
- core.checkout_package
- commandline.Osc.do_checkout

What do you think?

dmach:
> What do you think?

I've removed the code as you suggested.

lethliel:
LGTM and as already discussed we want to drop Windows support here

PR #996, opened 2022-01-27
COPYING: Use the latest version from gnu.org

dmach:
https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt

The original version is too old, it's even referring to the GNU **Library** General Public License (which was renamed to **Lesser** years ago).


lethliel:
LGTM
Thanks for the PR

PR #998, opened 2022-01-30
add support for building preinstall images

dirkmueller:
This is yet another build type supported by the obs-build
script intended to build images that speed up the build setup
phase.

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -122,7 +122,7 @@ class Buildinfo:
         if self.buildtype == 'livebuild':
             self.pacsuffix = 'deb'
         if self.buildtype == 'snapcraft':
-            # atm ubuntu is used as base, but we need to be more clever when 
+            # atm ubuntu is used as base, but we need to be more clever when
             # snapcraft also supports rpm
             self.pacsuffix = 'deb'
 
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2022-01-30 15:21:17 -0800, Dirk Mueller wrote:
...
> +    if build_as_user() or build_type == 'preinstallimage':
>         # preinstallimage extraction will fail
>         bi.preinstallimage = None

Hmm why is this needed?

Btw., could you also add the missing "_preinstallimage" pieces to
"def parse_repoarchdescr" in commandline.py? (Just to support the
"_preinstallimage" file autodetection.)

Thanks a lot for the patch!

dirkmueller:
>  +    if build_as_user() or build_type == 'preinstallimage':
>         # preinstallimage extraction will fail
>         bi.preinstallimage = None
> Hmm why is this needed?

when you fetch the preinstallimage, it will be unpacked and then repackaged into the preinstallimage making it twice as big as before. server side there must be  a hack somewhere to build with --nopreinstallimage --clean and this is supposed to fake that on local builds.

dirkmueller:
@marcus-h  any remaining concern?

dirkmueller:
@dmach  could you please review this as well? it blocks the "helm building" support that I'd really like to get in.

dmach:
> > * if build_as_user() or build_type == 'preinstallimage':
> >   # preinstallimage extraction will fail
> >   bi.preinstallimage = None
> >   Hmm why is this needed?
> 
> when you fetch the preinstallimage, it will be unpacked and then repackaged into the preinstallimage making it twice as big as before. server side there must be a hack somewhere to build with --nopreinstallimage --clean and this is supposed to fake that on local builds.

Could you add this or a similar explanation to the code?
It would be good to keep it there, because nobody will probably track it to this PR when trying to understand what the code does.
Looks good to me otherwise.

PR #1002, opened 2022-02-12
Enable unlocking and creation of gnone keyring when running headless

matthewdva:
This patch allows osc to unlock and create if necessary a GNOME keyring when running on a headless system

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -75,6 +75,36 @@ GNOME_KEYRING = False
 try:
     import keyring
     GENERIC_KEYRING = True
+    try:
+        import secretstorage
+        if 'DISPLAY' in os.environ:
+            # Looks lie we have desktop session
+            # Continue to the normal desktop unlock procedures.
+            pass
+        if 'DBUS_SESSION_BUS_ADDRESS' in os.environ:
+            # Ensure DBUS has launced for headless GNOME_KEYRING.
+            pass
+        else:
+            raise SystemExit ("""
+The DBUS_SESSION_BUS_ADDRESS environment variable s not set which
+probably means that DBUS is not running.
+
+If you have just installed osc or gnome keyring then you need to
+logout and login again, or source /etc/profile.
+""")
+        if 'GNOME_KEYRING_CONTORL' in os.environ:
+            # Ensure the GNOME_KEYRING is started.
+            pass
+        else:
+            raise SystemExit ("""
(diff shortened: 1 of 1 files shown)


marcus-h:
On 2022-02-12 14:54:21 -0800, matthewdva wrote:
> This patch allows osc to unlock and create if necessary a GNOME keyring when running on a headless system

I just had a quick glance at your patch... I'm not really sure if this
is the right approach for fixing the gnome keyring issue. See below.

Excerpt from your patch:

> @@ -853,7 +917,16 @@ def add_section(filename, url, user, passwd, creds_mgr_descriptor=None):
>          creds_mgr = creds_mgr_descriptor.create(cp)
>      else:
>          creds_mgr = _get_credentials_manager(url, cp)
> -    creds_mgr.set_password(url, user, passwd)
> +    try:
> +        creds_mgr.set_password(config['apiurl'], config['user'], config['pass'])
> +    except keyring.errors.InitError:
> +        print ('Creating a new keyring.')
> +        _unlock_gnome_keyring('New keyring password: ')

Hmm this does not seem to be correct. Other python-keyring backends
could also raise keyring.errors.InitError. That is, unconditionally
trying to unlock the gnome keyring is probably not a good idea.

> +        creds_mgr.set_password(config['apiurl'], config['user'], config['pass'])
> +    except keyring.eror.KeyringLocked:

Typo?

> +        print ('Unlock keyring.')
> +        _unlock_gnome_keyring('Keyring password: ')

As above.

> +        creds_mgr.set_password(config['apiurl'], config['user'], config['pass'])
>      write_config(filename, cp)


Why isn't the existing unlocking code in keyring/backends/SecretService.py
(see the python-keyring package) sufficient? IMHO, it would be more
appropriate to enhance the existing python-keyring backend instead.
What do you think?

matthewdva:
Thank you for looking at this so quickly
      This is patch I have been using for years privately.  As soon as I submitted the PR it occurred to me that this was not the correct approach.  Let cancel this PR and I will rework it in the correct spot (hopefully).  I sincerely apologize for the noise this weekend.

marcus-h:
On 2022-02-12 18:10:54 -0800, matthewdva wrote:
> Let cancel this PR and I will rework it in the correct spot (hopefully).

Cool!

> I sincerely apologize for the noise this weekend.
>
There is no need to apologize! A pull request is always welcome (regardless
whether it is eventually merged or not)!:)

PR #1003, opened 2022-02-17
Run tests via calling 'setup.py test'

dmach:


dmach:
I just realized - suite.py used `XMLTestRunner` when it was available.
Is it something we want to preserve?

lethliel:
LGTM and I like this approach too

Issue #495, opened 2018-12-14
osc getbinaries at project level for multibuild


darix:
Follow up for 
https://github.com/openSUSE/open-build-service/issues/6199

nickbroon:
Any progress on this? `osc ls -b` also does not list multibuild binaries.

PR #1006, opened 2022-02-23
Fix getbinaries command to fetch also multibuild packages

dmach:
Fixes: https://github.com/openSUSE/osc/issues/495
Related to: https://github.com/openSUSE/open-build-service/issues/9715


dmach:
I've tested this from nearly all perspectives, I believe it works as expected now.
FTR, the tests currently live here: https://github.com/dmach/osc/tree/behave/behave/features (getbinaries*.feature),
they'll be evetually moved into upstream osc git repo.

PR #1009, opened 2022-03-02
Support flavors in aggregatepac

dmach:


dmach:
I believe I've tested this sufficiently and there's no feedback for 3 months. Merging.

PR #1010, opened 2022-03-04
Allow formatting of the sccache uri

Firstyear:
This allows a format to be specified in the sccache uri, specifically
the file uri so that a per-package cache can be created. This way
an osc build locally doesn't ruin your cache moving between different
packages.

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -682,16 +682,6 @@ def main(apiurl, opts, argv):
     if opts.pkg_ccache:
         buildargs.append('--pkg-ccache=%s' % opts.pkg_ccache)
         xp.append('ccache')
-    if opts.sccache_uri or config['sccache_uri'] or opts.sccache or config['sccache']:
-        if opts.pkg_ccache or opts.ccache or config['ccache']:
-            raise oscerr.WrongArgs('Error: sccache and ccache can not be enabled at the same time')
-        if opts.sccache_uri:
-            buildargs.append('--sccache-uri=%s' % opts.sccache_uri)
-        elif config['sccache_uri']:
-            buildargs.append('--sccache-uri=%s' % config['sccache_uri'])
-        else:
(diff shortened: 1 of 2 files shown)


dmach:
The code looks good to me (although I haven't tested it). Merging.

marcus-h:
On 2022-03-03 20:12:01 -0800, Firstyear wrote:
> index 8796c36b..ff3d3155 100644
> --- a/osc/build.py
> +++ b/osc/build.py

<SNIP>

> +    # We configure sccache after pacname, so that in default cases we can have an sccache for each
> +    # package to prevent cross-cache polutions. It helps to make the local-use case a bit nicer.
> +    if opts.sccache_uri or config['sccache_uri'] or opts.sccache or config['sccache']:
> +        if opts.pkg_ccache or opts.ccache or config['ccache']:
> +            raise oscerr.WrongArgs('Error: sccache and ccache can not be enabled at the same time')
> +        sccache_arg = "--sccache-uri=/var/tmp/osbuild-sccache-{pkgname}.tar"
> +        if opts.sccache_uri:
> +            sccache_arg = '--sccache-uri=%s' % opts.sccache_uri
> +        elif config['sccache_uri']:
> +            sccache_arg = '--sccache-uri=%s' % config['sccache_uri']
> +        # Format the package name.
> +        sccache_arg = sccache_arg.format(pkgname=pacname)

Hmm the new {pkgname} syntax clashes with our existing %(package)s syntax,
which can be, for instance, used in the build-root config option (maybe
we should/could also take %(project)s etc. into account?).

> +        buildargs.append(sccache_arg)
> +        xp.append('sccache')
> +

mlschroe:
Could you please keep the osc options somewhat in sync with the official options from the 'build' tool? The '--sccache*' versions should not be used when calling 'build'.

Issue #1014, opened 2022-03-21
0.176  is not to download some important files anymore


adrianschroeter:
https://github.com/openSUSE/osc/commit/08a70952c05f745f1c56ee04740c589f0f89585b

basically disables all additional files of build results. This breaks quite some usecases as for example also the _buildenv file is skipped, so you can not reproduce any former build anymore.

IMHO we should:

1) add an option to allow all build results
2) only block known meta files, we do not want have by default, but not everything by default.
3) _buildenv is IMHO definitive important enough to get downloaded by default, not only for reproducibility rebuilds, but also for analyzation tasks.

Currently this breaks reproductibility rebuilds of @bmwiedemann.

PR #1015, opened 2022-03-21
Fix getbinaries command by downloading files in subdirs named by the packages

dmach:


dmach:
Fixes https://github.com/openSUSE/osc/issues/1014

Examples of how the downloaded packages are placed on disk
 - Packages get downloaded into subdirs when a package is **not** known: [here](https://github.com/dmach/osc/blob/behave/behave/features/getbinaries-project-repo-arch.feature)
 - The subdirs do not get created when a package is known: [here](https://github.com/dmach/osc/blob/behave/behave/features/getbinaries-project-package-repo-arch.feature)

@adrianschroeter @bmwiedemann could you check if this meets your expectations?

adrianschroeter:
works for me

marcus-h:
On 2022-03-21 16:22:24 +0000, Daniel Mach wrote:
> index 26bc277f..7b9ae196 100644
> --- a/osc/commandline.py
> +++ b/osc/commandline.py
> @@ -7518,8 +7518,10 @@ def do_getbinaries(self, subcmd, opts, *args):
>
>
>          if package is None:
> +            package_specified = False
>              package = meta_get_packagelist(apiurl, project, deleted=0)
>          else:
> +            package_specified = True
>              if opts.multibuild_package:
>                  packages = []
>                  for subpackage in opts.multibuild_package:
> @@ -7554,7 +7556,13 @@ def do_getbinaries(self, subcmd, opts, *args):
>                              continue
>                          if i.name.find('-debugsource-') >= 0:
>                              continue
> -                    fname = '%s/%s' % (target_dir, i.name)
> +
> +                    if package_specified:
> +                        fname = '%s/%s' % (target_dir, i.name)
> +                    else:
> +                        # download into a subdir to avoid overwriting logs and metadata
> +                        fname = '%s/%s/%s' % (target_dir, pac, i.name)
> +
Hmm while this "fix" is reasonable wrt. the "meta" files etc. it
might break existing use cases/scripts where people just care about
the binaries itself and expect them in target_dir.

>                      if os.path.exists(fname):
>                          st = os.stat(fname)
>                          if st.st_mtime == i.mtime and st.st_size == i.size:
>

lethliel:
The use cases / scripts just break if the getbinaries command is issued on project level...

marcus-h:
On 2022-03-30 01:33:44 -0700, Marco Strigl wrote:
> The use cases / scripts just break if the getbinaries command is issued on project level...
>
Huh? At least

$> cd openSUSE:Tools
$> osc getbinaries openSUSE_Tumbleweed x86_64

seems to work for me.

dmach:
@lethliel @marcus-h I've made another change: when downloading a project, `*.log` and `_*` get downloaded into subdirs, everything else into the target dir as usual. If this doesn't work for you, please note that I ran out of ideas and I'll need exact specification of the desired behavior from you :)

lethliel:
> On 2022-03-30 01:33:44 -0700, Marco Strigl wrote: The use cases / scripts just break if the getbinaries command is issued on project level...
> Huh? At least $> cd openSUSE:Tools $> osc getbinaries openSUSE_Tumbleweed x86_64 seems to work for me.

No that is not what i meant. The change breaks only the expected behavior on project level due to the insertion of a directory..

lethliel:
This looks good to me..

Issue #1016, opened 2022-03-22
Downloads often fail hdrmd5 checks due to dod urls


Firstyear:
It's extremely common to see:

[46/68] (SUSE:SLE-15:Update) libldap-data-2.4.46-9.64.1.noarch.rpm: 100%|...| Time: 0:00:00
SUSE:SLE-15:Update/libldap-data: attempting download from api, since the hdrmd5 did not match

This is caused by .osc/_buildinfo-standard-x86_64.xml which has an invalid hdrmd5 contained (d0d0dd0d0d0d ... I think). I assume this is meant to mean it's a place-holder md5, but in reality what this causes is that when attempting to download from a mirror, these downloads will ALWAYS fail because the downloaded file (which is valid) does not match what the hdrmd5 in the xml has. This then forces the downloaded files to be discarded, and the buildservice api is used instead, which is MUCH slower compared to using the mirrors (especially when you live in Australia).

dmach:
Thanks for the report, the `d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0` hash really seems to have a special meaning in OBS.
@Firstyear could you provide a reproducer?

adrianschroeter:
On Dienstag, 22. März 2022, 09:23:46 CET Daniel Mach wrote:
> Thanks for the report, the `d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0` hash really seems to have a special meaning in OBS.
> @Firstyear could you provide a reproducer?

it references to not yet downloaded packages configured via download on demand repositories.

-- 

Adrian Schroeter ***@***.***>
Build Infrastructure Project Manager

SUSE Software Solutions Germany GmbH, Maxfeldstr. 5, 90409 Nürnberg, Germany
(HRB 36809, AG Nürnberg) Geschäftsführer: Ivo Totev

Firstyear:
@dmach I'm not sure of the exact reproduction steps, beside "I just use osc and it does this about 1/3rd of the time". About the only special setup I have is that I'm using the option:

```
[https://api.opensuse.org]
downloadurl = https://mirrorcache-au.opensuse.org
```

This is because downloads in Australia are significantly faster with a local mirror. So I'm assuming it's some interaction there where the fact that the downloaduri option turns of api downloads (but so does download_api_only = 0 I think). 

[...]

PR #1019, opened 2022-03-29
Fix hdrmd5 check of downloaded packages from DoD repos

dmach:


marcus-h:
On 2022-03-29 08:00:00 +0000, Daniel Mach wrote:
> index dc18c939..6ed67340 100644
> --- a/osc/fetch.py
> +++ b/osc/fetch.py
> @@ -262,11 +262,13 @@ def run(self, buildinfo):
>                          # if the checksum of the downloaded package doesn't match,
>                          # delete it and mark it for downloading from the API
>                          hdrmd5 = packagequery.PackageQuery.queryhdrmd5(i.fullfilename)
> -                        if not hdrmd5 or hdrmd5 != i.hdrmd5:
> +                        # packages with hdrmd5 == 'd0d...' come from 'download on demand' repos
> +                        # and their checksum is not known to the server yet, so we skip their checksum check
> +                        if not hdrmd5 or (hdrmd5 != i.hdrmd5 and i.hdrmd5 != 'd0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0'):
>                              print('%s/%s: attempting download from api, since the hdrmd5 did not match'
>                                  % (i.project, i.name))

Hmm we should probably not directly skip the check in case of dod.
What we could do is:
- skip check and always fetch the dod file from the API (assumption: API
  always delivers a "correct" file)

> -                        os.unlink(i.fullfilename)
> -                        self.__add_cpio(i)
> +                            os.unlink(i.fullfilename)
> +                            self.__add_cpio(i)
>
Hmm does this work as expected? From a quick glance, it should fail
in build.py then.

>                  except KeyboardInterrupt:
>                      print('Cancelled by user (ctrl-c)')

dmach:
> Hmm we should probably not directly skip the check in case of dod. What we could do is: - skip check and always fetch the dod file from the API (assumption: API always delivers a "correct" file)

Maybe it's the right way, but the original report (https://github.com/openSUSE/osc/issues/1016) is about avoiding API downloads and preferring the local mirror.

> Hmm does this work as expected? From a quick glance, it should fail
in build.py then.

I'm not sure about it.

The original code is:
```
if not os.path.isfile(i.fullfilename):
    # package doesn't exist -> download from API
else:
    # package exists (was downloaded)
    if <file has invalid md5hdr>
        # print warning
    # always remove the file and download the API - this should happen only if checksum is incorrect
    os.unlink(i.fullfilename)
    self.__add_cpio(i)
```
Fixing indentation fixes the code, I believe.

marcus-h:
On 2022-03-29 11:14:45 +0000, Daniel Mach wrote:
> > Hmm we should probably not directly skip the check in case of dod. What we could do is: - skip check and always fetch the dod file from the API (assumption: API always delivers a "correct" file)
>
> Maybe it's the right way, but the original report (https://github.com/openSUSE/osc/issues/1016) is about avoiding API downloads and preferring the local mirror.
>
We could also use the binaryversions view to retrieve the hdrmd5:
GET /build/project/repo/arch/_repository?view=binaryversions&binary=foo&binary=bar

In case of a dod binary, the backend will fetch the binary and return the
correct hdrmd5.

> > Hmm does this work as expected? From a quick glance, it should fail
> in build.py then.
>
> I'm not sure about it.
>
> The original code is:
> ```
> if not os.path.isfile(i.fullfilename):
>     # package doesn't exist -> download from API
> else:
>     # package exists (was downloaded)
>     if <file has invalid md5hdr>
>         # print warning
>     # always remove the file and download the API - this should happen only if checksum is incorrect
>     os.unlink(i.fullfilename)
>     self.__add_cpio(i)
> ```
> Fixing indentation fixes the code, I believe.
>
No, the hdrmd5 check in build.py should still fail.

Firstyear:
Seems to have done the trick! Can I suggest though:

```
diff --git a/osc/fetch.py b/osc/fetch.py
index 6ed67340..4cd31d03 100644
--- a/osc/fetch.py
+++ b/osc/fetch.py
@@ -265,8 +265,8 @@ class Fetcher:
                         # packages with hdrmd5 == 'd0d...' come from 'download on demand' repos
                         # and their checksum is not known to the server yet, so we skip their checksum check
                         if not hdrmd5 or (hdrmd5 != i.hdrmd5 and i.hdrmd5 != 'd0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0'):
-                            print('%s/%s: attempting download from api, since the hdrmd5 did not match'
-                                % (i.project, i.name))
+                            print('%s/%s: attempting download from api, since the hdrmd5 did not match - %s != %s'
+                                % (i.project, i.name, hdrmd5, i.hdrmd5))
                             os.unlink(i.fullfilename)
                             self.__add_cpio(i)
```

This way it is a bit easier to see what the checksums are that are failing?

Firstyear:
So it gets weirder. I added that output and I see a lot of :

```
[117/273] (SUSE:SLE-15-SP1:Update) python3-wheel-0.32.3-6.2.12.noarch.rpm: 100%|#############################################################################################################################################| Time: 0:00:00
SUSE:SLE-15-SP1:Update/python3-wheel: attempting download from api, since the hdrmd5 did not match - 4690f188993e8c1afcd58204fc72e011 != d667e91cfc3715f49065299ca120033f
[118/273] (SUSE:SLE-15-SP4:GA) python3-ordered-set-4.0.2-150400.1.1.noarch.rpm: 100%|########################################################################################################################################| Time: 0:00:00
SUSE:SLE-15-SP4:GA/python3-ordered-set: attempting download from api, since the hdrmd5 did not match - db3290f03d73addd70b627de087c9f7c != 902911f764124b05001db0ae12ab22ca
```

It looks like the *internal* mirrors are hosting invalid content in somecases. Is something weird going on where IBS/OBS are rebuilding packages and updating the hdrmd5, but then never pushing it to the mirrors because the file name is the same?

Out of 250 packages this happened 17 times, so about ~5% of packages are affected by this at a rough guess.

dirkmueller:
@dmach  rebase?

Thaodan:
> @dmach rebase?

I rebased the patch here:
https://github.com/Thaodan/osc/commits/fix/1016-hdrmd5-checks-dod

dmach:
I've improved the patch to correspond with the latest codebase.
The current code reports packages with `d0d...` hdrmd5 and still uses it, because there doesn't seem to be any better option. It is handled in all places where also `disable_hdrmd5_check` is handled - I hope this covers all possible scenarios.

@adrianschroeter could you take a look if the patch makes sense to you?

dmach:
We have discussed this in detail with @adrianschroeter and @mlschroe and decided to move this to open-build-service which should avoid sending the `d0d...` checksums in the first place.

dmach:
We have discussed this in detail with @adrianschroeter and @mlschroe and decided to move this to open-build-service which should avoid sending the `d0d...` checksums in the first place.

Thaodan:
Should this still be merged so osc will continue to work with older
versions of the obs serice to continue to work?

dmach:
> Should this still be merged so osc will continue to work with older versions of the obs serice to continue to work?

Couldn't you just use the new `disable_hdrmd5_check` option?
It allows keeping the downloaded file when the checksum doesn't match and that applies to the `d0d...` checksums as well.

Thaodan:
Sure however it is easier for users with the patch instead of first
having to discover the bug and then apply the option.

Firstyear:
> > Should this still be merged so osc will continue to work with older versions of the obs serice to continue to work?
> 
> Couldn't you just use the new `disable_hdrmd5_check` option? It allows keeping the downloaded file when the checksum doesn't match and that applies to the `d0d...` checksums as well.

The problem is that the hdrmd5 is broken by default, and the "config option" is hard to discover. You may need to actually say in the error message "to disable this check set X in Y.conf"

Issue #1018, opened 2022-03-24
osc ls -b can't cope with Unicode


jengelh:
```
(api.opensuse.org)

$ osc ls -b home:jengelh/test
Traceback (most recent call last):
  File "/usr/bin/osc", line 45, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.8/site-packages/osc/babysitter.py", line 67, in run
    return prg.main(argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 347, in main
    return self.cmd(args)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 370, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 504, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 1235, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.8/site-packages/osc/commandline.py", line 392, in do_list
    results.append((repo, get_binarylist(apiurl, project, repo.name, repo.arch, package=package, verbose=opts.verbose)))
  File "/usr/lib/python3.8/site-packages/osc/core.py", line 5719, in get_binarylist
    f = http_GET(u)
  File "/usr/lib/python3.8/site-packages/osc/core.py", line 3413, in http_GET
    def http_GET(*args, **kwargs):    return http_request('GET', *args, **kwargs)
  File "/usr/lib/python3.8/site-packages/osc/core.py", line 3404, in http_request
    return urlopen(req)
  File "/usr/lib64/python3.8/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python3.8/urllib/request.py", line 525, in open
    response = self._open(req, data)
  File "/usr/lib64/python3.8/urllib/request.py", line 542, in _open
    result = self._call_chain(self.handle_open, protocol, protocol +
  File "/usr/lib64/python3.8/urllib/request.py", line 502, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.8/site-packages/osc/oscssl.py", line 222, in https_open
    h.request(req.get_method(), request_uri, req.data, headers)
  File "/usr/lib64/python3.8/http/client.py", line 1256, in request
    self._send_request(method, url, body, headers, encode_chunked)
[...]

PR #1020, opened 2022-03-29
Quote unicode characters in URL path

dmach:
Also throw URLError if illegal characters are found in URL path.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -32,7 +32,7 @@ except ImportError:
     distro = None
 
 try:
-    from urllib.parse import urlsplit, urlunsplit, urlparse, quote_plus, urlencode, unquote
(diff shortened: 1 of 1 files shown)


dmach:
Fixes https://github.com/openSUSE/osc/issues/1018

lethliel:
LGTM. Thanks for the fix.

marcus-h:
On 2022-03-30 01:23:50 -0700, Marco Strigl wrote:
> LGTM. Thanks for the fix.
>
Arghs... this will break several codepaths that already do the
quoting on their own. Yes, this is a mess in osc. For instance,
now something like "osc co openSUSE:Tools osc" should break because
openSUSE:Tools is quoted twice...

PR #1021, opened 2022-04-01
Revert "Quote unicode characters in URL path"

lethliel:
Reverts openSUSE/osc#1020


lethliel:
1020 breaks already quoted calls of osc.core.makeurl because they are the quoted twice

PR #1022, opened 2022-04-04
Only prompt for a password if the server asks for it

mlschroe:
This replaces the weird "defer" hack in the credentials code
with a simpler idea: use a small proxy object for the password
and only ask the credential manager if the stringify method
is called.

In many cases the session cookie is already available, so there
is no need to ask for a password. Also, we can use a non-password
based authorization type if the server asks for it.


marcus-h:
On 2022-04-04 08:21:47 +0000, Michael Schroeder wrote:
> This replaces the weird &quot;defer&quot; hack in the credentials code
> with a simpler idea: use a small proxy object for the password
> and only ask the credential manager if the stringify method
> is called.
>
> In many cases the session cookie is already available, so there
> is no need to ask for a password.

Oh, good point!:)

> Also, we can use a non-password
> based authorization type if the server asks for it.

> index 07c6c967..0e7ed0cb 100644
> --- a/osc/conf.py
> +++ b/osc/conf.py

<SNIP>

> @@ -880,12 +880,26 @@ def _get_credentials_manager(url, cp):
>      return credentials.PlaintextConfigFileCredentialsManager(cp, None)
>
>
> -class APIHostOptionsEntry(dict):
> -    def __getitem__(self, key, *args, **kwargs):
> -        value = super(self.__class__, self).__getitem__(key, *args, **kwargs)
> -        if key == 'pass' and callable(value):
> -            value = value()
> -        return value
> +class APILazyPassword(object):
> +    """Delay prompting for a password until the server wants authorization.
> +       We only need to implement the __str__ method, as this is the only
> +       call done by the basic/digest authorization backends in requests.
> +    """
> +
> +    def __init__(self, url, conffile, user, creds_mgr):
> +        self.url = url
> +        self.conffile = conffile
> +        self.user = user
> +        self.creds_mgr = creds_mgr
> +        self.password = None
> +
> +    def __str__(self):
> +        if self.password is None:
> +            self.password = self.creds_mgr.get_password(self.url, self.user)
> +            if self.password is None:
> +                raise oscerr.ConfigMissingCredentialsError('No password found in section %s' % self.url, \
> +                                                           self.conffile, self.url)
> +        return self.password
>
Maybe we should make this a bit more robust because the current code
depends on an implementation detail in
AbstractBasicAuthHandler.retry_http_basic_auth: If, for instance,

    raw = "%s:%s" % (user, pw)

is changed to (for whatever stupid reason...)

    raw = user + ":" + pw

our code would fail (because __radd__ is not implemented).


>  def get_config(override_conffile=None,
> @@ -979,9 +993,6 @@ def get_config(override_conffile=None,
>          user = _extract_user_compat(cp, url, creds_mgr)
>          if user is None:
>              raise oscerr.ConfigMissingCredentialsError('No user found in section %s' % url, conffile, url)
> -        password = creds_mgr.get_password(url, user)
> -        if password is None:
> -            raise oscerr.ConfigMissingCredentialsError('No password found in section %s' % url, conffile, url)
>
Hmm that's "bad": if, for instance, the PlaintextConfigFileCredentialsManager
is used and there is no "pass" entry in apiurl's section, then the "Please
enter new credentials." codepath in Osc.postoptparse is not triggered and the
user has to manually "fix" it via "osc config ...".

What about moving the lazy password logic into the credentials manager
itself? It probably knows best whether a lazy password is supported and
how a missing password can be retrieved (and potentially persisted (for
instance, in case of the PlaintextConfigFileCredentialsManager)).
That is, something like the keyutils.osc._PasswordRetriever [1] (of
course, __call__ should be renamed to __str__).


[1] https://github.com/marcus-h/python-keyring-keyutils/blob/master/keyutils/osc.py#L35

mlschroe:
Ok, I updated the pull request:

- The lazy password class is now in credentials.py and defines more duner menthods. I.e. + and len() works
- lazy passwords will not be used for the Plaintext/Obfuscated credential manager

A missing password in .oscrc is handled somewhat weird, by the way:
```/home/osc/osc-wrapper.py up
No password found in section http://buildservice:5352
Please enter new credentials.
Username: mlschroe
Password: 

NUM NAME              DESCRIPTION
1   Transient         Do not store the password and always ask for it [secure, in-memory]
2   Obfuscated config Store the password in obfuscated form in the osc config file [insecure, persistent]
3   Config            Store the password in plain text in the osc config file [insecure, persistent]
Select credentials manager [default=1]: 2
Password:
```
So it first asks for the password, then it asks for the manager, and then it asks again for the password. I don't think this is caused by my changes.

marcus-h:
On 2022-04-07 06:26:26 -0700, Michael Schroeder wrote:
> Ok, I updated the pull request:
>
> - The lazy password class is now in credentials.py and defines more duner menthods. I.e. + and len() works
> - lazy passwords will not be used for the Plaintext/Obfuscated credential manager
>
Cool! Thanks! (See below.)

> A missing password in .oscrc is handled somewhat weird, by the way:
> ```/home/osc/osc-wrapper.py up
> No password found in section http://buildservice:5352
> Please enter new credentials.
> Username: mlschroe
> Password:
>
> NUM NAME              DESCRIPTION
> 1   Transient         Do not store the password and always ask for it [secure, in-memory]
> 2   Obfuscated config Store the password in obfuscated form in the osc config file [insecure, persistent]
> 3   Config            Store the password in plain text in the osc config file [insecure, persistent]
> Select credentials manager [default=1]: 2
> Password:
> ```
> So it first asks for the password, then it asks for the manager, and then it asks again for the password. I don't think this is caused by my changes.
>
Hmm if the 2 would be 1, it would make sense from a code's POV (after
setting the credentials manager, the config is read again). Are you
sure there is no copy and paste "error"?

> index 8dceb3f6..558d0ad2 100644
> --- a/osc/credentials.py
> +++ b/osc/credentials.py
> @@ -32,6 +32,30 @@ except BaseException as e:
>  from . import conf
>  from . import oscerr
>
> +class LazyPassword(object):

Maybe we should call it _LazyPassword for now (to indicate that it is a
private API and we can change it later without worrying about backward
compatibility). (Inheriting from "object" is not needed anymore in python3.)

> +    def __init__(self, pwfunc):
> +        self.pwfunc = pwfunc
> +        self.password = None

Hmm actually, the attributes should be prefixed with an underscore, too.

> +
> +    def __str__(self):
> +        if self.password is None:
> +            self.password = self.pwfunc()
> +            if self.password is None:
> +                raise RuntimeError('No password given')

Hmm we never handle a RuntimeError. Using something like

raise OscIOError(None, 'Unable to retrieve credentials')
or
raise OscIOError(None, 'Unable to retrieve password')

might be the better choice.


<SNIP>


> @@ -64,14 +88,12 @@ class AbstractCredentialsManager(object):
>      def create(cls, cp, options):
>          return cls(cp, options)
>
> -    def get_password(self, url, user, defer=True):
> -        # If defer is True a callable can be returned
> -        # and the password is retrieved if the callable
> -        # is called. Implementations are free to ignore
> -        # defer parameter and can directly return the password.
> -        # If defer is False the password is directly returned.
> +    def get_password(self, url, user):
>          raise NotImplementedError()
>
> +    def get_password_lazy(self, url, user):

Arghs. IMHO, this API looks strange. Can't we simply keep the existing
"defer" parameter in get_password? If you want to have a default
implementation, we could just introduce a new subclass

class AbstractLazyPasswordCredentialsManager(AbstractCredentialsManager):
    def _get_password(self, url, user):
        raise NotImplementedError()

    def get_password(self, url, user, defer=True):
        if defer:
            return _LazyPassword(lambda: self._get_password(url, user))
        return self._get_password(url, user)


and let the appropriate classes inherit from it (and rename their
get_password(...) method to _get_password(...)).

The only "issue" left is regarding _potential_ 3rd party credentials
managers that return a callable and rely on the APIHostOptionsEntry
hack... Maybe we could keep the hack for "grace" period like
2-3 releases/months (with an additional deprecation warning) and
remove it afterwards?

Advantage: we do not break any APIs.

Does this make sense/what do you think?

marcus-h:
On 2022-04-07 15:55:58 +0000, Marcus Hüwe wrote:
> On 2022-04-07 06:26:26 -0700, Michael Schroeder wrote:
> > A missing password in .oscrc is handled somewhat weird, by the way:
> > ```/home/osc/osc-wrapper.py up
> > No password found in section http://buildservice:5352
> > Please enter new credentials.
> > Username: mlschroe
> > Password:
> >
> > NUM NAME              DESCRIPTION
> > 1   Transient         Do not store the password and always ask for it [secure, in-memory]
> > 2   Obfuscated config Store the password in obfuscated form in the osc config file [insecure, persistent]
> > 3   Config            Store the password in plain text in the osc config file [insecure, persistent]
> > Select credentials manager [default=1]: 2
> > Password:
> > ```
> > So it first asks for the password, then it asks for the manager, and then it asks again for the password. I don't think this is caused by my changes.
> >
> Hmm if the 2 would be 1, it would make sense from a code's POV (after
> setting the credentials manager, the config is read again). Are you
> sure there is no copy and paste "error"?
>
Hmm or is possible that you configured the transient credentials manager
for your default apiurl (which is not http://buildservice:5352)?

mlschroe:
Updated again.

- about that "asks for password again": please disregard. It was caused because I mistyped the password
- underscore prefixing: done
- it now uses OscIOError
- added back the "defer" argument. Unfortunately it defaults to "True" for compatibility. I didn't create a new class, I just changed the base class
- added compatibility code to deal with get_password returning a callable

marcus-h:
On 2022-04-08 08:01:22 +0000, Michael Schroeder wrote:
> Updated again.
>
> - about that "asks for password again": please disregard. It was caused because I mistyped the password
> - underscore prefixing: done
> - it now uses OscIOError
> - added back the "defer" argument. Unfortunately it defaults to "True" for compatibility. I didn't create a new class, I just changed the base class
> - added compatibility code to deal with get_password returning a callable
>
I just had a very quick glance at the callable stuff: If you have a bunch
of apiurls that use a credentials manager, which returns a callable, osc
will ask for a password for each them during startup (even if they are not
used).
I'll have a look at the rest later.

mlschroe:
I've reverted that part (as silly as it is).

marcus-h:
On 2022-04-08 01:16:22 -0700, Marcus Hüwe wrote:
> On 2022-04-08 08:01:22 +0000, Michael Schroeder wrote:
> > Updated again.
> >
> > - about that "asks for password again": please disregard. It was caused because I mistyped the password
> > - underscore prefixing: done
> > - it now uses OscIOError
> > - added back the "defer" argument. Unfortunately it defaults to "True" for compatibility. I didn't create a new class, I just changed the base class
> > - added compatibility code to deal with get_password returning a callable
> >
Commit 7dd43f34d3f9892e2fa44acfa3c7488fe752175b looks good to me!
It would be great if you could adjust the commit message (we still
have the stupid defer hack:/ - maybe state it will be removed soon).

If you are still in the mood, you could add something like this

diff --git a/osc/conf.py b/osc/conf.py
index 6c679dac..8cd6ceb3 100644
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -885,6 +885,8 @@ class APIHostOptionsEntry(dict):
     def __getitem__(self, key, *args, **kwargs):
         value = super(self.__class__, self).__getitem__(key, *args, **kwargs)
         if key == 'pass' and callable(value):
+            print('Warning: use of a deprecated credentials manager API.',
+                  file=sys.stderr)
             value = value()
         return value


> index 07c6c967..fb34283c 100644
> --- a/osc/conf.py
> +++ b/osc/conf.py
> @@ -979,7 +979,7 @@ def get_config(override_conffile=None,
>          user = _extract_user_compat(cp, url, creds_mgr)
>          if user is None:
>              raise oscerr.ConfigMissingCredentialsError('No user found in section %s' % url, conffile, url)
> -        password = creds_mgr.get_password(url, user)
> +        password = creds_mgr.get_password(url, user, defer=True)

Minor minor minor: the defer argument can be omitted

>          if password is None:
>              raise oscerr.ConfigMissingCredentialsError('No password found in section %s' % url, conffile, url)
>
>

mlschroe:
Ok, print statement added and force-pushed.

I did that explicit `defer=True` on purpose, because:
- I think the default is exactly wrong, it should be defer=False
- the way forward is to remove the default when it's ok to break the API

marcus-h:
Hmpf github ignored my mail again...
```
On 2022-04-11 08:53:53 +0000, Michael Schroeder wrote:
> Ok, print statement added and force-pushed.
>
> I did that explicit `defer=True` on purpose, because:
> - I think the default is exactly wrong, it should be defer=False
> - the way forward is to remove the default when it's ok to break the API
>
Ok. Merged. Thanks a lot!
```

PR #1023, opened 2022-04-04
parseRevisionOption(): Raise an exception on invalid revisons

dmach:
Enhances #622

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -7048,7 +7048,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             project = args[0]
             package = args[1]
 
-        rev, rev_upper = parseRevisionOption(opts.revision, allow_md5=False)
+        rev, rev_upper = parseRevisionOption(opts.revision)
         if rev and not checkRevision(project, package, rev, apiurl, opts.meta):
             print('Revision \'%s\' does not exist' % rev, file=sys.stderr)
             sys.exit(1)
--- a/osc/core.py
+++ b/osc/core.py
@@ -6707,12 +6707,15 @@ def parseRevisionOption(string, allow_md5=True):
     revisions = [None, None]
     if string:
         parts = string.split(':')
-        for i, revision in enumerate(parts[0:2], 0):
+
+        if len(parts) > 2:
+            raise oscerr.OscInvalidRevision(string)
+
+        for i, revision in enumerate(parts, 0):
             if revision.isdigit() or (allow_md5 and revision.isalnum() and len(revision) == 32):
                 revisions[i] = revision
             elif revision != '' and revision != 'latest':
-                print('your revision \'%s\' will be ignored' % string, file=sys.stderr)
-                return None, None
+                raise oscerr.OscInvalidRevision(string)
 
     return tuple(revisions)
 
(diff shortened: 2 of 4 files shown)


lethliel:
LGTM
I would like to merge this.
@marcus-h any objections?

marcus-h:
Hmpf github ignored my mail again...
```
On 2022-04-04 01:57:21 -0700, Daniel Mach wrote:
> index c67f906d..f3646b79 100644
> --- a/osc/core.py
> +++ b/osc/core.py
> @@ -6707,12 +6707,15 @@ def parseRevisionOption(string, allow_md5=True):

Hmm while we are at it: do we really need the allow_md5 parameter at all?
It was introduced in commit 6a3976dc22ac6e98c4598b8f9b593250c8e8a9a9
("osc/core: parseRevisionOption(): cleanup and make logic consistent."),
which just breaks "osc log -r <md5> ...". Can't we simply remove it again?


>      revisions = [None, None]
>      if string:
>          parts = string.split(':')
> -        for i, revision in enumerate(parts[0:2], 0):
> +
> +        if len(parts) > 2:
> +            raise oscerr.OscInvalidRevision(string)
> +
> +        for i, revision in enumerate(parts, 0):
>              if revision.isdigit() or (allow_md5 and revision.isalnum() and len(revision) == 32):

We could be a bit more strict here (only a-f instead of alnum) but the
current code is fine IMHO.

>                  revisions[i] = revision
>              elif revision != '' and revision != 'latest':
> -                print('your revision \'%s\' will be ignored' % string, file=sys.stderr)
> -                return None, None
> +                raise oscerr.OscInvalidRevision(string)
> 
>      return tuple(revisions)
> 

<SNIP>

> diff --git a/tests/test_core.py b/tests/test_core.py
> new file mode 100644
> index 00000000..4715a839
> --- /dev/null
> +++ b/tests/test_core.py
> @@ -0,0 +1,52 @@
> +from osc.core import parseRevisionOption
> +from osc.oscerr import OscInvalidRevision
> +
> +import unittest
> +
> +
> +class TestParseRevisionOption(unittest.TestCase):
> +    def test_empty(self):
> +        expected = None, None
> +        actual = parseRevisionOption("")
> +        self.assertEqual(expected, actual)
> +
> +    def test_colon(self):
> +        expected = None, None
> +        actual = parseRevisionOption(":")
> +        # your revision ':' will be ignored
> +        self.assertEqual(expected, actual)
> +
> +    def test_invalid_multiple_colons(self):
> +        self.assertRaises(OscInvalidRevision, parseRevisionOption, ":::::")
> +
> +    def test_one_number(self):
> +        expected = ("1", None)
> +        actual = tuple(parseRevisionOption("1"))

Hmm tuple(...) can be omitted.

> +        self.assertEqual(expected, actual)
> +
> +    def test_two_numbers(self):
> +        expected = ("1", "2")
> +        actual = tuple(parseRevisionOption("1:2"))

Hmm tuple(...) can be omitted.


Otherwise, LGTM (except that it "breaks" the API but the old one is
really nuts...)
```

dmach:
@marcus-h applied fixes you've suggested (except the `isalnum()` part, but I can also throw that in if you consider it really important).
md5 revisions for `osc log` are enabled now; not sure if there's a reason why we shouldn't enable them.

PR #1031, opened 2022-04-21
Restore fetching of metadata files in most cases

dirkmueller:
using osc getbinaries <repository> <arch> used to fetch the metadata
files as well, but 08a70952c05f745f1c56ee04740c589f0f89585b changed
that behavior even for cases where there would not be any collission.

Restore fetching of metadata files by expanding per architecture
or by fetching just as before if the architecture is given.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -7516,7 +7516,6 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         if architecture is None:
             arches = [i.arch for i in repos if repository == i.name]
 
-
         if package is None:
             package_specified = False
             package = meta_get_packagelist(apiurl, project, deleted=0)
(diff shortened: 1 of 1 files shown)


dmach:
@dirkmueller I have another fix (https://github.com/openSUSE/osc/pull/1031) waiting for review. Would that work for you?
To be honest, I don't like the behavior your patch introduces.
If something relies on the single arch behavior and 2nd arch is added, things break.
It would be better if osc behaved consistently in all cases.

dirkmueller:
@dmach  #1015 lgtm as well. I don't think storing the build results in subdirectories would have been a problem, but this patch appears to be a slightly more compatible approach, so I don't mind. 

I rebased this on top of yours to include the stylistic cleanups that were intermixed.

dmach:
Looks good to me, merging.

PR #1032, opened 2022-04-27
Implement the Signature authentication scheme

mlschroe:


mlschroe:
Let the discussions begin ;-)

dmach:
I'm using the following configuration:
```
[https://apitest.suse.de]
aliases=apitest
user=dmach
pass=
#sshkey=~/.ssh/id_rsa
```

If `pass=` is not specified, osc wants to enter new credentials (username, password, select credentials manager).
That also makes me think, that this probably should be implemented as another credentials manager.

When sshkey is specified, it works fine.
When it's empty, auth fails because it uses `~/.ssh/id_ed25519` that is not known to the server.
Also running `ssh-add ~/.ssh/id_rsa` doesn't help and I'm unable to authenticate.

dmach:
I thought I'll update this PR with a fixup and an additional commit, but I failed (probably because it comes from master or doesn't have the 'allow edit' flag set).
@mlschroe please check my commits in https://github.com/dmach/osc/tree/pr/1032 and consider adding them to this PR.

mlschroe:
Hi Daniel, I added most of your changes to this PR.

osc/conf.py
```diff
@@ -45,6 +45,8 @@
 import ssl
 import warnings
```
dmach, line 46: The `warnings` module is no longer used. Can you remove it?

osc/conf.py
```diff
@@ -62,10 +65,11 @@
     from urlparse import urlsplit
     from urllib2 import URLError, HTTPBasicAuthHandler, HTTPCookieProcessor, HTTPPasswordMgrWithDefaultRealm, ProxyHandler, AbstractBasicAuthHandler
```
dmach, line 66: `AbstractBasicAuthHandler` is no longer used, can you remove it?

dmach, line 66: `AbstractBasicAuthHandler` is no longer used, can you remove it?

dmach (changes requested):
I've highlighted 2 unused imports that can be removed.
It would be good to remove them.
The code looks good to me otherwise.

osc/conf.py
```diff
@@ -519,10 +533,139 @@ def _rewind_request(self, req):
...
+                return stdout.splitlines()
+            else:
+                return []
+
+        def is_ssh_private_keyfile(self, keyfile_path):
+            with open(keyfile_path, "r") as f:
+                line = f.readline(100).strip()
+                if line == "-----BEGIN OPENSSH PRIVATE KEY-----":
+                    return 1
+            return 0
```
dmach, line 576: This is ok, but ideally you should be returning `True` and `False`.

osc/conf.py
```diff
@@ -519,10 +533,139 @@ def _rewind_request(self, req):
...
+        def http_error_401(self, req, fp, code, msg, headers):
+            authreq = headers.get('www-authenticate', None)
+            if authreq:
+                scheme = authreq.split()[0]
+                if scheme.lower() == 'signature':
+                    return self.retry_http_signature_auth(req, authreq)
+                raise ValueError("OscHTTPSignatureAuthHandler does not support"
+                                 " the following scheme: '%s'" % scheme)
+
+        def sshkey_known(self):
```
dmach, line 667: Add the following lines otherwise the keys do not get guessed at all.
At least it didn't work for me with the keys imported into ssh-agent.
```
            if not self.sshkey:
                self.sshkey = self.guess_keyfile()
```

osc/conf.py
```diff
@@ -519,10 +533,139 @@ def _rewind_request(self, req):
...
+        def list_ssh_agent_keys(self):
+            cmd = ['ssh-add', '-l']
+            proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            stdout, _ = proc.communicate()
+            if proc.returncode == 0 and stdout.strip():
+                return stdout.splitlines()
+            else:
+                return []
+
+        def is_ssh_private_keyfile(self, keyfile_path):
```
dmach, line 571: Test if `keyfile_path` is a file first:
```
            if not os.path.isfile(keyfile_path):
                return False
```

dmach (changes requested):

osc/conf.py
mlschroe: No, we can't do that. We can only use signature auth if the user has stored his pubkey in the authentication service (univention/privacyidea). As we do not know this, we fall back to basic auth if it is available.
If only signature auth is provided, we will not call sshkey_known().

dmach: `self.ssh_key` is set in `OscHTTPSignatureAuthHandler.__init__()` regardless if pubkey is available on the server or not. If `self.ssh_key` is not set explicitly in the config file, it should be guessed. Maybe I've proposed guessing in a wrong place. How about to move this to `__init__()`? For example:

```
        def __init__(self, user, sshkey):
            super(self.__class__, self).__init__()
            self.user = user
            self.sshkey = sshkey or self.guess_keyfile()
```

mlschroe: No, we must not guess if the server offers both signature and basic authentication.

dmach:
This is good to go. Merging.

osc/conf.py
```diff
@@ -519,10 +533,141 @@ def _rewind_request(self, req):
...
+            signature = decode_it(stdout)
+            match = re.match(r"\A-----BEGIN SSH SIGNATURE-----\n(.*)\n-----END SSH SIGNATURE-----", signature, re.S)
+            if not match:
+                raise oscerr.OscIOError(None, 'could not extract ssh signature')
+            return base64.b64decode(match.group(1))
+
+        def get_authorization(self, req, chal):
+            realm = chal.get('realm', '')
+            now = int(time.time())
+            sigdata = "(created): %d" % now
```
crazyscientist, line 640: Hi there, I'm currently implementing this logic for [OSC-Tiny](https://github.com/crazyscientist/osc-tiny) and noticed that the substring `(created)` would have been available from the response header, too.

If you don't mind my asking, any particular reason for hard-coding this substring?

crazyscientist: Thanks for sharing the background info.

>  the client sends what it actually used (which can be a superset).

After looking into the specs of [HTTP Message Signatures](https://tools.ietf.org/id/draft-cavage-http-signatures-12.html), my understanding would be similar, but the server keeps responding with 401, if I try to use additional data in the signature string.

But this would be a(n academic) topic for the server-side implementation.

Issue #1026, opened 2022-04-08
Crash in osc linkdiff


marxin:
It crashes when I do the following steps:

```
$ osc co Base:System/autogen
$ cd $_
$ touch a.patch
$ osc add a.patch
$ rm a.patch
$ osc linkdiff
diff working copy against last committed version

Traceback (most recent call last):
  File "/usr/bin/osc", line 45, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.8/site-packages/osc/babysitter.py", line 67, in run
    return prg.main(argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 347, in main
    return self.cmd(args)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 370, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 504, in onecmd
    return self._dispatch_cmd(handler, argv)
  File "/usr/lib/python3.8/site-packages/osc/cmdln.py", line 1235, in _dispatch_cmd
    return handler(argv[0], opts, *args)
  File "/usr/lib/python3.8/site-packages/osc/commandline.py", line 4026, in do_diff
    for i in pac.get_diff(rev1):
  File "/usr/lib/python3.8/site-packages/osc/core.py", line 2096, in get_diff
    yield diff_add_delete(f, True, revision)
  File "/usr/lib/python3.8/site-packages/osc/core.py", line 2017, in diff_add_delete
    if binary_file(fname):
  File "/usr/lib/python3.8/site-packages/osc/core.py", line 4800, in binary_file
    return binary(open(fn, 'rb').read(4096))
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/Base:System/autogen/a.patch'
```

dmach:
PR https://github.com/openSUSE/osc/pull/1033 should fix the problem by raising a proper exception on a missing file.

dmach:
The linked PR has been merged.

PR #1033, opened 2022-04-27
linkdiff: Raise an exception when an added file is missing

dmach:
Fixes:
- https://github.com/openSUSE/osc/issues/1026


lethliel:
LGTM and thanks for the fix :-)

PR #1034, opened 2022-04-28
Fix check for empty details

andreas-schwab:


osc/core.py
```diff
@@ -5868,7 +5868,7 @@ def get_results(apiurl, project, package, verbose=False, printJoin='', *args, **
                 if len(l) != 2 or l[1] not in multibuild_packages:
                     continue
             res['status'] = res['code']
-            if verbose and res['details'] != '':
+            if verbose and res['details'] is not None:
```
dmach, line 5871: Wouldn't be better to do this?
```
if verbose and res['details']:
```
That would cover the existing empty string the comparison to `None` you're suggesting.

Issue #875, opened 2020-11-28
`osc help` requires a login


abitrolly:
This issues is in line with #812, which is about `osc co` requiring login.

```bash
# osc help

Your user account / password are not configured yet.
You will be asked for them below, and they will be stored in
/root/.config/osc/oscrc for future use.

Creating osc configuration file /root/.config/osc/oscrc ...
Username: ^Cinterrupted!
```

`osc --help` works without login.

`osc co -h` requires loging too.

marcus-h:
On 2020-11-28 04:31:37 -0800, Anatoli Babenia wrote:
> This issues is in line with #812, which is about `osc co` requiring login.
>
> ```bash
> # osc help
>
> Your user account / password are not configured yet.
> You will be asked for them below, and they will be stored in
> /root/.config/osc/oscrc for future use.
>
> Creating osc configuration file /root/.config/osc/oscrc ...
> Username: ^Cinterrupted!
> ```
>
> `osc --help` works without login.
>
> `osc co -h` requires loging too.
>
Hmm IMHO there is not much benefit from changing the current behavior.
As of now, most osc commands require credentials.

abitrolly:
Even CLI for proprietary services, who presumably want to track user behaviour, doesn't force users to register to see help.

abitrolly:
@dmach so what was the consensus?

dmach:
@abitrolly sorry for not adding more details when closing, it's fixed and merged already: #1035

PR #1035, opened 2022-05-05
Never require login in the help command

dmach:
Fixes
 - #875


lethliel:
LGTM and thanks for the fix

Issue #425, opened 2018-05-28
Wrong version


mcepl:
When running with setuptools, ``setup.py`` indicates wrong version of the package:
```
/usr/lib/python2.7/site-packages/setuptools/dist.py:407:
UserWarning: The version specified ('0.162.1git') is an invalid version,
this may not work as expected with newer versions of setuptools, pip, and PyPI.
Please see PEP 440 for more details.
```

dmach:
Fixed already in #1037

PR #1037, opened 2022-05-05
Make osc git version compatible with pip to mute the PEP 440 warning

dmach:
Fixes:
   - #425

Please note that it fixes the warning, but the versions are not used correctly.
```
from packaging import version
assert version.parse("0.177.0") < version.parse("0.178.0")
assert version.parse("0.177.0+git") < version.parse("0.178.0")
assert version.parse("0.178.0+git") < version.parse("0.178.0")  # this fails
```

We should either use the released version + `+git` or avoid using the `+git` suffix completely.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+osc/util/git_version.py export-subst
+
--- a/osc/core.py
+++ b/osc/core.py
@@ -5,7 +5,10 @@
 
 from __future__ import print_function
 
-__version__ = '0.179'
+
+from .util import git_version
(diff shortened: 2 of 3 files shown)


lethliel:
@marcus-h Any opinions on this one?

dmach:
The git version is set automatically on `git archive`. Calling `pip install -e .` also works fine, in this case generated version `0.179.0+1.git.624ecb04`. This means that commits opening development (such as `open 0.178 development`) are no longer necessary.

PR #1038, opened 2022-05-06
Show the md5s that are failing to validate

Firstyear:
Rather than just saying an md5 mismatch occured, show the exact values of the md5s that are failing. This generally indicates a mirror-synchronisation issue or something else that has gone wrong


dmach:
Looks good to me. Let's ignore the PEP8 error, because it wasn't created in your patch. Merging.

PR #1041, opened 2022-05-10
Add support for project separator

AdamMajer:
The motivation is to have a configurable project separator and not just use hardcoded colon. The main obstacle in using colons as project separator is the inability of including colon in environment variables like PATH or LD_LIBRARY_PATH. This then breaks in not so subtle and yet wonderfully mysterious ways.

An example is to do `npm install` and then doing `npm run` and getting "command not found" because the cwd cannot be added to PATH. Another is trying to run a unit test that fixes a segfault in a program, yet the ctest keeps failing because the patch contains : and thus is unusable in the LD_LIBRARY_PATH that you expect to run...

As a workaround, I've now taken advantage of UTF-8 and can have checkouts like,

     home🚀adamm🚀valgrind/valgrind

by ading this to the config file (.oscrc),

    project_separator = 🚀

(FYI: I'm using a rocket emoji to lift my productivity in this Web 3.0 world)

Regarding patches, the small one allows for checkouts. The big one is to make things work the other way, allowing you to copy paste the altered checkout.

I've marked this as WIP because I haven't yet checked if I've managed to find all the places where project is parsed as osc argument in the big patch.

Opinions and feedback is welcome.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -4704,6 +4704,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             prj_dir = opts.output_dir if opts.output_dir else project
             if not opts.output_dir and conf.config['checkout_no_colon']:
                 prj_dir = prj_dir.replace(':', '/')
+            else:
+                prj_dir = prj_dir.replace(':', conf.config['project_separator'])
             if os.path.exists(prj_dir):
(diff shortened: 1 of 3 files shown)


dmach:
I've tested this briefly and it doesn't seem to break the default behavior and the separator worked nicely for me.
Merging.

PR #1043, opened 2022-05-20
Fix quoting of download urls

dirkmueller:
this helps with building for example in devel:languages:c_c++


dmach:
I have discussed with @dirkmueller and I'm going to merge this.
Escaping in the original code was broken and there's a case with `c_c++` that needs to be fixed.

PR #1045, opened 2022-05-28
Add python-rpmextras_require for rpm signature verification

Thaodan:
The dependency was added in 8fb9669ae414f06e61dbe112704540e6b8ce55ca but not specified.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,6 +102,9 @@ setuptools.setup(
     packages=['osc', 'osc.util'],
     data_files=data_files,
     install_requires=['cryptography', 'urllib3'],
+    extras_require={
+       'RPM signature verification': ['rpm'],
+    },
     entry_points={
       'console_scripts': [
           'osc=osc.babysitter:main'


dmach:
@Thaodan how should we use this? I couldn't google any steps that would work for me.
Please note that rpm exists on pypi but is empty: https://pypi.org/project/rpm/

Thaodan:
What do you mean by using?

dmach:
> What do you mean by using?

Why we're adding the `extras_require`? What is the purpose?
Can we somehow call pip / setup.py to install rpm (which is empty on pypi)?

Thaodan:
> > What do you mean by using?
> 
> Why we're adding the `extras_require`? What is the purpose? Can we somehow call pip / setup.py to install rpm (which is empty on pypi)?

I assume the rpm package is empty on pypy because its usually build together with rpm

The purpose is that pip checks if rpm-python is installed and it gets into the required symbols up when building an osc rpm.

PR #1048, opened 2022-06-01
ssh key authentification fixes

adrianschroeter:
- do not crash when having binary files in ~/.ssh
- support also RSA key setups


dmach:
Thanks for catching and fixing this. Merging.

PR #1049, opened 2022-06-02
RFC: ssh: recognize gpg keys (yubikey usage)

mwilck:
When using ssh keys from gpg, there are no private key files on
disk. The public keys are available from "ssh-add -L". Conveniently,
users store the public keys in some ".pub" file under ~/.ssh
(see e.g. https://serverfault.com/questions/906871/force-the-use-of-a-gpg-key-as-an-ssh-key-for-a-given-server;
this is also necessary to use IdentityFile= in ssh itself).

Thus public key files can't be ignored any more in list_ssh_dir_keys().
"ssh-keygen -Y sign" works nicely with a public key file if the agent
has access to the private key.

NOTE: OBS authenticfication **did not work** for me with this method yet (error 401 Unauthorized). But that seems to be a problem on the server side; I am attaching [rsa-verify.zip](https://github.com/openSUSE/osc/files/8827558/rsa-verify.zip)  the data from a failed attempt and a small script that can be used to verify the signature successfully:

```
$ ssh-keygen -Y verify -n 'Use your developer account' -I mwilck@suse.com -s signature.dat -f allowed <challenge.dat
Good "Use your developer account" signature for mwilck@suse.com with RSA key SHA256:JdVGip75WbGRI6Qmp1+wOxkThhXUbCf6OVfUSheZp4k
```

Here `challenge.dat` was an original challenge from apitest.suse.de, and signature.dat the actual signature generated by osc. The signature is fine;  for some reason, the verification failed on the server side.

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -588,17 +588,20 @@ def _build_opener(apiurl):
             sshdir = os.path.expanduser('~/.ssh')
             keys_in_home_ssh = {}
             for keyfile in os.listdir(sshdir):
-                if keyfile.endswith(".pub"):
-                    continue
                 keyfile_path = os.path.join(sshdir, keyfile)
-                if not self.is_ssh_private_keyfile(keyfile_path):
+                if (not os.path.isfile(keyfile_path) or
+                        keyfile.startswith("authorized_keys") or
+                        keyfile == "config" or
+                        keyfile.startswith("agent-") or
+                        keyfile.startswith("known_hosts")):
                     continue
+                is_private = self.is_ssh_private_keyfile(keyfile_path)
                 cmd = ["ssh-keygen", "-lf", keyfile_path]
                 proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
                 stdout, _ = proc.communicate()
                 if proc.returncode == 0:
                     fingerprint = stdout.strip()
-                    if fingerprint:
+                    if fingerprint and (fingerprint not in keys_in_home_ssh or is_private):
                         keys_in_home_ssh[fingerprint] = keyfile_path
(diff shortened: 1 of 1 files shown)


osc/conf.py
```diff
@@ -588,17 +588,20 @@ def list_ssh_dir_keys(self):
...
             for keyfile in os.listdir(sshdir):
-                if keyfile.endswith(".pub"):
-                    continue
                 keyfile_path = os.path.join(sshdir, keyfile)
-                if not self.is_ssh_private_keyfile(keyfile_path):
+                if (not os.path.isfile(keyfile_path) or
+                        keyfile.startswith("authorized_keys") or
+                        keyfile == "config" or
+                        keyfile.startswith("agent-") or
+                        keyfile.startswith("known_hosts")):
```
m-ildefons, line 596: I'm not sure this logic is sound. In cases where a user has multiple key pairs and only one of them is supposed to be used for `osc`, this may choke on other private keys, no?
It should probably be `if (not self.is_ssh_private_keyfile) and (not self.is_ssh_public_keyfile)`.

mwilck: Right, the logic doesn't work with multiple keys. But that was the also the case previously, unless I'm mistaken (the code would have picked up every private key file). If you have multiple keys, you need to put `sshkey=...` in your `.oscrc`.

> It should probably be if (not self.is_ssh_private_keyfile) and (not self.is_ssh_public_keyfile).

I was too lazy to code up `is_ssh_public_keyfile()` :grin:, I believe that's non-trivial. I was relying on the fact that `ssh-keygen -lf` would fail for anything that's not either a private or a public key file.

dmach:
Based on discussion with @mwilck, I created #1074 with this PR reworked.

PR #1051, opened 2022-06-07
Warn when trying to commit a prj/pac managed in scm

dmach:


osc/commandline.py
```diff
@@ -5015,10 +5015,16 @@ def _commit(self, subcmd, opts, args):
...
             skip_local_service_run = True
-        arg_list = args[:]
-        for arg in arg_list:
+
+        for arg in args.copy():
             if conf.config['do_package_tracking'] and is_project_dir(arg):
                 prj = Project(arg)
+
+                if prj.scm_url:
+                    print("WARNING: Skipping project '{}' because it is managed in scm (git)".format(prj.name))
```
adrianschroeter, line 5024: please output also the scm_url, so that the user has a chance to follow up

osc/commandline.py
```diff
@@ -5037,6 +5043,12 @@ def _commit(self, subcmd, opts, args):
 
         pacs, no_pacs = findpacs(args, fatal=False)
 
+        for pac in pacs.copy():
+            if pac.scm_url:
+                print("WARNING: Skipping package '{}' because it is managed in scm (git)".format(pac.name))
```
adrianschroeter, line 5048: same here

adrianschroeter (changes requested):

osc/commandline.py
dmach: Right, display some context. Done.

PR #1067, opened 2022-07-04
Fix checkout of '_project' package

dmach:


dmach:
There's a better fix in #1070

PR #1068, opened 2022-07-04
Fix emoji in messages

dmach:


AdamMajer:
FWIW, this code is everywhere to check that there is no unicode in the text and then to escape it.

I think the question we should ask here is whether OBS UTF-8 safe?

AdamMajer:
@lethliel

coolo:
OBS is utf-8. The commit reverted came for #816 and is *very* doubtful indeed. I'm all for reverting this.

coolo (approved):

dmach:
After doing a detailed review of the revert with @eduardoj and the feedback in this PR, I have decided to merge this.

PR #1069, opened 2022-07-07
fix product build rpm caching

adrianschroeter:
* src/noarch rpm packaages needs to be stored in scheduler architecture to avoid
  conflicts of the multiple versions
* avoid removal of every downloaded file


dmach:
Looks good to me, merging. Thanks!

PR #1077, opened 2022-07-24
Fix failure to create config in current dir

abitrolly:
Otherwise `osc` fails with the trace.

```
$ osc -d -v -c fff dists
...
  File "/tmp/osh/osc/osc/conf.py", line 962, in write_initial_config
    write_config(conffile, cp)
  File "/tmp/osh/osc/osc/conf.py", line 832, in write_config
    os.makedirs(os.path.dirname(fname), mode=0o700)
  File "/usr/lib64/python3.10/os.py", line 225, in makedirs
    mkdir(name, mode)
FileNotFoundError: [Errno 2] No such file or directory: '' 
```


osc/conf.py
```diff
@@ -827,12 +827,9 @@ def write_config(fname, cp):
...
     # create directories to the config file (if they don't exist already)
-    if not os.path.exists(os.path.dirname(fname)):
-        try:
-            os.makedirs(os.path.dirname(fname), mode=0o700)
-        except OSError as e:
-            if e.errno != errno.EEXIST:
-                raise
+    fdir = os.path.dirname(fname)
+    if fdir and not os.path.exists(fdir):
+        os.makedirs(fdir, mode=0o700)
```
dmach, line 832: Why didn't you keep the `try/except` around `os.makedirs`?
osc module is not considered thread-safe, but why to make it worse?
(`os.makedirs` fails when a parallel thread creates the directory in the meantime)

dmach (changes requested):

osc/conf.py
abitrolly: `os.path.dirname()` already checks that repository doesn't exist, so I see no way to fail here on `EEXIST`. I also do not see that `osc` uses any of Python modules for multithreading https://docs.python.org/3/library/concurrency.html and without it I believe Python is a single limited GIL process.

dmach: This can happen if you run multiple instances of osc or write a threaded application using osc as a library.
Could you please keep the `try/except`?

abitrolly: Got `try/except` back. Why not to use proper locking? https://github.com/tox-dev/py-filelock

dmach: I'm sorry for being such a nit-picker, but couldn't you merge your commits into one?
Removing `try/except` in one commit and bringing it back in another in the same PR doesn't look good to me.

abitrolly: No problem. Done.

dmach:
Thank you, merging.

PR #1000, opened 2022-02-03
Switch http_request() to urllib3

dmach:
Use connection pools for better performance.
M2crypto is no longer needed, we use urllib3's ssl context.

:exclamation: Please note that this code no longer works with python2.
I'm going to make several more improvements, fix TODOs and properly test it before moving this PR from draft.

Issues to be closed when this is finished:

- https://github.com/openSUSE/osc/issues/976
- https://github.com/openSUSE/osc/issues/179
- https://github.com/openSUSE/osc/issues/373
- https://github.com/openSUSE/osc/issues/427
- https://github.com/openSUSE/osc/issues/455
- https://github.com/openSUSE/osc/pull/773
- https://github.com/openSUSE/osc/issues/774
- https://github.com/openSUSE/osc/pull/506
- https://github.com/openSUSE/osc/pull/428
- https://github.com/openSUSE/osc/pull/542
- https://github.com/openSUSE/osc/issues/179
- https://github.com/openSUSE/osc/pull/1078

PR #1078, opened 2022-07-24
No need to install Python 3 explicitly

abitrolly:
It's pulled as a dependency to python3-* packages.
This should fix test failures on `fedora:rawhide`.

Fedora install step fails right now, because `dnf` outputs this to `stderr`.
```
Complete!
Last metadata expiration check: 0:00:53 ago on Sun Jul 24 17:32:14 2022.
Package python3-3.11.0~b4-1.fc37.x86_64 is already installed.
Package python3-rpm-4.18.0-0.beta1.2.fc37.x86_64 is already installed.
Error: 
(try to add '--skip-broken' to skip uninstallable packages)
 Problem: conflicting requests
  - nothing provides python(abi) = 3.10 needed by python3-m2crypto-0.38.0-2.fc35.x86_64
Error: Process completed with exit code 1.
```
And GitHub Actions treat non-empty `stderr` as error.

### DIFF
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -58,7 +58,7 @@ jobs:
             dnf -y makecache
             dnf -y distro-sync
             dnf -y install git-lfs
-            dnf -y install diffstat diffutils python3 python3-chardet python3-m2crypto python3-pip python3-rpm python3-setuptools
+            dnf -y install diffstat diffutils python3-chardet python3-m2crypto python3-pip python3-rpm python3-setuptools
 
       - name: 'Install packages (Debian/Ubuntu)'
         if: ${{ startsWith(matrix.container, 'debian:') || startsWith(matrix.container, 'ubuntu:')  }}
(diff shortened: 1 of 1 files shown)


dmach:
I believe that the patch doesn't fix the problem because the problem is elsewhere.
python3-m2crypto is build against python 3.10 while there's python 3.11 in Rawhide.

The problem seems to be in m2crypto failing to build with the latest OpenSSL:
https://bugzilla.redhat.com/show_bug.cgi?id=2021972

```
$ rpm -q python3
python3-3.11.0~b4-1.fc37.x86_64

$ dnf install python3-m2crypto
 Problem: conflicting requests
  - nothing provides python(abi) = 3.10 needed by python3-m2crypto-0.38.0-2.fc35.x86_64
(try to add '--skip-broken' to skip uninstallable packages)
```

dmach:
> Or maybe there is no need to use `m2crypto`
Yes, that's the plan. There's a huge pull-request #1000 that implements moving from m2crypto to cryptography.
I'm planning to release osc-0.181.0 in a day or two and then bump version to 1.0 and merge #1000.

abitrolly:
I don't think it is possible to resolve this test failure on `osc` level better than in #1000.

PR #1087, opened 2022-07-28
Color diff

mcepl:
Wouldn’t be beautiful if osc diff was colourful? Yes, I can build some aliases around it using colordiff, but do I have to?

Just POC, obviously the configuration should be done better.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -32,6 +32,14 @@ from urllib.error import HTTPError
 from urllib.request import pathname2url
 from xml.etree import ElementTree as ET
 
+# Poor man's colorama
+GREEN = b'\x1b[32m'
+RED = b'\x1b[31m'
+BLUE = b'\x1b[34m'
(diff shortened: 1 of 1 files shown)


osc/core.py
```diff
@@ -4844,7 +4865,11 @@ def get_source_file_diff(dir, filename, rev, oldfilename = None, olddir = None,
             d[i] += b'\n\\ No newline at end of file'
             if i+1 != len(d):
                 d[i] += b'\n'
-    return d
+
+    if COLORIZE:
```
dirkmueller, line 4869: needs a check for whether it is connected to a terminal or not (isatty)

mcepl: Thank you! Fixed.

mcepl:
Except this breaks `osc ci` and then I have to take care of already set `LESS` environmental variable, and … Too much time for something which can be overcome by one alias. Sorry for bothering you.

PR #1090, opened 2022-07-29
Remove Python 2 code

abitrolly:
This removes function that is no longer needed.

There is also the Python 2 code that catches invalid arguments by parsing text of exception messages. I don't feel like I can touch that safely.

https://github.com/openSUSE/osc/blob/4a67768cbdf3b8b98187b15d94c6f37d2225e7ba/osc/cmdln.py#L93-L100

### DIFF
--- a/osc/cmdln.py
+++ b/osc/cmdln.py
@@ -48,22 +48,8 @@ import time
 from pprint import pprint
 from datetime import datetime
 
-# this is python 2.x style
-def introspect_handler_2(handler):
-    # Extract the introspection bits we need.
-    func = handler.im_func
-    if func.func_defaults:
-        func_defaults = func.func_defaults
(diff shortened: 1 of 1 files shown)


dmach:
I'm going to merge it because it moves the project in the right direction,
but I think the whole `cmdln.py` will have to be replaced with something else.
Please note that it is a 3rd party code that is bundled in osc and is no longer maintained.

abitrolly:
Thanks. I didn't notice that this [cmd](wrapper over https://docs.python.org/3/library/cmd.html
) wrapper is an external project. Although already archived https://github.com/trentm/cmdln

Does it make sense to use the latest version https://github.com/trentm/cmdln/blob/master/lib/cmdln.py ?

PR #863, opened 2020-11-14
Pretty print `osc dists` output

abitrolly:
This table formatting code calculates the widest column before printing. It can be generalized for other tabular data if needed.

### Before

```python
distribution              project                   repository                reponame
------------              -------                   ----------                --------
openSUSE Tumbleweed       openSUSE:Factory          snapshot                  openSUSE_Tumbleweed
openSUSE Leap 15.2        openSUSE:Leap:15.2        standard                  openSUSE_Leap_15.2
openSUSE Leap 15.2 ARM    openSUSE:Leap:15.2:ARM    ports                     openSUSE_Leap_15.2_ARM
openSUSE Leap 15.2 PowerPC openSUSE:Leap:15.2:PowerPC ports                     openSUSE_Leap_15.2_PowerPC
openSUSE Leap 15.1        openSUSE:Leap:15.1        standard                  openSUSE_Leap_15.1
openSUSE Leap 15.1 ARM    openSUSE:Leap:15.1:ARM    ports                     openSUSE_Leap_15.1_ARM
openSUSE Leap 15.1 PowerPC openSUSE:Leap:15.1:PowerPC ports                     openSUSE_Leap_15.1_PowerPC
openSUSE Factory RISCV    openSUSE:Factory:RISCV    standard                  openSUSE_Factory_RISCV
openSUSE Factory ARM      openSUSE:Factory:ARM      standard                  openSUSE_Factory_ARM
openSUSE Factory PowerPC  openSUSE:Factory:PowerPC  standard                  openSUSE_Factory_PowerPC
openSUSE Factory zSystems openSUSE:Factory:zSystems standard                  openSUSE_Factory_zSystems
openSUSE Backports for SLE 15 SP2 openSUSE:Backports:SLE-15-SP2 standard                  SLE_15_SP2_Backports
openSUSE Backports for SLE 15 SP1 openSUSE:Backports:SLE-15-SP1 standard                  SLE_15_SP1_Backports
openSUSE Backports for SLE 15 openSUSE:Backports:SLE-15 standard                  SLE_15_Backports
openSUSE Backports for SLE 12 SP5 openSUSE:Backports:SLE-12-SP5 standard                  SLE_12_SP5_Backports
openSUSE Backports for SLE 12 SP4 openSUSE:Backports:SLE-12-SP4 standard                  SLE_12_SP4_Backports
openSUSE Backports for SLE 12 SP3 openSUSE:Backports:SLE-12-SP3 standard                  SLE_12_SP3_Backports
[...]

PR #1093, opened 2022-07-29
Move out table formatting in `osc dist` into helper function

abitrolly:
This is a cleanup of #863 and refactoring of internals.

* fixed typo in core `get_distributions` method name
* moved out formatting code into `util.helper.format_table` function
* made `core` method return list of dicts instead of list of strings

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -5603,7 +5603,7 @@ def get_repositories(apiurl):
     return r
 
 
-def get_distibutions(apiurl, discon=False):
+def get_distributions(apiurl, discon=False):
     r = []
 
     # FIXME: this is just a naming convention on api.opensuse.org, but not a general valid apparoach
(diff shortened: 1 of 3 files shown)


osc/core.py
```diff
@@ -5603,11 +5603,13 @@ def get_repositories(apiurl):
     return r
 
 
-def get_distibutions(apiurl, discon=False):
-    r = []
+def get_distributions(apiurl, discon=False):
+    """Returns list of dicts with headers
+      'distribution', 'project', 'repository', 'reponame'"""
```
dmach, line 5608: This doesn't apply to the `DISCONTINUED` distributions below.
Could you rewrite the code to return list of dicts too?

dmach (changes requested):

osc/core.py
abitrolly: The option to list discontinued distributions was commented out about 9 years ago.

https://github.com/openSUSE/osc/blame/5d7bd223400b181b061935a871faf6db9355d1d4/osc/commandline.py#L5329

So I propose to just remove this block.

dmach: Agreed.
Could you remove the `discon=False` argument and the whole related block of code then?

abitrolly: Removed and rebased.

dmach (approved):

PR #1095, opened 2022-08-03
Reworded `build` help and added debug prints

abitrolly:
Debug prints are needed for troubleshooting ARCH selection problem in
https://github.com/openSUSE/osc/pull/241#issuecomment-1203463612

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -87,6 +87,11 @@ class Osc(cmdln.Cmdln):
         sys.stderr = safewriter.SafeWriter(sys.stderr)
         sys.stdout = safewriter.SafeWriter(sys.stdout)
 
+    def _debug(self, *args):
+        # if options are not initialized, still allow to use it
+        if not hasattr(self, 'options') or self.options.debug:
+            print(*args, file=sys.stderr)
+
     def get_version(self):
         return get_osc_version()
 
@@ -2460,8 +2465,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
 
                 ## FIXME -B not implemented!
                 if opts.bugowner:
-                    if (self.options.debug):
-                        print('list: option --bugowner ignored: not impl.')
+                    self._debug('list: option --bugowner ignored: not impl.')
 
                 if subcmd == 'review':
                     # FIXME: do the review list for the user and for all groups he belong to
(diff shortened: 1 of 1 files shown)


osc/commandline.py
```diff
@@ -6621,12 +6624,11 @@ def parse_repoarchdescr(self, args, noinit = False, alternative_project = None,
...
 
-        You need to call the command inside a package directory, which should be a
-        buildsystem checkout. (Local modifications are fine.)
+        The command works from a package checkout (local changes are fine).
 
-        The arguments REPOSITORY and ARCH can be taken from the first two columns
-        of the 'osc repos' output. BUILD_DESCR is either a RPM spec file, or a
-        Debian dsc file.
+        You can use `osc repos` to look up REPOSITORY and ARCH arguments. If
+        they are not set, `osc` choses suitable at random. BUILD_DESCR is
```
hramrach, line 6630: I don't think it's good to say the repository is picked at random.

The order of repositories is deterministic, it's just not documented how the repositories are ordered, specifically.

abitrolly: Fair enough. Documented the exact order.

osc/commandline.py
```diff
@@ -6380,6 +6380,11 @@ def parse_repoarchdescr(self, args, noinit = False, alternative_project = None,
             arg_repository, arg_arch, arg_descr = args
 
         arg_arch = arg_arch or osc.build.hostarch
+        if self.options.debug:
+            print("hostarch: ", osc.build.hostarch)
+            print("arg_arch: ", arg_arch)
+            print("arg_repository: ", arg_repository)
+            print("arg_descr: ", arg_descr)
```
dmach, line 6387: How about to print debug messages to stderr?
In my opinion: debug belongs to stderr, verbose increases regular output verbosity and belongs to stdout.

abitrolly: I stole that part from here.

https://github.com/openSUSE/osc/blob/43337afdf5aa258267d407fdaf59feb3d0b15a78/osc/commandline.py#L2461-L2464

I'll make a helper to make it easier.

abitrolly: Added `self_debug` helper, which prints to stderr.

dmach:
Looks good to me, merging.
Thanks!

PR #1097, opened 2022-08-03
Another take to use arch when selecting repo for build target

abitrolly:
This is another attempt to solve problem that `osc build` selects `OpenSuse_Tumbleweed` even if this selection doesn't fit the required architecture.

https://github.com/openSUSE/osc/pull/241#issuecomment-1203673363

* [ ] trace the role of `no_repo`
* [ ] trace the role of `repo_names`
* [ ] check offline logic is not broken (check that offline still
      needed a non-empty repo list before this change)

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -6396,16 +6396,23 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                 # only persist our own repos
                 Repo.tofile(repolistfile, repositories)
 
+        if not repositories:
+            raise oscerr.WrongArgs('no configured repos to build against')
+
+        arch_repos = sorted({r.name for r in repositories if r.arch == arg_arch})
+        if arg_repository not in arch_repos:
+            raise oscerr.WrongArgs('no configured repos for arch %s' % (arg_arch))
+
         no_repo = False
         repo_names = sorted({r.name for r in repositories})
-        if not arg_repository and repositories:
+        if not arg_repository:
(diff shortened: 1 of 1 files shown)


hramrach:
`r.arch == arg_arch` looks fishy, shouldn't there be a list somewhere?

abitrolly:
It is a string comparison of arch that `osc repos` gives.

hramrach:
So it returns the same repo multiple times with different arch

hramrach:
Does not work for me:

localhost:~/home:michals/vertest # osc -d build
hostarch:  armv7l
arg_arch:  armv7l
arg_repository:  None
arg_descr:  None
no configured repos for arch armv7l

abitrolly:
That was a mistake indeed. Fixed that part.

hramrach:
Looks like this works, thanks

osc -d build
hostarch:  armv7l
arg_arch:  armv7l
arg_repository:  None
arg_descr:  None
Building vertest.spec for openSUSE_Leap_15.2_Ports/armv7l

dmach:

abitrolly:
@dmach it looks like this PR improves error messages, but I don't remember what I was trying to do and if it reaches that point. Feel free to close it if it has no value.

dmach:
The current behavior is not perfect, but is tested by using it for years.
If you have an exact scenario where `osc build` misbehaves, please open a bug report and we'll look into that.
I'm closing this one.

PR #1102, opened 2022-08-16
DebQuery: Handle uncompressed Debian packages

raimue:
Debian packages can also be uncompressed in which case the archive contains a control.tar file. Such a package can be created with `dpkg-deb -Znone`.

As an example for reproduction, the [linux-image-unsigned-*.deb](http://security.ubuntu.com/ubuntu/pool/main/l/linux/linux-image-unsigned-5.15.0-46-generic_5.15.0-46.49_amd64.deb) from Ubuntu 22.04 is not compressed.

With master:
```
$ python3 -m osc.util.debquery linux-image-unsigned-5.15.0-46-generic_5.15.0-46.49_amd64.deb 
missing control.tar
```
With this PR branch:
```
$ python3 -m osc.util.debquery linux-image-unsigned-5.15.0-46-generic_5.15.0-46.49_amd64.deb 
b'linux-image-unsigned-5.15.0-46-generic' b'5.15.0' b'46.49' b'amd64'
[...]
```

The first commit adds support for this uncompressed Debian package format. The second commit refactors the control.tar extraction in DebQuery to avoid the conditional cascade.

### DIFF
--- a/osc/util/debquery.py
+++ b/osc/util/debquery.py
@@ -23,11 +23,6 @@ except ImportError:
     HAVE_ZSTD = False
 
 
-if (not hasattr(itertools, 'zip_longest') and hasattr(itertools, 'izip_longest')):
-    # python2 case
-    itertools.zip_longest = itertools.izip_longest
-
-
 class DebError(packagequery.PackageError):
     pass
 
@@ -65,9 +60,6 @@ class DebQuery(packagequery.PackageQuery, packagequery.PackageQueryResult):
             raise DebError(self._path, 'missing control.tar')
         try:
             name = './control'
-            # workaround for python2.4's tarfile module
-            if 'control' in tar.getnames():
-                name = 'control'
             control = tar.extractfile(name)
         except KeyError:
             raise DebError(self._path,
@@ -78,14 +70,13 @@ class DebQuery(packagequery.PackageQuery, packagequery.PackageQueryResult):
     def _open_tar(self, arfile):
         control = arfile.get_file(b'control.tar')
         if control:
-            # XXX: python2.4 relies on a name
-            return tarfile.open(name="control.tar", fileobj=control)
+            return tarfile.open(fileobj=control)
         return None
 
     def _open_tar_gz(self, arfile):
         control = arfile.get_file(b'control.tar.gz')
         if control:
-            return tarfile.open(name='control.tar.gz', fileobj=control)
+            return tarfile.open(fileobj=control)
         return None
 
     def _open_tar_xz(self, arfile):
@@ -94,8 +85,7 @@ class DebQuery(packagequery.PackageQuery, packagequery.PackageQueryResult):
             if not HAVE_LZMA:
                 raise DebError(self._path, 'can\'t open control.tar.xz without python-lzma')
             decompressed = lzma.decompress(control.read())
-            return tarfile.open(name="control.tar.xz",
-                                fileobj=BytesIO(decompressed))
+            return tarfile.open(fileobj=BytesIO(decompressed))
         return None
 
     def _open_tar_zst(self, arfile):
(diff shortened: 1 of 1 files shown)


osc/util/debquery.py
```diff
@@ -86,6 +72,40 @@ def read(self, all_tags=False, self_provides=True, *extra_tags):
         self.__parse_control(control, all_tags, self_provides, *extra_tags)
         return self
 
+    def __open_tar(self, arfile):
+        control = arfile.get_file(b'control.tar')
+        if control:
+            # XXX: python2.4 relies on a name
```
dmach, line 78: This is probably not needed as we support python 3.6+
Maybe just remove the comment? Not sure if it is still relevant.

osc/util/debquery.py
```diff
@@ -51,29 +51,15 @@ def read(self, all_tags=False, self_provides=True, *extra_tags):
...
+                          self.__open_tar_zst,
+                          self.__open_tar]:
+            tar = open_func(arfile)
+            if tar is not None:
+                break
+        if tar is None:
+            raise DebError(self.__path, 'missing control.tar')
         try:
             name = './control'
             # workaround for python2.4's tarfile module
```
dmach, line 65: This is probably not needed as we support python 3.6+

osc/util/debquery.py
```diff
@@ -86,6 +72,40 @@ def read(self, all_tags=False, self_provides=True, *extra_tags):
...
+
+    def __open_tar_xz(self, arfile):
+        control = arfile.get_file(b'control.tar.xz')
+        if control:
+            if not HAVE_LZMA:
+                raise DebError(self.__path, 'can\'t open control.tar.xz without python-lzma')
+            decompressed = lzma.decompress(control.read())
+            return tarfile.open(name="control.tar.xz",
+                                fileobj=BytesIO(decompressed))
+        return None
```
dmach, line 96: It might be better to reverse the logic (here and in all similar methods):
```
        control = arfile.get_file(b'control.tar.xz')
        if not control:
            return

        if not HAVE_LZMA:
            raise DebError(self.__path, 'can\'t open control.tar.xz without python-lzma')
        decompressed = lzma.decompress(control.read())
        return tarfile.open(name="control.tar.xz", fileobj=BytesIO(decompressed))
```

It saves indentation spaces and makes code a bit more readable.
But I'm rather nit-picking, if you like the existing code, feel free to keep it as is.

osc/util/debquery.py
```diff
@@ -86,6 +72,40 @@ def read(self, all_tags=False, self_provides=True, *extra_tags):
         self.__parse_control(control, all_tags, self_provides, *extra_tags)
         return self
 
+    def __open_tar(self, arfile):
```
dmach, line 75: I'm not sure if the leading double underscores are a good thing, because they lead to name mangling.
From https://peps.python.org/pep-0008/ :
```
__double_leading_underscore: when naming a class attribute, invokes name mangling
(inside class FooBar, __boo becomes _FooBar__boo).
```

Single leading underscores should be just fine.

dmach (changes requested):

raimue:
Hey, thank you for the review and the feedback. I agree with your suggestions, I wasn't sure about which python version was still meant to be supported, so I left it all as it was.

Unfortunately, I did not manage to get back to this before my summer vacation. Please bear with me that I am not working on this right now, but I will pick it up as soon as I return.

osc/util/debquery.py
raimue: I removed all of the python 2.x stuff for compatibility in395df87f.

This one referred to the `name=` argument that was always required for python 2.4 even when using `fileobj=`, but for later versions only the latter is required.

osc/util/debquery.py
raimue: I removed all of the python 2.x stuff for compatibility in395df87f.

osc/util/debquery.py
raimue: Looks reasonable. I added 241b34d6 to return early in these functions.

raimue:
I rebased the patch series onto the latest master to address the conflicts with master.

dmach:
The patch looks good to me, thanks!

PR #1135, opened 2022-09-06
Support ssh-agent forwarding

dmach:


mlschroe:
There's so much code deleted which kinda worries me. Does the code still work for users like me that don't use an ssh-agent?

dmach:
> There's so much code deleted which kinda worries me. Does the code still work for users like me that don't use an ssh-agent?

Yes, the code should work for you as usual.
The deleted code scans `~/.ssh` for keys and compares them with keys in ssh agent.
This is no longer needed, because we extract the first key from the ssh agent, store it in a temporary file and use it.
If the agent is not available/running or has no keys, we try to use one of the hard-coded keyfiles: `keyfiles = ('id_ed25519', 'id_ed25519_sk', 'id_rsa', 'id_ecdsa', 'id_ecdsa_sk', 'id_dsa')`

Could you eventually test this change by removing osc cookiejar and using osc from this pull request (https://github.com/dmach/osc/tree/ssh-agent-forwarding)?

Werkov:
Hi. I see this is merged (tagged in 1.0.0b2) but the osc on Tumbleweed (0.182) is a different branch.
What are the plans with this? (I'm just asking, I can live without the forwarding somehow.)

Issue #1141, opened 2022-09-12
Cannot search for requests created by myself


mlschroe:
`osc rq -M` should return requests created by myself, but that currently does not work.

We have a couple of issues:

Despite the request creator being shown as part of the history, it is no longer stored in the database like that. This means searching for `history.@who` will not work. This is clearly a bug in the API. We can add a workaround in osc to mitigate this by also searching for `@creator`, but old OBS versions may not support this.

The current code seems to treat `-M` as an abbreviation for `-U myusername`, which is not what the documentation says. We can either change the documentation or the implementation.

We should add a new option to search for the creator (`--creator` or `-C`). The can be done in current API versions by matching `@creator`.

PR #1144, opened 2022-09-15
Properly handle osc rq -M/--mine option

dmach:
fixes: #1141

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -1673,7 +1673,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                       % (devloc, dst_package))
                 sys.exit(1)
 
-        reqs = get_request_list(apiurl, dst_project, dst_package, req_type='submit', req_state=['new', 'review'])
+        reqs = get_request_collection(apiurl, project=dst_project, package=dst_package, types=['submit'], states=['new', 'review'])
         user = conf.get_apiurl_usr(apiurl)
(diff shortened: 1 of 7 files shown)


marcus-h:
Just a very high-level comment...

On 2022-09-15 12:18:56 -0700, Daniel Mach wrote:
> -- Commit Summary --
>
>   * Always list requests with &#39;new,review&#39; states by default
>   * Do not list declined requests by default - a change to osc library defaults

Hmm what about changing the defaults only for the newly introduced API and
keep the semantics of, for instance, get_request_list, as is?

>   * Properly handle osc rq -M/--mine option

Ugh, why do you explicitly add all these isinstance assertions?
IMHO, we should rather rely on duck-typing and assume that the API users
pass an instance that implements the "correct" interface...

>   * Improve &#39;osc rq show&#39; output
>   * core.Request - implement ordering compatible with py3

dmach:
>> * Do not list declined requests by default - a change to osc library defaults
>
> Hmm what about changing the defaults only for the newly introduced API and keep the semantics of, for instance, get_request_list, as is?

Good point, I'll do it that way.

>> * Properly handle osc rq -M/--mine option
>
> Ugh, why do you explicitly add all these isinstance assertions? IMHO, we should rather rely on duck-typing and assume that the API users pass an instance that implements the "correct" interface...

I'm aware of duck-typing and type checking.
I considered these asserts relatively safe (considering osc context), but I'm ok with removing them.

PR #1145, opened 2022-09-16
Be verbose about osc vc using the EDITOR envar to the help output

luc14n0:
This is a nice to have information when we use `osc help vc` / `osc vc --help` so people can use whatever editor they feel comfortable with.


dmach:
Thanks for the patch, merging.

PR #1151, opened 2022-09-29
add create-pbuild-config(cpc) command

adrianschroeter:
which allows to create a working config for pbuild tool based on OBS server side data

The cross build support is not yet done

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -8345,6 +8345,35 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             if len(user) == 3:
                 print("%s: \"%s\" <%s>" % (user[0], user[1], user[2]))
 
+    @cmdln.name("create-pbuild-config")
+    @cmdln.alias('cpc')
+    @cmdln.option("repository")
+    @cmdln.option("arch")
+    def do_create_pbuild_config(self, subcmd, opts):
+        """
+        This command is creating the necessary files to build using pbuild tool.
+        It basically creates _config and _pbuild file in the project directory.
+        Changes from there can not get submitted back, except the project is managed
+        in git.
+
+        Examples:
+            osc cpc REPOSITORY ARCH
+        """
+
+        apiurl = self.get_api_url()
+        project = None
+        project_dir = os.curdir
+        if is_project_dir(project_dir):
+            project = store_read_project(project_dir)
+        elif is_package_dir(project_dir):
+            project_dir += '/..'
+            project = store_read_project(project_dir)
+        else:
+            raise oscerr.WrongArgs('Creating pbuild only works in a checked out project or package')
+
+        create_pbuild_config(apiurl, project, opts.repository, opts.arch, project_dir)
+
+
     @cmdln.option('-r', '--revision', metavar='rev',
                   help='print out the specified revision')
     @cmdln.option('-e', '--expand', action='store_true',
--- a/osc/core.py
+++ b/osc/core.py
@@ -6379,6 +6379,39 @@ def get_buildconfig(apiurl, prj, repository, path=None):
     return f.read()
 
 
+def create_pbuild_config(apiurl, project, repository, arch, project_dir):
+    """
+    This is always replacing a possible exiting config for now
+    we could extend the _pbuild file easily, but what should we do with multiple instances of the _config?
+    """
+    # get expanded buildconfig for given project and repository
+    bc = get_buildconfig(apiurl, project, repository)
+    if not bc:
+        msg = "Failed to get build config for project '{project}', repository '{repository}'"
+        raise oscerr.NotFoundAPIError(msg)
+
+    with open(os.path.join(project_dir, '_config'), "w") as f:
+        f.write(decode_it(bc))
+
+    # create the _pbuild file based on expanded repository path informations
+    pb = ET.fromstring('<pbuild></pbuild>')
+    tree = ET.ElementTree(pb)
+    preset = ET.SubElement(pb, 'preset', name=repository, default="")  # default should be empty, but ET crashes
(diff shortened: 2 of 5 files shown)


osc/commandline.py
```diff
@@ -8342,6 +8342,36 @@ def do_whois(self, subcmd, opts):
             if len(user) == 3:
                 print("%s: \"%s\" <%s>" % (user[0], user[1], user[2]))
 
+    @cmdln.alias('cpc')
+    def do_createpbuildconfig(self, subcmd, opts, *args):
```
dmach, line 8346: Wouldn't be better use name with dashes: `create-pbuild-config` ?
There's no explicit support for that in cmdln.py, but I could add it easily.

osc/core.py
```diff
@@ -6378,6 +6378,39 @@ def get_buildconfig(apiurl, prj, repository, path=None):
     f = http_GET(u)
     return f.read()
 
+# This is always replacing a possible exiting config for now
+# we could extend the _pbuild file easily, but what should we do with multiple instances of the _config?
+def create_pbuild_config(apiurl, project, repository, arch, project_dir):
+    # get expanded buildconfig for given project / repository
+    bc = get_buildconfig(apiurl, project, repository)
+    if bc == None:
```
dmach, line 6386: Do not compare with None using `==`.
Please change to `if not bc:`

osc/core.py
```diff
@@ -6378,6 +6378,39 @@ def get_buildconfig(apiurl, prj, repository, path=None):
...
+# cross compile setups are not yet supported
+#    for path in root.findall('hostsystem'):
+#        ET.SubElement(preset, 'hostrepo').text = path.get('url')
+    for path in root.findall('path'):
+        ET.SubElement(preset, 'repo').text = path.get('url')
+
+    ET.SubElement(preset, 'arch').text = arch
+    try:
+        ET.indent(tree, '  ')
+    except:
```
dmach, line 6409: lgtm checker already complained: `1 for Except block handles 'BaseException'`
Could you specify the exception type?

osc/core.py
```diff
@@ -6378,6 +6378,39 @@ def get_buildconfig(apiurl, prj, repository, path=None):
...
+        # old versions do not support it, but the file is just not nicely formated then ...
+        None
```
dmach, line 6411: Replace with `pass` ?

dmach (changes requested):

osc/commandline.py
```diff
@@ -8342,6 +8342,36 @@ def do_whois(self, subcmd, opts):
...
+
+        Examples:
+            osc cpc REPOSITORY ARCH
+        """
+
+        apiurl = self.get_api_url()
+        if len(args) != 2:
+            raise oscerr.WrongArgs('Wrong number of arguments, require a repository and architecture')
+        repository = args[0]
+        arch = args[1]
```
dirkmueller, line 8361: ```repository, arch = args```

osc/core.py
```diff
@@ -6378,6 +6378,39 @@ def get_buildconfig(apiurl, prj, repository, path=None):
     f = http_GET(u)
     return f.read()
 
+# This is always replacing a possible exiting config for now
+# we could extend the _pbuild file easily, but what should we do with multiple instances of the _config?
+def create_pbuild_config(apiurl, project, repository, arch, project_dir):
```
dirkmueller, line 6383: add "_" as prefix?

osc/core.py
```diff
@@ -6378,6 +6378,39 @@ def get_buildconfig(apiurl, prj, repository, path=None):
     f = http_GET(u)
     return f.read()
 
+# This is always replacing a possible exiting config for now
```
dirkmueller, line 6381: move comment into a ```""""..."""``` section at the beginning of the function

osc/core.py
```diff
@@ -6378,6 +6378,39 @@ def get_buildconfig(apiurl, prj, repository, path=None):
...
+# we could extend the _pbuild file easily, but what should we do with multiple instances of the _config?
+def create_pbuild_config(apiurl, project, repository, arch, project_dir):
+    # get expanded buildconfig for given project / repository
+    bc = get_buildconfig(apiurl, project, repository)
+    if bc == None:
+        e.osc_msg = 'failed to get build config for project \'%s\' repository \'%s\'' % (project, repository)
+        raise
+
+    file = open(project_dir + '/_config', 'w')
```
dirkmueller, line 6390:
``` with open(...) as f:
       f.write(decode_it(bc))
```

avoids the flush and the fd leak

osc/core.py
```diff
@@ -6378,6 +6378,39 @@ def get_buildconfig(apiurl, prj, repository, path=None):
...
+    for path in root.findall('path'):
+        ET.SubElement(preset, 'repo').text = path.get('url')
+
+    ET.SubElement(preset, 'arch').text = arch
+    try:
+        ET.indent(tree, '  ')
+    except:
+        # old versions do not support it, but the file is just not nicely formated then ...
+        None
+    tree.write(project_dir + '/_pbuild', encoding="utf-8", xml_declaration=True)
```
dirkmueller, line 6412:
```suggestion
    tree.write(os.path.join(project_dir, '_pbuild'), encoding="utf-8", xml_declaration=True)
```

adrianschroeter:
the _ prefix for the method is not working, but if you mind it, we can just move the code into commandline

apart from that I should be done, @dmach please take over or continue to complain :)

osc/core.py
dmach: I've removed this code, replacing it with `xmlindent(tree)`

osc/commandline.py
dmach: no longer needed, args are handled via argparse and that includes throwing an error if they're not specified

dmach (approved):

PR #1153, opened 2022-09-30
Wrong documentation URL for maintenancerequest and createincident

mackdk:
This PR fixes the documentation URL for the commands `maintenancerequest` and `createincident`. The documentation is currently pointing to the non-existing http://openbuildservice.org/help/manuals/obs-reference-guide/cha.obs.maintenance_setup.html, while the correct page should be https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.maintenance_setup.html.

This also reflects to the man page of the osc command.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -3199,7 +3199,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         """
         Create a maintenance incident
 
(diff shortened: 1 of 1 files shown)


dmach:
Thanks for fixing the URLs!

PR #1154, opened 2022-09-30
Use a more readable 'if ... in ' construct

dirkmueller:


dmach:
This makes the code more readable, thanks!

PR #1164, opened 2022-10-07
expand "." in copypac and rdelete as referencing the workdir project

dirkmueller:
dot (".") is already used as magic "current project/package" in other commands like "ls" for example, but in the most useful ones like copypac and rdelete it was missing. this adds a function that does the dot expansion and adds it to the respective command expansions.


dirkmueller:
sorry, accidentally pushed an old branch

dmach:
Looks good to me, merging.

Issue #1150, opened 2022-09-27
osc request list -i -m ok --involved-projects -s new doesn't send to Factory on accept


mcepl:
**Describe the bug**
When going through my request with `osc request list -i -m ok --involved-projects -s new` and accept a request, it is not send (and it is not offered) to Factory.

**Versions**
- Operating system: Tumbleweed
- Packages: 0.182.0

**To Reproduce**
1. Run command `osc request list -i -m ok --involved-projects -s new`
2. accept a request

**Expected behavior**
New request is created to Factory (as when the request is accepted in the web interface)

PR #1166, opened 2022-10-12
sr accept: Enable forwarding requests to the parent projects; Introduce new osc._private module

dmach:
fixes: #1150


mcepl:
So, I have declined a request and osc still offers sending a new one to Factory. Whoops!

dmach:
> So, I have declined a request and osc still offers sending a new one to Factory. Whoops!

Good catch!
I hope I've fixed this in the last push.

dirkmueller (approved):

PR #1167, opened 2022-10-12
Add "--extra-pkgs-from" ("-X") option to osc build

OlegGirko:
This option adds extra packages listed in the specified file to build.

For now, `osc` does not support automatic buildrequires.

When a package has automatic buildrequires, `osc` just returns error code 9 that is returned by build,
but build leaves a list of missing dependencies in `.build.packages/OTHER/_generated_buildreqs` file inside build root.

These extra packages can be added using `--extra-pkgs` (`-x`) option, but this is very inconvenient if there are many of them.

Allowing to add extra dependencies listed in a file makes building packages with automatic buildrequires much more convenient: just do a first stage build, resulting in a file with list of extra dependencies, and then add extra packages from this file using `--extra-pkgs-from` (`-X`) option that is added by this change.

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -868,6 +868,12 @@ def main(apiurl, opts, argv):
     elif opts.extra_pkgs != ['']:
         extra_pkgs = opts.extra_pkgs
 
+    if opts.extra_pkgs_from:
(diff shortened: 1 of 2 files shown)


dmach:
:+1:

PR #1168, opened 2022-10-13
Use tuples when possible

dirkmueller:


osc/commandline.py
```diff
@@ -2324,9 +2324,9 @@ def do_request(self, subcmd, opts, *args):
 
         apiurl = self.get_api_url()
 
-        if cmd in ['list']:
+        if cmd in ('list'):
```
dmach, line 2327: This is incorrect, the parentheses are ignored and the following is evaluated: `if cmd in 'list'`
To use a tuple properly, you need to add a trailing comma: `if cmd in ('list', )` or compare the value directly: `if cmd == 'list'`

dmach (changes requested):

osc/commandline.py
dirkmueller: sure, fixed

dmach (approved):

PR #1172, opened 2022-10-13
Mute ssl/tls deprecation warnings

dmach:
ssl.OP_NO_SSLv2 - removed, because it's deprecated since py3.6
ssl.OP_NO_SSLv3 - removed, because it's deprecated since py3.6
ssl.OP_NO_TLSv1 - enabled only on py3.6, because it's deprecated since py3.7
ssl.OP_NO_TLSv1_1 - enabled only on py3.6, because it's deprecated since py3.7

Link to docs: https://docs.python.org/3/library/ssl.html

### DIFF
--- a/osc/oscssl.py
+++ b/osc/oscssl.py
@@ -28,10 +28,15 @@ def create_ssl_context():
     but we restrict crypto even more.
     """
     ssl_context = create_urllib3_context()
-    ssl_context.options |= ssl.OP_NO_SSLv2
-    ssl_context.options |= ssl.OP_NO_SSLv3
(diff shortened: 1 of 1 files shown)


dmach:
> according to https://docs.python.org/3/library/ssl.html#ssl.SSLContext.options it is recommended to use `ssl_context.minimum_version` instead, but I don't see that here. can we add an explicit test or set for minimum_version? to TLSv1_2 ?

Good idea.
The `minimum_version` is supported from py3.7 while osc is py3.6+, but I'm sure I'll figure it out :)

Issue #1173, opened 2022-10-18
osc co $PACKAGE is not working anymore


adrianschroeter:
We used to have the convinience functionality to checkout a package while CWD is a project checkout. 

This is very handy in larger projects which you do not want to checkout entirely. current osc from OBS:Server:Unstable is crashing with

# osc co python-home-assistant-bluetooth
Traceback (most recent call last):
  File "/usr/bin/osc", line 33, in <module>
    sys.exit(load_entry_point('osc==1.0.0b2', 'console_scripts', 'osc')())
  File "/usr/lib/python3.10/site-packages/osc/babysitter.py", line 208, in main
    sys.exit(run(commandline.Osc()))
  File "/usr/lib/python3.10/site-packages/osc/babysitter.py", line 56, in run
    return prg.main(argv)
  File "/usr/lib/python3.10/site-packages/osc/cmdln.py", line 262, in main
    cmd(self.options.command, self.options, *self.args)
  File "/usr/lib/python3.10/site-packages/osc/commandline.py", line 4536, in do_checkout
    checkout_package(apiurl, project, package, rev, expand_link=expand_link,
  File "/usr/lib/python3.10/site-packages/osc/core.py", line 5128, in checkout_package
    prj_dir = prj_dir.replace(':', conf.config['project_separator'])
TypeError: Path.replace() takes 2 positional arguments but 3 were given

PR #1174, opened 2022-10-18
prj_dir is a path object. Need to be cast to str.

lethliel:
And only run replace code if 'project_separator' is set in oscrc

fixes https://github.com/openSUSE/osc/issues/1173


dirkmueller:
@lethliel sorry, I had noticed this also, but fixed it slightly differently in #1175

dmach:
Let's use #1175 instead because it contains more fixes and fixes this problem as well.

PR #1175, opened 2022-10-18
Merge _expand_as_project_name into _process_project_name

dirkmueller:
This function was already doing most of the work, so we can do this
extra there as well. simplifies the code a little. also fix a regression
in previous change where Pwd.cwd().replace() would rename the directory
(rather than doing a replace on the string value).

fixes https://github.com/openSUSE/osc/issues/1173


dmach:
I've tested this and it fixes #1173.
It seems we should refactor the commandline code to be able to write unit tests - possibly add a hook after parsing arguments and setting local variables.

Issue #1176, opened 2022-10-19
osc rq list -a no longer accepted


dirkmueller:
**Describe the bug**

invoking `osc rq list -a` gives:

```
$ osc rq list -a
Sorry, the options '--all' and '--state' are mutually exclusive.
```

`osc rq list -s all` is still working.

**Versions**
- Operating system: `Tumbleweed`
- Packages: `osc 1.0.0b2+100`

PR #1178, opened 2022-10-20
rq list --all: Avoid conflict with the --state option

dmach:
Fixes: #1176

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -2169,7 +2169,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                         help='all states. Same as\'-s all\'')
     @cmdln.option('-f', '--force', action='store_true',
                         help='enforce state change, can be used to ignore open reviews')
-    @cmdln.option('-s', '--state', default='new,review',
+    @cmdln.option('-s', '--state',
                         help='only list requests in one of the comma separated given states (new/review/accepted/revoked/declined) or "all" [default="new,review"]')
     @cmdln.option('-D', '--days', metavar='DAYS',
                         help='only list requests in state "new" or changed in the last DAYS.')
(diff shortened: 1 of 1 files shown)


osc/commandline.py
```diff
@@ -2300,6 +2300,12 @@ def do_request(self, subcmd, opts, *args):
             raise oscerr.WrongOptions('Sorry, the options \'--interactive\' and '
                                       '\'--non-interactive\' are mutually exclusive')
 
+        if opts.all:
+            state_list = ["all"]
+        else:
+            state_list = opts.state.split(",")
```
dirkmueller, line 2306: where is the default handling gone to ?

osc/commandline.py
```diff
@@ -2423,15 +2429,9 @@ def do_request(self, subcmd, opts, *args):
...
-                if state_list == ['']:
-                    state_list = ()
-                if opts.all:
-                    state_list = ['all']
-                else:
-                    for s in state_list:
-                        if s not in states and not s == 'all':
-                            raise oscerr.WrongArgs('Unknown state \'%s\', try one of %s' % (s, ','.join(states)))
+                for s in state_list:
+                    if s not in states and not s == 'all':
```
dirkmueller, line 2433: NIT: I know it has been that way before

```suggestion
                    if s!= 'all' and s not in states:
```

osc/commandline.py
dmach: Good catch, fixed.

osc/commandline.py
dmach: The proposed change makes sense to me - it's definitely an improvement to code readability. Thanks.

dirkmueller (approved):

PR #1179, opened 2022-10-21
Add type annotations

SchoolGuy:
This PR should achieve two things:

- Add a reasonable amount of type annotations for `core.py` to enable external consumers of `osc` as a library to better know what should be passed to methods.
- Automatically check in the CI that the code quality according to [mypy](http://mypy-lang.org/) and [pylint](https://pylint.pycqa.org/en/latest/) is not getting worse.

@dmach The new CI will obviously go nuts on the current code. I hope to collaboratively work out the details of a configuration that is useful for the current state of the project that is ultimately getting stricter the more the project improves over time.

### DIFF
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -13,6 +13,29 @@ on:
       - '**.py'
 
 jobs:
+  mypy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+      - run: pip install mypy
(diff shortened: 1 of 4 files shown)


SchoolGuy:
@dmach We talked about this PR a while ago. Since the changes are rather big if we turn on the linters directly, do you see a way to apply this somehow in better reviewable chunks for you? The linters won't be able to run on Diffs of a PR. My experiments failed in this regard sadly.

dmach:
LGTM, merging. Thanks!

PR #1181, opened 2022-10-24
Introduce Store class, migrate some functionality to it

dmach:


osc/_private/store.py
```diff
@@ -0,0 +1,316 @@
...
+        return project_name
+
+    @classmethod
+    def resolve_package_name(cls, package_name):
+        """
+        Resolve special package names:
+            "." -> package name from the current working copy of a package
+        """
+        if package_name == ".":
+            store = cls(os.getcwd())
```
dirkmueller, line 39: Use `Path.cwd()|

dmach: I have decided to remove these classmethods completely and bring them back when they'll be used.
I'm not entirely sure how they should look like.

PR #1184, opened 2022-10-26
Cleanups

dirkmueller:


dmach:
@dirkmueller unittests are failing because you made `boolean_opts` and `integer_opts` unavailable.
It might be good to revert the change, maybe prefix the variables with an underscore to make clear they're private.
Another option is cloning the variables to the tests.

PR #1188, opened 2022-11-02
Avoid crash when deleting a password

mig4:
When using keyring, osc would crash when called as `osc config ENDPOINT --change-password`
and when the password didn't exist in the backend.

This prevents it by first checking if a password exists.


dmach:
Thanks for the patch, it looks good to me, merging.

PR #1200, opened 2022-12-07
mypy: Add types to methods

SchoolGuy:
This is a bulk add of typing that also includes the formatting with darker.


dmach:
LGTM, merging. Thanks!

PR #1027, opened 2022-04-11
add another exception for github URLs for "osc add"

adrianschroeter:
osc#905

PR #1204, opened 2022-12-12
Fix osc add for github /archive/ URLs

z3ntu:
"osc add https://github.com/foo/bar/archive/12345.tar.gz" should treat the URL as archive and not as git source.

---

Similar to #1027


dmach:
Thanks for the patch, merging.

PR #1208, opened 2022-12-21
Tokens: Implement all current operations

### LINKED ISSUE #1194: osc token --trigger should allow additional arguments
osc token --trigger --operation release only allows to pass project and package, while the `/trigger/release` route accepts also additional argments. 

https://build.opensuse.org/apidocs-new/#/Trigger/post_trigger__operation_ documents repository and arch - while IBS seems to be patched to also support targetproject and others (hurray!)

SchoolGuy:
Fixes #1194 

This also introduces a module that enables osc to validate the XML answers according to the XML Schemas that an OBS instance is making available.

> Note: This is a Draft atm and subject to change. However please comment your opinions as this will influence my progress heavily.

### DIFF
--- a/osc/_private/response.py
+++ b/osc/_private/response.py
@@ -1,0 +1,53 @@
+"""
+This module contains all common responses that osc will regularly parse.
+"""
+
+from dataclasses import dataclass, field
+from typing import List, TYPE_CHECKING
+
+
+from osc._private.xml import XmlParser, XmlSchemaType
+
+if TYPE_CHECKING:
+    from lxml import etree
+
+
+@dataclass
+class Status:
+    """
+    Represents https://build.opensuse.org/schema/status.xsd
+    """
+    summary = ""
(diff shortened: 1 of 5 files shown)


SchoolGuy:
I assumed that `lxml` is used for XML parsing. It seems this would be a new dependency. I will take a look if it is possible to use the stdlib or if `lxml` is really required.

Conan-Kudo:
`lxml` is a drop-in replacement that makes XML processing go a lot more quickly.

SchoolGuy:
This PR is still waiting for the server-side PR to be merged. The discussion about `lxml` needs to be done once the implementation is finished.

dmach:
I've submitted an alternative implementation that follows the latest best practices: #1550

SchoolGuy:
I believe your new implementation @dmach is a much better fit than this PR. As such I will close my PR now.

PR #1212, opened 2023-01-16
behave: Speed running tests up by preparing containers in advance

dmach:


dmach:
Execution of `tests / behave` took 11m instead of 12m30s.
The improvement (roughly 10%) is not as great as on my laptop, but still counts...

PR #1214, opened 2023-01-17
GHA: Print pylint logs before diffing them

dmach:


.github/workflows/linters.yaml
```diff
@@ -64,11 +68,30 @@ jobs:
           repository: ${{ matrix.plugin_repository }}
           path: plugin
 
+      - name: 'Run pylint with osc from master'
+        run: |
+            cd plugin
+            find . -type f -name "*.py" | PYTHONPATH="../osc-master" xargs pylint --errors-only >> pylint-osc-master.log || :
```
dirkmueller, line 74: this should be `find | sort | xargs ..`

.github/workflows/linters.yaml
```diff
@@ -64,11 +68,30 @@ jobs:
...
+
+            echo
+            echo ">>>>> pylint-osc-master.log <<<<<"
+            cat pylint-osc-master.log
+
+      - name: 'Run pylint with osc from pull-request'
+        run: |
+            cd plugin
+            find . -type f -name "*.py" | PYTHONPATH="../osc-pr" xargs pylint --errors-only >> pylint-osc-pr.log || :
```
dirkmueller, line 83: here as well

dirkmueller (changes requested):
the cat into a sorted file without actually sorting it is confusing.

dirkmueller (approved):
LGTM

PR #1242, opened 2023-01-29
Fix regression in osc rebuild

andreas-schwab:


dmach:
Thanks for the patch. It fixes the main problem.
I'll probably throw some additional fixes and tests later in a new PR...

Issue #1258, opened 2023-02-08
patchinfo crashes on path handling


bhdn:
**Describe the bug**

When running `osc patchinfo -f` inside a project it crashes.

**Versions**
- Operating system: openSUSE Tumbleweed
- Packages: osc-1.0.0~b4-1.1.noarch

**To Reproduce**

```
osc patchinfo -f
Creating new patchinfo...
Traceback (most recent call last):
  File "/usr/bin/osc", line 33, in <module>
    sys.exit(load_entry_point('osc==1.0.0b4', 'console_scripts', 'osc')())
  File "/usr/lib/python3.10/site-packages/osc/babysitter.py", line 209, in main
    sys.exit(run(commandline.Osc()))
  File "/usr/lib/python3.10/site-packages/osc/babysitter.py", line 57, in run
    return prg.main(argv)
  File "/usr/lib/python3.10/site-packages/osc/cmdln.py", line 262, in main
    cmd(self.options.command, self.options, *self.args)
  File "/usr/lib/python3.10/site-packages/osc/commandline.py", line 1026, in do_patchinfo
    checkout_package(apiurl, project, patchinfo, prj_dir=project_dir)
  File "/usr/lib/python3.10/site-packages/osc/core.py", line 5394, in checkout_package
    prj_dir = prj_dir.replace(':', conf.config['project_separator'])
TypeError: Path.replace() takes 2 positional arguments but 3 were given
```

bhdn:
After a bisec session, it turned to be introduced by
bac162a7123524fb3521e52c90d05f6577b7c25e.

PR #1259, opened 2023-02-08
checkout_package: Use pathlib.Path

bhdn:
Modify it to use pathlib.Path along with its callers. Some str(path) was necessary to avoid having a large changeset.

It addresses #1258 and continues the work started in bac162a7123524fb3521e52c90d05f6577b7c25e.


dmach:
Looks good to me, thanks.

PR #1265, opened 2023-02-16
Pass --force down to delete_package

andreas-schwab:


dmach:
@andreas-schwab what problem are you trying to solve?
The `--force` option is to `force commit, even if there were no changes`
You seem to be using it for something else.

andreas-schwab:
See 5d31fc14.

andreas-schwab:
See 5d31fc1.

PR #1266, opened 2023-02-17
Fix list no copy attribute on python2

Alix82:
On python2 there is no list.copy method, fix this by using python 2 copy style.

Reference bsc#1208010


dmach:
Looks good to me, merging.
Thanks!

Issue #1230, opened 2023-01-25
Persistent hdrmd5 mismatches between builds


bnavigator:
**Describe the bug**

I am seeing this a lot lately, causing the re-download of some packages on every local `osc build`:

```
28.1% cache miss. 394/548 dependencies cached.

[1/154] (devel:languages:python) python38-pip-22.3.1-1.11.noarch.rpm: 100%|##########################################################################################################| Time: 0:00:00
devel:languages:python/python38-pip: allowing invalid file, probably an OBS bug - hdrmd5 did not match - 0d675712efdf39f805f5df5883704313 != 5e6287d2f64d397a9ef27a682a2aad16
[2/154] (devel:languages:python) python39-pip-22.3.1-1.11.noarch.rpm: 100%|##########################################################################################################| Time: 0:00:00
devel:languages:python/python39-pip: allowing invalid file, probably an OBS bug - hdrmd5 did not match - 03b5b08bd88c633316b80ce712e05ae3 != 6e969fc0b97d29284c429eb3a63d7bd4
...
Verifying integrity of cached packages
using keys from openSUSE:Tumbleweed, devel:languages:python, openSUSE:Tumbleweed:OldKey
WARNING: OBS BUG hdrmd5 mismatch for /var/tmp/osbuild-packagecache/devel:languages:python/openSUSE_Tumbleweed/x86_64/python38-pip-22.3.1-1.11.noarch.rpm: 0d675712efdf39f805f5df5883704313 != 5e6287d2f64d397a9ef27a682a2aad16
WARNING: OBS BUG hdrmd5 mismatch for /var/tmp/osbuild-packagecache/devel:languages:python/openSUSE_Tumbleweed/x86_64/python39-pip-22.3.1-1.11.noarch.rpm: 03b5b08bd88c633316b80ce712e05ae3 != 6e969fc0b97d29284c429eb3a63d7bd4
```

On the next build only a few seconds later:
```
23.5% cache miss. 419/548 dependencies cached.

[1/129] (devel:languages:python) python38-pip-22.3.1-1.11.noarch.rpm: 100%|##########################################################################################################| Time: 0:00:00
devel:languages:python/python38-pip: allowing invalid file, probably an OBS bug - hdrmd5 did not match - 0d675712efdf39f805f5df5883704313 != 5e6287d2f64d397a9ef27a682a2aad16
[...]

PR #1267, opened 2023-02-20
build: New option 'disable_hdrmd5_check' to ignore hdrmd5 mismatches

dmach:
Fixes: #1230

### DIFF
--- a/osc/fetch.py
+++ b/osc/fetch.py
@@ -205,6 +205,7 @@ class Fetcher:
         return urllist
 
     def run(self, buildinfo):
+        apiurl = buildinfo.apiurl
         cached = 0
         all = len(buildinfo.deps)
         for i in buildinfo.deps:
@@ -221,12 +222,24 @@ class Fetcher:
                 cached += 1
                 if not i.name.startswith('container:') and i.pacsuffix != 'rpm':
                     continue
+
+                hdrmd5_is_valid = True
                 if i.hdrmd5:
                     if i.name.startswith('container:'):
                         hdrmd5 = dgst(i.fullfilename)
+                        if hdrmd5 != i.hdrmd5:
+                            hdrmd5_is_valid = False
                     else:
                         hdrmd5 = packagequery.PackageQuery.queryhdrmd5(i.fullfilename)
-                    if not hdrmd5 or hdrmd5 != i.hdrmd5:
+                        if hdrmd5 != i.hdrmd5:
+                            if conf.config["api_host_options"][apiurl]["disable_hdrmd5_check"]:
+                                print(f"Warning: Ignoring a hdrmd5 mismatch for {i.fullfilename}: {hdrmd5} (actual) != {i.hdrmd5} (expected)")
+                                hdrmd5_is_valid = True
+                            else:
+                                print(f"The file will be redownloaded from the API due to a hdrmd5 mismatch for {i.fullfilename}: {hdrmd5} (actual) != {i.hdrmd5} (expected)")
+                                hdrmd5_is_valid = False
+
+                    if not hdrmd5_is_valid:
                         os.unlink(i.fullfilename)
                         cached -= 1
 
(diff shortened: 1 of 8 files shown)


Firstyear (approved):
Looks good to me :)

osc/fetch.py
```diff
@@ -221,12 +221,22 @@ def run(self, buildinfo):
                 cached += 1
                 if not i.name.startswith('container:') and i.pacsuffix != 'rpm':
                     continue
+
+                hdrmd5_is_valid = True
                 if i.hdrmd5:
                     if i.name.startswith('container:'):
                         hdrmd5 = dgst(i.fullfilename)
+                        if not hdrmd5 or hdrmd5 != i.hdrmd5:
+                            hdrmd5_is_valid = False
```
adrianschroeter, line 230: pitnick: could be simplified to

if hdrmd5 != i.hdrmd5:

because None is also not equal, no?

osc/fetch.py
```diff
@@ -221,12 +221,22 @@ def run(self, buildinfo):
...
                     else:
                         hdrmd5 = packagequery.PackageQuery.queryhdrmd5(i.fullfilename)
-                    if not hdrmd5 or hdrmd5 != i.hdrmd5:
+                        if not hdrmd5 or hdrmd5 != i.hdrmd5:
+                            # FIXME: We're currently ignoring the hdrmd5 verification of rpm packages.
+                            # look for "HDRMD5 MISMATCH EXPLANATION" below
+                            print("Warning: Ignoring a hdrmd5 mismatch for %s: %s (actual) != %s (expected)" % (i.fullfilename, hdrmd5, i.hdrmd5))
+                            hdrmd5_is_valid = True
```
adrianschroeter, line 237: this should be really only done via a switch. Or one looks endless for issues which are not reproducible locally due to broken noarch rpms (which are in fact not noarch)

osc/fetch.py
```diff
@@ -258,19 +268,26 @@ def run(self, buildinfo):
...
+                        # OBS builds the noarch packages once per binary arch.
+                        # Such noarch packages are supposed to be nearly identical across all build arches,
+                        # any discrepancy in the payload and dependencies is considered a packaging bug.
+                        # But to guarantee that the local builds work identically to builds in OBS,
+                        # using the arch-specific copy of the noarch package is required.
+                        # Unfortunatelly only one of the noarch packages gets distributed
+                        # and can be downloaded from a local mirror.
+                        # All other noarch packages are available through the OBS API only.
+                        # Since there is currently no information about hdrmd5 checksums of published noarch packages,
+                        # we download them, verify hdrmd5 and re-download the package from OBS API on mismatch.
```
adrianschroeter, line 281: noarch are the most prominent examples, but the same can also happen for architecture depend packages. eg. when you have multiple candidates in a project or when someone is messing around with the source history or the release number handling in a way that it is not increasing.

osc/fetch.py
```diff
@@ -258,19 +268,26 @@ def run(self, buildinfo):
...
                         #
-                        # wbrown 2022 - is there a reason to keep these md5's at all? md5 is
-                        # broken from a security POV so these aren't a trusted source for validation
-                        # of the file content. They are often incorrect forcing download via the API
-                        # which for anyone outside the EU is excruciating. And when they are ignored
-                        # builds work and progress anyway? So what do they even do? What are they
-                        # for? They should just be removed.
+                        # FIXME: We're currently ignoring the hdrmd5 verification of rpm packages.
+                        # A proper fix requires OBS change such as described in
+                        # https://github.com/openSUSE/open-build-service/issues/13821
```
adrianschroeter, line 285: please do not ignore it by default. The default should be really safe and reproduce the server side state as highest goal.

An optional switch to disable it should be implemented within a minute ....

adrianschroeter (changes requested):
Please do not switch off the check by default, the reproducability of server side builds is really our highest goal with osc build.

osc/fetch.py
dmach: Agreed. `disable_hdrmd5_check` is going to be part of the changes I'm about to force-push.

osc/fetch.py
dmach: Yes, the `disable_hdrmd5_check` option will default to `0` (the original behavior when the exact checksums were required)

osc/fetch.py
dmach: Thanks for the details, I've used part of your text in the comment explaining the situation.

osc/fetch.py
dmach: Changed.

adrianschroeter (approved):

bnavigator:
The description of this PR is incomplete. It does not only add an option to ignore the hdrmd5 but actually makes sure to download the correct package from API instead of published repos if ignoring is not an option.

PR #1286, opened 2023-03-30
Prefer f-strings ove c-style string expansion

dirkmueller:
This exists since Python >= 3.6 and is shorter and easier to read.


dmach:
Looks good to me, merging.

PR #1290, opened 2023-04-03
Fix creq command not working

hellcp-work:
I discovered that this command no longer worked, but I needed it, so I fixed it.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -2369,27 +2369,6 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             change_request_state(apiurl, opts.supersede, 'superseded',
                                  opts.message or '', result)
 
-    def _actionparser(self, opt_str, value, parser):
-        value = []
-        if not hasattr(parser.values, 'actiondata'):
-            setattr(parser.values, 'actiondata', [])
-        if parser.values.actions is None:
-            parser.values.actions = []
-
-        rargs = parser.rargs
-        while rargs:
-            arg = rargs[0]
-            if ((arg[:2] == "--" and len(arg) > 2) or
-                    (arg[:1] == "-" and len(arg) > 1 and arg[1] != "-")):
-                break
-            else:
-                value.append(arg)
-                del rargs[0]
-
-        parser.values.actions.append(value[0])
-        del value[0]
-        parser.values.actiondata.append(slash_split(value))
-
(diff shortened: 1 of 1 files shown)


osc/commandline.py
```diff
@@ -2663,9 +2642,8 @@ def _set_bugowner(self, args, opts):
 
         return actionxml
 
-# TODO: fix ValueError: unknown action "callback"
-#    @cmdln.option('-a', '--action', action='callback', callback = _actionparser, dest = 'actions',
-#                  help='specify action type of a request, can be : submit/delete/change_devel/add_role/set_bugowner')
+    @cmdln.option('-a', '--action', action='append', nargs='+', metavar=('ACTION', '[ARGS]'), dest = 'actions',
```
dmach, line 2645: Remove spaces from `dest = 'actions'` -> `dest='actions'` (nit-picking, but since you're touching the line...)
Add `default=[]` otherwise `osc creq` fails with a traceback.

osc/commandline.py
```diff
@@ -2723,42 +2701,29 @@ def do_createrequest(self, subcmd, opts, *args):
...
                 actionsxml += self._add_user(args, opts)
-                i = i + 1
-            elif ai == 'set_bugowner':
-                args = opts.actiondata[i]
+            elif action == 'set_bugowner':
                 actionsxml += self._set_bugowner(args, opts)
-                i = i + 1
             else:
-                raise oscerr.WrongArgs('Unsupported action %s' % ai)
+                raise oscerr.WrongArgs('Unsupported action %s' % action)
```
dmach, line 2726: Would be nice to replace it with a double-quoted f-string:
`raise oscerr.WrongArgs(f"Unsupported action {action}")`

osc/commandline.py
```diff
@@ -2723,42 +2701,29 @@ def do_createrequest(self, subcmd, opts, *args):
...
         apiurl = self.get_api_url()
 
-        i = 0
         actionsxml = ""
         supersede = set()
-        for ai in opts.actions:
-            if ai == 'submit':
-                args = opts.actiondata[i]
-                i = i + 1
+        for actiondata in opts.actions:
```
dmach, line 2706: If the default mentioned above is not set, this fails on iterating over `None`.

dmach (changes requested):

PR #1301, opened 2023-04-14
clean-up .old folder if an exception happens

marxin:
Fixes: openSUSE/obs-build#920


dmach:
This always fixes a scenario when user hits Ctrl-C.
It also does the cleanup on `SIGTERM`, but only if there's a `SIGTERM` handler set in the code, which is the case of `osc/babysitter.py`.

Issue #1280, opened 2023-03-15
Avoid the need to cascade over --build-opt to pass build tool options


rhabacker:
**Does your feature request relate to a problem? Please describe it.**
With https://github.com/openSUSE/obs-build/pull/923, the obs build tool `build` has been given a new command line option `--buildtool-opt=OPTIONS` to pass generator options (for example, `--rpmfcdebug` to track the rpmbuild dependency generator), which can be used by osc as follows
```sh
osc build --build-opt="--buildtool-opt=OPTIONS" ...
```

**Describe the solution you want**.
To avoid the need to cascade over `--build-opt`, it would be nice if osc had a similar `--buildtool-opt` as mentioned at https://github.com/openSUSE/obs-build/pull/923#issuecomment-1468089075.

**Describe the alternatives you considered**.
No alternative available

dmach:
@rhabacker How can I pass multiple options to rpmbuild this way?
I've implemented `--buildtool-opt` as an append option in osc so you can specify it multiple times.

When I concatenate them and use build's `--buildtool-opt` option only once, I get an error:
```
    if opts.buildtool_opt:
        buildtool_opt_str = " ".join(opts.buildtool_opt)
        buildargs += [f"--buildtool-opt={buildtool_opt_str}"]

$ osc-wrapper.py build --buildtool-opt='--foo' --buildtool-opt='--bar'
exec rpmbuild -ba --define '_srcdefattr (-,root,root)' '--foo --bar' ...
rpmbuild: --foo --bar: unknown option
```

Passing the options individually leads to ignoring all but the last option:
```
    if opts.buildtool_opt:
        buildargs += [f"--buildtool-opt={opt}" for opt in opts.buildtool_opt]

$ osc-wrapper.py build --buildtool-opt='--foo' --buildtool-opt='--bar'
exec rpmbuild -ba --define '_srcdefattr (-,root,root)' --bar
rpmbuild: --bar: unknown option
^^^ no --foo option
```

Doesn't obs-build require an additional fix?

rhabacker:
> osc-wrapper.py build --buildtool-opt='--foo' --buildtool-opt='--bar'

`--foo` and `--bar` is not supported,  you need to add a supported option, try the `--rpmfcdebug` option, mentioned above or

```
   osc build --buildtool-opt="--define '_myvar xyz' "
```

> Doesn't obs-build require an additional fix?

[...]

PR #1304, opened 2023-04-26
Add '--buildtool-opt' option passing options to underlying rpmbuild to the 'build' command

dmach:
Fixes: #1280

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -7012,6 +7012,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                   help='define macro X with value Y')
     @cmdln.option('--build-opt', metavar='OPT', action='append',
                   help='pass option OPT to the build command')
+    @cmdln.option('--buildtool-opt', metavar='OPT', action='append',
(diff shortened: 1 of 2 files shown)


osc/commandline.py
```diff
@@ -7012,6 +7012,8 @@ def parse_repoarchdescr(self, args, noinit=False, alternative_project=None, igno
                   help='define macro X with value Y')
     @cmdln.option('--build-opt', metavar='OPT', action='append',
                   help='pass option OPT to the build command')
+    @cmdln.option('--buildtool-opt', metavar='OPT', action='append',
+                  help='pass option OPT to the build command')
```
rhabacker, line 7016: This is currently printed as
```
    --buildtool-opt=OPT
                        pass option OPT to the build command
    --build-opt=OPT     pass option OPT to the build command
```

The help text clashes with  the help from the `--build-opt=OPT`. It should be better `help='pass option OPT to the build tool command'`

dmach: Good catch, thanks!

Issue #477, opened 2018-11-28
build -p <dir> broken logic if dir/repodata/ is there but empty


lethliel:
The following code will assume that it is a repository even if repodata is empty (for whatever reason):

https://github.com/openSUSE/osc/blob/74ed6a5b9a0a5ce95fd33bb8157689ff6659c50b/osc/build.py#L386-L409

I will change it to check for valid data in repodata/ and if no data is there just retrun None

PR #1307, opened 2023-04-27
Fix build.get_repo() to return only directory that contains 'repodata/repomd.xml'

dmach:
Fixes: #477


dmach:
Checked with @lethliel who reported the issue and he thinks the patch should be ok. Merging...

Issue #870, opened 2020-11-23
Commit message is lost when server rejects request


jengelh:
osc-0.171

Imagine accidentally operating on a wrong request number:
```
» osc rq accept 850109
[$EDITOR opens e.g. /tmp/osc-editor3l0m12k_.diff]
[Type message & save]
makeurl: https://api.opensuse.org ['request', '850109'] {'cmd': 'changestate', 'newstate': 'accepted'}
POST https://api.opensuse.org/request/850109?cmd=changestate&newstate=accepted
HTTP Error 403: Forbidden
post_request_no_permission
Request is in review state. You may use the force parameter to ignore this.
```
The commit message is subsequently lost and needs to be typed up again. If the server rejects the request or times out, it should be retained for the next osc call or otherwise saved for reuse.

marcus-h:
On 2020-11-23 00:41:01 -0800, Jan Engelhardt wrote:
> Imagine accidentally operating on a wrong request number:
> ```
> » osc rq accept 850109
> [$EDITOR opens e.g. /tmp/osc-editor3l0m12k_.diff]
> [Type message & save]
> makeurl: https://api.opensuse.org ['request', '850109'] {'cmd': 'changestate', 'newstate': 'accepted'}
> POST https://api.opensuse.org/request/850109?cmd=changestate&newstate=accepted
> HTTP Error 403: Forbidden
> post_request_no_permission
> Request is in review state. You may use the force parameter to ignore this.
> ```
> The commit message is subsequently lost and needs to be typed up again. If the server rejects the request or times out, it should be retained for the next osc call or otherwise saved for reuse.
>
Hmm I'm not really sure what we could do here... The "problem" is that
the request related commands do not require a project or package working
copy. Hence, we cannot store the message in a working copy.

Some ideas (in case of an error):
- store the message in a tempfile and print the filename (probably not
  a good idea because it is not guaranteed that the file still exists
  after osc exits)
- prompt the user for a filename where the message will be stored
- simply print the message to stdout

Honestly, I don't like these ideas... Any suggestions?:)

lethliel:
Perhaps we could just fall back in the editor. So that the user can take care of the message himself?

PR #1314, opened 2023-05-04
Backup edited messages and notify user about them when osc errors out

dmach:
The messages are kept in ~/.cache/osc/edited-messages for 1 day.

Fixes: #870


lethliel:
This looks good to me. We should consider making it package (wc) dependent in the future

PR #1340, opened 2023-06-05
Fix crash when submiting a package which has no _link

adrianschroeter:


dmach:
Changed the code to always initialize `linkinfo`.
Also initializing empty `serviceinfo`.

Issue #1343, opened 2023-06-13
support building from dist-git


dcermak:
**Is your feature request related to a problem? Please describe.**

We are currently experimenting with migrating the build descriptions from the OBS internal version control to git. In the first iteration, it would look like the repositories in https://gitea.opensuse.org/rpm. However, this currently poses a challenge for testing changes in the git repository directly, as there is no easy way how one can build a package out of the git repository (with the `.osc` directory missing).

**Describe the solution you'd like**

Ideally osc would support building out of dist-git repos natively, but as the current setup is not even remotely defined I wouldn't invest too much time there _yet_.

I think for the start it would be enough to support a `--dist-git` flag (or something similar) that would instruct osc to not look for the `.osc` directory. The necessary metadata for the operation would then have to be fetched via other means, e.g. via a `--obs-project` flag, where one would specify the project name whoose repositories & prjconf should be taken for the build.

**Describe alternatives you've considered**

Branching the package, copying the `.osc` directory and then copying it into the git repository. But that adds too many steps for any volunteers to consider.

PR #1344, opened 2023-06-13
Support building directly from git

dmach:
Fixes: #1343


dmach:
The new code integrates nicely, no massive hacks in the existing code.
The only big unresolved thing is hard-coded project that is determined from branch name. We'll fix it once we have the information available in git, gitea, obs or any other place.

Issue #1346, opened 2023-06-14
osc rbl only prints partial log


Vogtinator:
I wanted to read the full log of the last FF build, so I ran

```
osc rbl --last openSUSE:Factory MozillaFirefox standard x86_64
```

The output stopped way too early, then there was a delay and I got the prompt back:

```
[  899s] 12:31.64 /home/abuild/rpmbuild/BUILD/obj/dist/include/nsTArray.h:2951:19:   required from ‘CopyableTArray<E>& CopyableTArray<E>::operator=(const CopyableTArray<E>&) [with E = mozilla::dom::workerinternals::JSSettings::JSGCSetting]â
[  899s] 12:31.64 /home/abuild/rpmbuild/BUILD/obj/dist/include/mozilla/dom/workerinternals/JSSettings.h:24:8:   required from here
[  899s] 12:31.64 /home/abuild/rpmbuild/BUILD/obj/dist/include/nsTArray.h:672:13: warning: â
fabian@fabians-envy:~> 
```

I suspected network issues, but curl behaves properly. What actually happened is that the received data was wrongly "decoded" from UTF-8 into UTF-8. curl printed `5D E2 80 99 0A`, but osc wrote `5D C3 A2 C2 80 C2 99 0A`, which is UTF-8 interpreted as Latin-1 reencoded as UTF-8.

The reason for this is actually already explained in a code comment:

https://github.com/openSUSE/osc/blob/03016a6f7b8182e3fe1eaa3dbbd26f8153c3d905/osc/core.py#L6928-L6930

I could confirm this by making the decoding strict instead with `sys.stdout.write(data.decode(encoding="utf-8", errors="strict"))`:

```
[  899s] 12:31.64 /home/abuild/rpmbuild/BUILD/obj/dist/include/nsTArray.h:2951:19:   required from ‘CopyableTATraceback (most recent call last):
  File "/usr/bin/osc", line 33, in <module>
    sys.exit(load_entry_point('osc==1.1.4', 'console_scripts', 'osc')())
  File "/usr/lib/python3.10/site-packages/osc/babysitter.py", line 236, in main
    sys.exit(run(commandline.OscMainCommand()))
  File "/usr/lib/python3.10/site-packages/osc/babysitter.py", line 69, in run
    prg.main(argv)
  File "/usr/lib/python3.10/site-packages/osc/commandline.py", line 531, in main
    cmd.run(args)
[...]

PR #1347, opened 2023-06-16
Fix printing utf-8 characters to stdout

dmach:
When utf-8 encoded text is transferred over network and read in chunks, if frequently happens that a character gets split between 2 chunks. Decoding such chunks fails on invalid multibyte sequence.

The solution is to forward the bytes to stdout's buffer and make the utf-8 decoding "someone else's problem".

Fixes: #1346


Vogtinator:
Confirmed to fix #1346

PR #1349, opened 2023-07-04
Object wrappers for OBS XML

dmach:


dmach:
Hi,
I'd like to get your feedback on this change.

What's going on:
* wrapping any OBS XML into objects that are generated from OBS RelaxNG schemas
  * I discussed state of the schemas with @adrianschroeter and @dirkmueller, and our conclusion was that we want `rng` as the source of truth
  * spme `xsd` schemas need to be migrated to `rng`
  * some schemas are missing
  * some schemas are a bit off
  * ... but all that can be fixed
* `osc.core` (and other osc code) should not work with XML directly but use the new objects internally. The old code should go away eventually.
* it should replace any custom OBS XML handling code outside osc
* :warning: the code is still under development
* :information_source: check `test_obs_api_*.py` for examples

@lethliel @marcus-h @adrianschroeter @dirkmueller
@crazyscientist - https://github.com/SUSE/osc-tiny
@dcermak - https://github.com/dcermak/py-obs (and possibly also other projects)
@Vogtinator @nilxam - https://github.com/openSUSE/openSUSE-release-tools
@mimi1vx - https://github.com/openSUSE/osc-plugin-qam
@SchoolGuy - https://github.com/openSUSE/sle-prjmgr-tools

What do you think?
Do you consider this a good idea?
If you have any concerns, suggestions of feature requests, please let me know.

Vogtinator:
Sounds like a good idea.

> it should replace any custom OBS XML handling code outside osc

How are the wrappers generated? Are there wrappers for XML documents not used by osc as well?

dmach:
> Sounds like a good idea.
> 
> > it should replace any custom OBS XML handling code outside osc
> How are the wrappers generated?
I have couple python scripts doing that. Extremely ugly code at the moment, but I'm planning to include it in this pull request once I clean it up.

> Are there wrappers for XML documents not used by osc as well?
Not yet, because I wanted to start with the document used by osc.
Throwing anything else in is possible, but I'll do it upon request.
I just don't want to introduce API nobody uses...

crazyscientist:
> What do you think? Do you consider this a good idea?

From the viewpoint of a developer, who needs to handle the XML directly (either to prepare a request or process a response), I certainly would consider this a good idea: Having Python objects means that the IDE can provide helpful information, which the dev does not need to look up manually. :+1: 

> wrapping any OBS XML into objects that are generated from OBS RelaxNG schemas 

"Unfortunately", I decided to use [lxml](https://lxml.de) for [osc-tiny](https://github.com/SUSE/osc-tiny) and let the invoking code deal with objects generated by `lxml.objectify`.

By the way, did you use a tool like "generateDS"?

osc/obs_api/choices.py
mimi1vx, line 1: Shouldn't be here `Enum`'s ?

osc/obs_api/xmlmodel/fields.py
```diff
@@ -0,0 +1,341 @@
+from .validators import ChoicesValidator
+from .validators import TypeValidator
+from .validators import OptionalValidator
+from .xml import ET
+
+
+class Field(property):
```
mimi1vx, line 7: Abstract Class?

osc/obs_api/xmlmodel/fields.py
```diff
@@ -0,0 +1,341 @@
...
+
+        super().__init__(fget=self.get, fset=self.set, fdel=self.delete, doc=help_text)
+
+    def get(self, model):
+        raise NotImplementedError()
+
+    def set(self, model, values):
+        raise NotImplementedError()
+
+    def delete(self, model):
```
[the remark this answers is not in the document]
mimi1vx: CodeQL is right here

mimi1vx:
I think we talked about it on OSC23 :D good thing. In the worst case, I modernize **my** plugin

osc/obs_api/choices.py
```diff
@@ -0,0 +1,78 @@
+BLOCK_MODES: tuple[str] = (
```
dcermak, line 1: Note that `tuple[str]` is a relatively recent syntax, might be Python 3.9+ or 3.10+. Also, I'd agree with @mimi1vx that these should be `StrEnum` instead.

osc/obs_api/xmlmodel/fields.py
```diff
@@ -0,0 +1,341 @@
...
+        else:
+            what = f"{self.get_property_name(model)}"
+
+        for validator in self.validators:
+            for value in values:
+                validator(value, what=what)
+
+
+class ModelField(Field):
+    def __init__(self, name, model_class, optional=False, help_text=None):
```
dcermak, line 214: Whether a field is optional should be obtainable from the type signature of the field. What I mean is that atm there's a bit of information duplication going on:
```Python
class Foo:
    trigger: Optional[str] = xmlmodel.AttributeField(
        "trigger",
        choices=choices.RELEASE_TRIGGERS,
        optional=True,
    )
```
We know that trigger is `optional=True` due to the type. Maybe it would be possible to hijack pydantic's validator syntax for this:
```Python
class Foo:
    @xmlmodel.AttributeField(choices.RELEASE_TRIGGERS)
    trigger: Optional[str]
```

dcermak (commented):
I really like this! This new API would make py-obs more or less completely obsolete (also, the API is more or less the same :wink:), except for the `asyncio` support, but I could live without that.

dmach:
> "Unfortunately", I decided to use [lxml](https://lxml.de) for [osc-tiny](https://github.com/SUSE/osc-tiny) and let the invoking code deal with objects generated by `lxml.objectify`.

`lxml.objectify` is great when you want to avoid dealing with XML directly, but I wanted something more robust.
Also I wanted to generate the classes so I can generate docs from them.
 
> By the way, did you use a tool like "generateDS"?

I was looking for existing tools but I missed this one.
I ended up writing my own in the end (to be published after a massive code cleanup).

SchoolGuy:
I think @gyr is the better person to ask but in any case here are my two cents:

I think the big advantage is that if we are using auto-generated code changes in the XML will now cause compile-time/linter-time issues which would cause us to see breakage before it happens rather than during runtime when the XML is being parsed.

For sure you can make the XML handling code robust enough to consider that the schema has changed but it still doesn't tell you that something is wrong before it is executed.

I am overall for this change. The recursive imports are worrying me a bit but I believe that this is a WIP PR and as such this can be addressed if required at a later point. From my side please go forward with this! Maybe at some point, we can go all the way to protobuf and gRPC to not maintain our own scripting for generating the code for this.

osc/util/models.py
```diff
@@ -390,3 +401,350 @@
...
+        Remove a node that has been fully processed and is now empty.
+        """
+        if len(node) != 0:
+            raise RuntimeError(f"Node {node} contains unprocessed child elements {list(node)}")
+        if node.attrib:
+            raise RuntimeError(f"Node {node} contains unprocessed attributes {node.attrib}")
+        if node.text is not None and node.text.strip():
+            raise RuntimeError(f"Node {node} contains unprocessed text {node.text}")
+        if parent is not None:
+            parent.remove(node)
```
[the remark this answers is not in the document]
dmach: To be honest, I don't see the problem.
Could someone else take a look?

dmach:
So, this is the second (public) iteration of the object wrapper for OBS XML.
It is relatively close to pydantic which gives us a slight chance to move from custom `BaseModel` and `Field` objects to pydantic someday. I decided to avoid pydantic for now is that v2 not widespread yet and there are differences between v1 and v2. Also, the latest version doesn't work on Python 3.6 which is a requirement for osc.

osc/util/models.py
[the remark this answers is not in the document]
dcermak: I don't see a problem either here

osc/util/models.py
```diff
@@ -390,3 +409,362 @@ def dict(self):
...
+        with tempfile.NamedTemporaryFile(mode="w+", encoding="utf-8", prefix="obs_xml_", suffix=".xml") as f:
+            original_data = self.to_string()
+            write_file(f, original_data)
+
+            while True:
+                run_editor(f.name)
+                try:
+                    edited_obj = self.__class__.from_file(f.name)
+                    f.seek(0)
+                    edited_data = f.read()
```
dcermak, line 751: Shouldn't we truncate here too?

dmach: No, because we truncate on `write()` while this performs `read()`

osc/util/models.py
```diff
@@ -390,3 +409,362 @@ def dict(self):
...
+                return value
+
+        if field.origin_type is int:
+            value = int(value)
+            return value
+
+        return value
+
+    @classmethod
+    def _remove_empty_node(cls, parent, node):
```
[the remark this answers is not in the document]
dmach: How about `_remove_processed_node()`?
You need to read the doc texts and comments to fully understand it anyway...

PR #1352, opened 2023-07-11
Update 'service' command to work with git SCM ('factory' branch only for now)

dmach:


dcermak (approved):
Tested it locally, works for me. Thanks a lot!

mcepl:
```
tumbleweed-pkg~/r/osc (git-support-in-service-command *)$ git diff
index 53f78c92..abccfbed 100644
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -7577,7 +7577,7 @@ Please submit there instead, or use --nodevelproject to force direct submission.
 
         try:
             store = osc_store.Store(Path.cwd())
-            store.assert_package()
+            store.assert_is_package()
             project = store.project
             package = store.package
         except oscerr.NoWorkingCopy:
tumbleweed-pkg~/r/osc (git-support-in-service-command *)$ 
```

mcepl:
```
tumbleweed-pkg~/r/osc (git-support-in-service-command *)$ odr
ERROR: Couldn't map git branch 'git-support-in-service-command' to a project
Directory '/home/matej/archiv/knihovna/repos/osc' is not a working copy
Current directory looks like git.
tumbleweed-pkg~/r/osc (git-support-in-service-command *)$ 
```

adrianschroeter:
please note that some service need to have a project and package container set.

So they would not produce the same output with this approach.

IMHO we should not implement osc commands which are producing different output depending on the different checkout methods.

Let's enhance the plain git tooling in build script instead to run services independend of OBS specific configuration.

osc/commandline.py
```diff
@@ -7627,7 +7645,21 @@ def do_service(self, subcmd, opts, *args):
...
+            # we're in a git repo
+            si = Serviceinfo()
+            si.apiurl = apiurl
+            si.project = project
+            si.package = package
+
+            if os.path.isfile("_service"):
+                si.read(ET.parse("_service"))
+
+            return si.execute(dir=".", callmode=mode, singleservice=singleservice)
```
dirkmueller, line 7662: I think this is pretty hacky .. we need to decide where to store the information. I wonder if checking `.osc/_project` into git wouldn't be the easiest solution (works with feature branches, doesn't need osc hacks). 

we could also make a fallback `.obsconfig/_project` which `Package()` falls back to for reading but not for writing.

dirkmueller (changes requested):
lets discuss a sustainable approach on branch name hacking. we could be renaming the "factory" branch to include the necessary information also for example.

adrianschroeter:
On Mittwoch, 12. Juli 2023, 09:47:12 CEST Dirk Mueller wrote:
> @dirkmueller commented on this pull request.
> 
> 
> 
> > +        if store:
> +            # we're in a working copy
> +            p = Package(".")
> +            return p.run_source_services(mode, singleservice, opts.verbose)
> +        else:
> +            # we're in a git repo
> +            si = Serviceinfo()
> +            si.apiurl = apiurl
> +            si.project = project
> +            si.package = package
> +
> +            if os.path.isfile("_service"):
> +                si.read(ET.parse("_service"))
> +
> +            return si.execute(dir=".", callmode=mode, singleservice=singleservice)
> 
> I think this is pretty hacky .. we need to decide where to store the information. I wonder if checking `.osc/_project` into git wouldn't be the easiest solution (works with feature branches, doesn't need osc hacks). 

we want to use the same git repo in many projects. So this can not be done.

The sources should be therefore agnostic to any project/package setup.

osc is the right tool to work in scope of a project/package, but that
is the reason why I suggest _not_ to use osc when working on git directly.

And therefore using our simpler tools instead to run services without the context
of any project/package.

> we could also make a fallback `.obsconfig/_project` which `Package()` falls back to for reading but not for writing. 

-- 

Adrian Schroeter ***@***.***>

SUSE Software Solutions Germany GmbH, Frankenstraße 146, 90461 Nürnberg, Germany
(HRB 36809, AG Nürnberg) Geschäftsführer: Ivo Totev

osc/commandline.py
dcermak: This could lead to discrepancies between git and obs, unless the a `git/osc clone/pull/update` automatically updates this with the settings from OBS.

dcermak:
> lets discuss a sustainable approach on branch name hacking. we could be renaming the "factory" branch to include the necessary information also for example.

you mean something like `factory_devel-languages-python` to encode the devel project in there as well? I must confess I really don't like these overly long branch names :-(

mcepl:
> > lets discuss a sustainable approach on branch name hacking. we could be renaming the "factory" branch to include the necessary information also for example.
> 
> you mean something like `factory_devel-languages-python` to encode the devel project in there as well? I must confess I really don't like these overly long branch names :-(

You surely meant `SUSE_SLE-12-SP3_Update_Products_Teradata_Update_python36_topic`, right?

mcepl:
I think that with https://src.opensuse.org/mcepl_tools/osc/commit/8dbe5a5c8eb7a255e89a672dd08c5cb44f065da7 this could be merged.

osc/commandline.py
```diff
@@ -7571,24 +7571,45 @@ def do_service(self, subcmd, opts, *args):
...
 
-        args = slash_split(args)
-        project = package = singleservice = mode = None
         apiurl = self.get_api_url()
+
         remote_commands = ('remoterun', 'rr', 'merge', 'wait')
 
+        try:
+            store = osc_store.Store(Path.cwd())
+            store.assert_package()
```
mcepl, line 7580: Fix is in https://src.opensuse.org/mcepl_tools/osc/commit/8dbe5a5c8eb7a255e89a672dd08c5cb44f065da7 (`assert_package` -> `assert_is_package`).

If we merge this as it is, then the situation won’t get worse, just not that much better, and it seems we still need both package `.osc/` and the project ones as well. So, unless better option is created (and that will be bigger undertaking), I would accept it as at-least-it-works-somehow.

mcepl (changes requested):

dmach:
I've just force-pushed a brand new iteration of this pull request.
It discards adding duplicate code such as custom handling of `Serviceinfo`.
The commandline do_service() method has been improved to use `Store` or the new `GitStore` class and then both obs scm and git scm use the same code paths.

I'm planning to make a release with this change soon in order to get it to the people for testing.
We still need to figure out how to do branching/forking properly and we shouldn't block the pull request on this.
The feature is experimental and we'll change it later as needed.

mcepl:
Nothing broken, seems like working.

PR #1353, opened 2023-07-11
Add highlighting for "osc diff" and similar commands

asdil12:
image


dmach:
I like the colors, thanks for the patch!

Issue #1355, opened 2023-07-13
Unable to build debbuild packages on Ubuntu (because its type determined wrongly)


isbm:
**Describe the bug**
On Debian/Ubuntu, `debbuild`-based packages (deb from RPM's spec) fails with the following:

     Error: cannot get hdrmd5 for <package>

**Versions**
Even latest/greatest as well

**To Reproduce**
Try to build a debbuild package on Ubuntu

**Expected behavior**
Package is built

The actual reason is because `build.py` takes `pacsuffix` variable as init from `Pac` class, which is determined by [`Buildinfo constructor` here](https://github.com/openSUSE/osc/blob/master/osc/build.py#L110-L120). Of course, since it is `.spec`, then `Buildinfo` class "thinks" we are building an RPM. Since it is RPM and `pacsuffix` is so, it is trying to get `hdrmd5` from a Debian package and crashes.

Possible cleanest solution is to only scream very loud on the console that hdrmd5 is not found, but do not kill the whole build. Another way is to look for "debbuild" inside the project config and then set suffix to `deb` as it should be, but this will tight it up to the debbuild itself and probably this is not the best idea. Since I am not sure what would be the best solution here, maybe @adrianschroeter you could suggest something?

The severity of this bug currently renders the entire `osc` unusable for any debbuilds on Debian/Ubuntu platforms. 😞

isbm:
OK, I took a bit more moment and I found a solution: well, if we are querying a target package, why not just look at that package full filename instead? 😉 

PR is here: https://github.com/openSUSE/osc/pull/1356

PR #1356, opened 2023-07-13
Do not determine queried package by the source package origin

isbm:
The bug report is here: #1355 

It would be better to look at queried package origin directly, instead of assuming its type by the source package. Otherwise, in case we are building `.spec` via `debbuild`, RPM tools will start querying a Debian package and this will fail on Debian platform.

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -1367,7 +1367,7 @@ def main(apiurl, opts, argv):
 
     for i in bi.deps:
         if i.hdrmd5:
-            if not i.name.startswith('container:') and i.pacsuffix != 'rpm':
+            if not i.name.startswith('container:') and not i.fullfilename.endswith(".rpm"):
                 continue
             if i.name.startswith('container:'):
                 hdrmd5 = dgst(i.fullfilename)


isbm:
@adrianschroeter maybe it is a better idea to first gradually fix the MyPy and linter errors and only then add it as a GitHub action check? 😆

dmach:
@isbm could you provide a reproducer I can use to test your patch?
Maybe a link to the package you used should do the job.
I'd like to run the code and see what actually happens in build.py.

isbm:
> @isbm could you provide a reproducer 

Unfortunately, this reproducer in our internal OBS instance. But any debbuild-utilising package on OBS should do, as long as you are testing that on Ubuntu/Debian. The installation package I took from here: https://build.opensuse.org/package/show/openSUSE:Tools/osc

> I can use to test your patch?

The patch is just that: `osc` originally in `Buildinfo` determines the extension of the source package by looking if it is `.spec` or not. If it is (as is always the case with `debbuild` in the first place), it sets extension value to "rpm". When it comes to the area of hdrmd5 (RPM-only stuff), it still hammering `.deb` package with it and therefore fails.

The easiest solution here is to instead assuming all further dependencies are also RPMs, just actually look at them if they are actually RPMs by checking their full filenames.

dmach:
> @adrianschroeter maybe it is a better idea to first gradually fix the MyPy and linter errors and only then add it as a GitHub action check? laughing

I did my best to set linters up reasonably.
Darker/black linter works fine, because it runs in a diff mode.
MyPy is a quite different story, I couldn't make the diff mode working, but I'll take a look if I couldn't improve it.

dmach:
Ok, I managed to make `debbuild` work and tested the patch.
I couldn't come with any better approach since there's really no other metadata we can use besides the `fullfilename`.
Merging.

isbm:
> Ok, I managed to make `debbuild` work and tested the patch. I couldn't come with any better approach since there's really no other metadata we can use besides the `fullfilename`. Merging.

Yeah, I had the same understanding.

What I still do not understand is this: why the _target dependency package_ is assumed by the source type of built package? It makes sense though, because normally if you are building RPMs, then all deps will be also RPMs and so on. However, `debbuild` idea actually is orthogonal to this. What if someone introduces a similar stuff, like "RPMbuild" on Debian or "archbuild" on Slackware etc...

So as an idea to make it more solid, we could also look into magic bytes of the queried target dependency file to firmly know what is it. But that might a bit slow down everything.

Thanks!

PR #1357, opened 2023-07-16
Add zsh completition made by Holger Macht and improve

Thaodan:
Right now the zsh completition is packaged inside the zsh package on
OpenSUSE.
Several distributions either package it or the user has to import it
themselves.

It is better if it is stored here so any can pick it up in there zsh
package and put fixes in here.

### DIFF
--- a/contrib/osc.spec
+++ b/contrib/osc.spec
@@ -10,6 +10,7 @@
 %define completion_dir_bash %{_datadir}/bash-completion/completions
 %define completion_dir_csh %{_sysconfdir}/profile.d
 %define completion_dir_fish %{_datadir}/fish/vendor_completions.d
+%define completion_dir_zsh %{_datadir}/zsh/functions/Completion
 %define osc_plugin_dir %{_prefix}/lib/osc-plugins
 # need to override python_sitelib because it is not set as we would expect on many distros
 %define python_sitelib %(RPM_BUILD_ROOT= %{use_python} -Ic "import sysconfig; print(sysconfig.get_path('purelib'))")
@@ -150,6 +151,7 @@ install -Dm0755 contrib/osc.complete %{buildroot}%{_datadir}/osc/complete
 install -Dm0644 contrib/complete.csh %{buildroot}%{completion_dir_csh}/osc.csh
 install -Dm0644 contrib/complete.sh %{buildroot}%{completion_dir_bash}/osc.sh
 install -Dm0644 contrib/osc.fish %{buildroot}%{completion_dir_fish}/osc.fish
+install -Dm0644 contrib/osc.zsh %{buildroot}%{completion_dir_zsh}/osc.zsh
 
 # install rpm macros
 install -Dm0644 macros.osc %{buildroot}%{_rpmmacrodir}/macros.osc
@@ -191,6 +193,7 @@ install -Dm0644 osc.1 %{buildroot}%{_mandir}/man1/osc.1
 %{completion_dir_bash}/*
 %{completion_dir_csh}/*
 %{completion_dir_fish}/*
+%{completion_dir_zsh}/*
 
 # osc owns the dirs to avoid the "directories not owned by a package" build error
 %dir %{_datadir}/fish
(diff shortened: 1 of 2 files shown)


contrib/osc.zsh
```diff
@@ -0,0 +1,149 @@
...
+# 
+# Use the variables $ZSH_OSC_BUILD_TARGETS_EXTRA and $ZSH_OSC_PROJECTS_EXTRA to
+# extend the list of possible completions in your ~/.zshrc like that:
+#  export OSC_PROJECTS_EXTRA="Base:System Base:shells"
+#
+# version 0.2
+#
+
+OSC_BUILD_TARGETS="openSUSE_13.1 openSUSE_13.2 openSUSE_Tumbleweed openSUSE_Factory SLE_11_SP3 SLE_12"
+OSC_PROJECTS="openSUSE:Factory openSUSE:Tumbleweed openSUSE:13.2 openSUSE:13.1"
```
dmach, line 20: I don't like the hard-coded lists.
On the other hand, I don't see any other reasonable alternative at the moment.
Could you update the values to the active repos and projects at least?

dmach (changes requested):
Are you planning to maintain this completion file?
I'm not interested in maintaining it and I'm actually considering using argcomplete in the future (it's not a priority for the moment, that's why I'm postponing it).

Could you also extend `contrib/osc.spec` to get the completion installed so it works out of the box?

Thaodan:
> Are you planning to maintain this completion file?

I can do so. So far it is maintained in the zsh package, I think here is the better place for doing so, more eyes can see the file.

Thaodan:
> I'm not interested in maintaining it and I'm actually considering using argcomplete in the future (it's not a priority for the moment, that's why I'm postponing it).

Ah if that is possible the file could be replaced. However if I understand correctly a wrapper for these is still needed as argcomplete needs to be activated.

Thaodan:
> Could you also extend `contrib/osc.spec` to get the completion installed so it works out of the box?

I can do that too. BTW why is the spec file in this repo but not the other package descriptions such as PKGBUILD?

dmach:
>  BTW why is the spec file in this repo but not the other package descriptions such as PKGBUILD?

It's because we primarily package osc as rpm and that's also focus of some of the upstream tests.
There's PKGBUILD in [openSUSE:Tools/osc](https://build.opensuse.org/package/show/openSUSE:Tools/osc) for example.

contrib/osc.zsh
dmach: > Sure I can do that. The bash completition uses `~/.osc.projects` instead maybe something like this works but also the path is wrong as it is not inside `$XDG_CACHE_HOME`.

Yes, that needs to be fixed. I've submitted https://github.com/openSUSE/osc/pull/1364

Thaodan: Projects are no longer depended on hardcoded list. Now on targets and arches are left.

Thaodan: Now targets are done.

Thaodan: Architectures are now fetched from the project repositories too. Should be all good now too.

Thaodan:
The competition is now much closer to the bash/tcsh or in some areas further than that.

Some commands don't have competition for the switches they provide thou.

dmach:
I've added a commit that updates spec with installing the completion.

I need more time to test this, because I'm still getting build targets such as `openSUSE_13.1` even when testing in a new container. I'm probably doing something wrong, just need to figure out what it is :)

Thaodan:
> I've added a commit that updates spec with installing the completion.
> 
> I need more time to test this, because I'm still getting build targets such as `openSUSE_13.1` even when testing in a new container. I'm probably doing something wrong, just need to figure out what it is :)

any news?

contrib/osc.spec
```diff
@@ -10,6 +10,7 @@
 %define completion_dir_bash %{_datadir}/bash-completion/completions
 %define completion_dir_csh %{_sysconfdir}/profile.d
 %define completion_dir_fish %{_datadir}/fish/vendor_completions.d
+%define completion_dir_zsh %{_sysconfdir}/zsh_completion.d
```
Thaodan, line 13: I don't think that's the correct path for completions.

dmach (approved):

PR #1359, opened 2023-07-17
Change 'checkout' command to print open requests only when running in an interactive terminal

dmach:


mcepl:
I am afraid it more than just `osc co`, also `osc up`. Otherwise, yes, this would be awesome to have. Let me test it.

tony-vaughan (approved):
Happy with that, makes sense, approved :)

dirkmueller (approved):

PR #1364, opened 2023-07-19
Use XDG locations in completion

dmach:


contrib/osc.complete
```diff
@@ -58,6 +58,22 @@ let count--
 test "${COMMAND_LINE:$last}" = " " && let count++
 unset last
 
+XDG_CACHE_HOME="${XDG_CACHE_HOME:-$HOME/.cache}"
+XDG_CONFIG_HOME="${XDG_CONFIG_HOME:-$HOME/.config}"
+
+for xdg_dir in "${XDG_CACHE_HOME}" "${XDG_CONFIG_HOME}"; do
```
Thaodan, line 64: You can skip line 61 and 62 by writing it like this: `"${XDG_CACHE_HOME:=$HOME/.cache}"`

contrib/osc.complete
```diff
@@ -58,6 +58,22 @@ let count--
...
+    if [ ! -d "${xdg_dir}" ]; then
+        mkdir -p "${xdg_dir}"
+    fi
+done
+
+projects="${XDG_CACHE_HOME}/osc.projects"
```
Thaodan, line 70: Would it make more sense to use a folder here and save the separate cache per alias in that osc folder in `$XDG_CACHE_HOME`?

contrib/osc.complete
dmach: Possibly, but I prefer setting the defaults explicitly like this rather than doing it elsewhere with a side-effect of setting the default.

contrib/osc.complete
dmach: Yes, that makes sense. Let me change it.

Thaodan: OK I will match the zsh competition to do the same.

dmach:
Unfortunately `-A/--apiurl` option do not seem to have the expected impact on completion and also the completion doesn't store cached data per apiurl. Fixing this is not my current priority, maybe moving to argcomplete would be easier after all.
Merging this as is.

Issue #1365, opened 2023-07-19
osc request list -i should have an option to see complete logs


mcepl:
**Is your feature request related to a problem? Please describe.**
`osc request list -i -m ok --involved-projects -s new` is absolutely awesome, but when going through open review requests I would sometime need to see complete log, not only the filtered output of `rpmlint`.

**Describe the solution you'd like**
Add another option to “the command toolbar” extermely similar to `rpm(li)ntlog` but showing the output of `osc remotebuildlog` in `$PAGER`.

**Describe alternatives you've considered**
I have to switch to the other terminal and run `osc rbl` manually (and find the exact project and name of the package to use) or use web browser, which defeats the purpose of the command.

PR #1367, opened 2023-07-20
Add an interactive option to display build log in 'request list -i' command

dmach:
Also show rpmlint logs in pager.

Fixes: #1365


mcepl (approved):
Works for me just fine.

PR #1370, opened 2023-07-26
Enhance '--force' parameter description

ncounter:
Originated by [this bug](https://github.com/openSUSE/open-build-service/issues/3687) (which is now [fixed](https://github.com/openSUSE/open-build-service/pull/14737)) we realized that '--force' parameter does ignore devel-package dependencies. This PR is to document it to make it clear while using it.


dmach:
Thanks for the patch. Merging.

PR #1376, opened 2023-08-01
Fix less pager by adding '-R' to LESS env

dmach:


osc/core.py
```diff
@@ -4401,11 +4401,21 @@ def run_pager(message, tmp_suffix=''):
...
+
+        env = os.environ.copy()
+
         pager = os.getenv("PAGER", default="").strip()
         pager = pager or get_default_pager()
+
+        # LESS env is not always set and we need -R to display escape sequences properly
+        less_opts = os.getenv("LESS", default="")
+        if "-R" not in less_opts:
+            less_opts += " -R"
```
jengelh, line 4413: When `LESS="-MRi"`, this would cause the condition to be false positive.
But that's not tragic, you could just unconditionally add `-R` at all times.

Unrelated to that; what if $PAGER isn't `less` to begin with?

Issue #1371, opened 2023-07-27
usage without a config file


dcermak:
**Is your feature request related to a problem? Please describe.**

osc is a bit inconvenient to use as a library because it will not launch without a configuration file. While this is not an issue on developer machines, it fails in a CI environment without an `.config/osc/osrc`. One then has to resort to additional steps like:
```yaml
      - name: create a oscrc
        run: |
          mkdir -p ~/.config/osc/
          echo "[general]" > ~/.config/osc/oscrc
          echo "apiurl = https://api.opensuse.org" >> ~/.config/osc/oscrc
          echo "[https://api.opensuse.org]" >> ~/.config/osc/oscrc
          echo "user = MY_USERNAME" >> ~/.config/osc/oscrc
          echo "pass = ${{ secrets.OSC_PASSWORD }}" >> ~/.config/osc/oscrc
```

**Describe the solution you'd like**

Preferably a `Client` or `Osc` or `Configuration` class, that takes the necessary configs from `.oscrc` as a parameter:
```Python
@dataclass
class Configuration:
    username: str
    password: str
    apiurl: str = "https://api.opensuse.org"
```

This sadly does not remove the need for having your password in the CI, but at least the requirement is now more obvious and one requires only to write python.

**Describe alternatives you've considered**

The yaml file above or using `py-obs`

**Additional context**
Example failure in a CI: https://github.com/OSInside/kiwi-functional-tests/actions/runs/5673751891/job/15375908319

PR #1377, opened 2023-08-07
Add conf.configure_apiurl() for configuring osc without reading oscrc

dmach:
Fixes: #1371

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -36,6 +36,7 @@ The configuration dictionary could look like this:
 """
 
 
+import copy
 import errno
 import getpass
 import os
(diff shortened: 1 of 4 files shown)


osc/conf.py
```diff
@@ -971,6 +972,65 @@ def get_config(override_conffile=None,
...
+        typ = data.get("type")
+        if typ == "str":
+            assert isinstance(value, str)
+        elif typ == "bool":
+            assert isinstance(value, bool)
+        elif typ == "int":
+            assert isinstance(value, int)
+        elif typ == "string-list":
+            assert isinstance(value, (str, list, tuple))
+            if isinstance(value, str):
```
dcermak, line 1019: Note that assertions are only run under certain conditions.

dcermak (commented):
Looks good to me in principle, but I deem this a bit too clumsy to use in practice, especially as it requires reading the code to understand what is actually going on.

dmach:
#1387 implements this way better.

Issue #1388, opened 2023-08-22
osc 1.3.0 and kernel scripts/osc_wrapper


tonyjonespdx:
Appears to be an issue using osc 1.3.0 from kernel scripts/osc_wrapper

$ scripts/osc_wrapper build --ibs kernel-source/kernel-default.spec 
osc -A https://api.suse.de build --no-service --local-package --alternative-project=Devel:Kernel:SLE12-SP5 --disable-debuginfo --no-checks --extra-pkgs=-brp-check-suse --extra-pkgs=-post-build-checks --define ignore_kabi_badness 0 --define klp_symbols 1 standard kernel-source/kernel-default.spec
WARNING: Using EXPERIMENTAL support for git scm. The functionality may change or disappear without a prior notice!
Traceback (most recent call last):
  File "/usr/bin/osc", line 33, in <module>
    sys.exit(load_entry_point('osc==1.3.0', 'console_scripts', 'osc')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/osc/babysitter.py", line 234, in main
    sys.exit(run(commandline.OscMainCommand()))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/osc/babysitter.py", line 69, in run
    prg.main(argv)
  File "/usr/lib/python3.11/site-packages/osc/commandline.py", line 545, in main
    cmd.run(args)
  File "/usr/lib/python3.11/site-packages/osc/commandline.py", line 241, in run
    cmd.run(args)
  File "/usr/lib/python3.11/site-packages/osc/commandline.py", line 489, in run
    self.func(args.command, args, *args.positional_args)
  File "/usr/lib/python3.11/site-packages/osc/commandline.py", line 7235, in do_build
    if opts.alternative_project == store.project:
                                   ^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/osc/git_scm/store.py", line 75, in project
    raise RuntimeError(f"Couldn't map git branch '{branch}' to a project")
RuntimeError: Couldn't map git branch 'SLE12-SP5' to a project

PR #1389, opened 2023-08-23
Fix 'build' command when '--alternative-project' option is specified …

dmach:
…and the git branch cannot be mapped to a project

Fixes: #1388


dmach:
I've verified that this patch fixes the original problem, merging.

PR #1401, opened 2023-09-04
Add '--just-print-buildroot' option to print build root path and exit to 'build' command

dmach:


darix:
@marcus-h JFYI this is a port of the print buildroot plugin that you did for me a while ago and which now broke with the rework of osc.

Issue #1158, opened 2022-10-05
OSC pins wrong certificate causing invalid certificate errors


Firstyear:
```
WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!
IT IS POSSIBLE THAT SOMEONE IS DOING SOMETHING NASTY!
offending certificate is at '/home/886b45ce-98b2-4d79-9069-c9ff25bc0232/.config/osc/trusted-certs/api.suse.de_443.pem'
Certificate Verification Error: remote host identification has changed
```

This happens even 90~ish days, on whatever the api.suse.de refresh time is. This appears to be because OSC is incorrectly trusting the *leaf* certificate rather than the *intermediate*, so when the certificate is rotated it causes this validation error. 

This is bad because it trains our users that certificate errors are to be ignored, which allows trivial MITM attacks to be carried out. This could be considered a security issue due to this. 

@mgerstner @msmeissn

msmeissn:
hmm, most people just import the SUSE SSL root CA ( from ca.suse.de ) instead of pinning.

But osc could pin both the intermediate and the leaf perhaps.

dmach:
@Firstyear which osc version are you using? 0.x or 1.x?

dmach:
I have discussed this with OBS people and we concluded that we don't want to support any kind if intermediate certificate pinning.
Self-signed certs can be explicitly trusted by the user, but everything else is supposed to be solved by installing trusted cacerts on the system (e.g. by installing `ca-certificates` or `ca-certificates-suse` package).

Please note that the latest osc-1.x doesn't allow trusting such certs and supports trusting self-signed certs only.

Firstyear:
Maybe osc should ... depend on ca-certificates-suse then. That way we don't have this janky situation occuring.

Firstyear:
@dmach the ca-certificates-suse package doesn't exist:


```

S  | Name                             | Summary                                               | Type
---+----------------------------------+-------------------------------------------------------+-----------
   | ca-certificates                  | Utilities for system wide CA certificate installation | srcpackage
i+ | ca-certificates                  | Utilities for system wide CA certificate installation | package
   | ca-certificates-cacert           | CAcert root certificates                              | srcpackage
   | ca-certificates-cacert           | CAcert root certificates                              | package
   | ca-certificates-mozilla          | CA certificates for OpenSSL                           | srcpackage
i+ | ca-certificates-mozilla          | CA certificates for OpenSSL                           | package
[...]

PR #1403, opened 2023-09-08
Fix api_host_options for custom CAs

ziQ:
Currently, osc seems to only support custom CAs if they are installed system-wide.

This PR adds an `api_host_option`, which allows custom CAs to be defined in `oscrc` API blocks. This is supposed to address concerns like #1158.

@dmach @Firstyear @mgerstner @msmeissn


dmach:
I actually bumped into the same thing few days ago.
I'm rewriting `osc.conf` module and I realized that per-host ca certs are not handled at all.
The support got lost during migration from m2crypto to urllib3 and nobody noticed until now.
 
This is related part of my osc.conf change: https://github.com/openSUSE/osc/pull/1387/files#diff-f40a3299d1d55f86eac1e22a9b88cf4050da62b5b72fb6cfafd4a57e2a72e0d4R412 (you need to load diff for conf.py to make the link working)

I'm planning to fix this in the next iteration of my pull request.

dmach:
@ziQ could you test if the `cafile` and `capath` options work as expected in https://github.com/openSUSE/osc/pull/1387 ?

PR #1416, opened 2023-09-28
add support for keep_packages_locked on request revoke

adrianschroeter:
jsc#OBS-134


dmach:
@adrianschroeter
I've pushed a commit that suggests using `--keep-packages-locked` along with the `request revoke` command.
Feel free to rephrase it or squash it to your commit.

This also makes me think - don't we want `--keep-packages-locked` on by default for certain revoke actions in order to have sane defaults? It sounds to me that we (nearly) always want to keep the packages locked when revoking release requests.

Issue #1419, opened 2023-10-02
Getting Server returned an error: HTTP Error 401: Unauthorized


mcepl:
**Describe the bug**
With any command against OBS (https://build.opensuse.org/) which requires authorization I get the error in %description%. E.g.,

```
$ osc ls devel:languages:python
Server returned an error: HTTP Error 401: Unauthorized
$
```

The same osc with the same credentials works perfectly well with IBS (https://build.suse.de/).

**Versions**
- Operating system: `openSUSE Tumbleweed` (running in distrobox inside of MicroOS)
- Packages: `osc-1.3.1-1.1.noarch` (but the same with `osc` from the git master).

**To Reproduce**
1. Run any command requiring authorization
2. See error

**Expected behavior**
Expected output

**Additional context**
Output of `osc config --dump`:

```
[general]
__name__ = general
check_for_request_on_action = 1
exclude_glob = .osc CVS .svn .* _linkerror *~ #*# *.orig *.bak *.changes.vctmp.* osc-build-log*.txt .gitignore
extra-pkgs = gdb vim
use_keyring = 1

[https://api.opensuse.org]
aliases = obs
credentials_mgr_class = osc.credentials.KeyringCredentialsManager:keyring_pass.PasswordStoreBackend
trusted_prj = CentOS:* openSUSE:* devel:* SUSE:* Education GNOME:* science editors Cloud:OpenStack:* M17N* Base:System security:* multimedia:libs server:* X11:* Fedora:* network:* home:mcepl:*
user = mcepl

[https://api.suse.de]
aliases = ibs
credentials_mgr_class = osc.credentials.KeyringCredentialsManager:keyring_pass.PasswordStoreBackend
email = mcepl@suse.com
sshkey = id_ed25519
trusted_prj = SUSE:* openSUSE.org:*
user = mcepl
```

Also, when playing with pdb, I think got to the point where `self.password` in https://github.com/openSUSE/osc/blob/master/osc/connection.py#L559 was falsy (not remember whether it was empty or `None`), and I haven’t analysed where it is supposed to be set and whether `BasicAuthHandler` was the one I was supposed to use.

Also, when running `keyring api.opensuse.org mcepl` I get the correct answer, so I suppose whole `keyring` mechanism is working correctly.

PR #1420, opened 2023-10-03
Fix loading password from keyring

dmach:
Fixes: #1419


mcepl (approved):
Yup, fixes the problem for me. Thank you!

PR #1424, opened 2023-10-06
Use typeguard during tests and in osc-wrapper.py

dmach:


contrib/osc.spec
```diff
@@ -30,6 +30,13 @@
 %bcond_with fdupes
 %endif
 
+# use typeguard during build on distros where typeguard is available
+%if (0%{?suse_version} > 1500 || 0%{?fedora} >= 37)
```
dcermak, line 34: We don't have to care about Fedora 36 and older
```suggestion
%if (0%{?suse_version} > 1500 || 0%{?fedora})
```

PR #1438, opened 2023-10-19
Allow undefined fields in Options and HostOptions

dmach:
Plugins seem to be using oscrc and osc.conf.config to store their config options. All fields that are not known to osc are now stored in the 'extra_fields' dictionary and handled in __getitem__() and __setitem__() as they were regular fields. Such values are not checked for their types and the dictionary simply holds strings obtained from oscrc or anything the plugins set through the python API.


kraih:
All legal reviews in OBS and IBS are currently broken because of this.

Issue #1439, opened 2023-10-22
KeyError: 'apiurl_aliases'


andreas-schwab:
$ osc config https://api.opensuse.org --delete aliases
Traceback (most recent call last):
  File "/usr/bin/osc", line 11, in <module>
    load_entry_point('osc==1.4.3', 'console_scripts', 'osc')()
  File "/usr/lib/python3.6/site-packages/osc/babysitter.py", line 226, in main
    sys.exit(run(commandline.OscMainCommand()))
  File "/usr/lib/python3.6/site-packages/osc/babysitter.py", line 69, in run
    prg.main(argv)
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 545, in main
    cmd.run(args)
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 241, in run
    cmd.run(args)
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 489, in run
    self.func(args.command, args, *args.positional_args)
  File "/usr/lib/python3.6/site-packages/osc/commandline.py", line 9863, in do_config
    opt, newval = conf.config_set_option(section, opt, ' '.join(val), delete=opts.delete, update=True, creds_mgr_descr=creds_mgr_descr)
  File "/usr/lib/python3.6/site-packages/osc/conf.py", line 1620, in config_set_option
    section = config['apiurl_aliases'].get(section, section)
  File "/usr/lib/python3.6/site-packages/osc/conf.py", line 148, in __getitem__
    return self.extra_fields[name]
KeyError: 'apiurl_aliases'

jengelh:
Also happens adding aliases:

```
12:06 a4:../telephony/linphone $ osc build
Building linphone.spec for openSUSE_Tumbleweed/x86_64
[...]
The build root needs packages from project 'network:telephony'.
Note that malicious packages can compromise the build result or even your system.
Would you like to ...
0 - quit (default)
1 - always trust packages from 'network:telephony'
2 - trust packages just this time
? 1
adding 'network:telephony' to oscrc: ['https://api.opensuse.org']['trusted_prj']
Traceback (most recent call last):
  File "/usr/bin/osc", line 33, in <module>
    sys.exit(load_entry_point('osc==1.4.3', 'console_scripts', 'osc')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1440, opened 2023-10-23
Fix apiurl_aliases handling in OscOptions.__getitem__

dmach:
Fixes: #1439


pallaswept:
Tested this and it fixes the issue for me

PR #1441, opened 2023-10-24
fix autocompletion for new locations

jirislaby:
When complete is installed in `/usr/share` or `/usr/lib64`, the last test command (`/usr/lib`) fails and causes whole `osc` autocompletion to fail.

Fix it by adding a `return` to the former two -- if they succeed.


jirislaby:
Downstream report: https://bugzilla.suse.com/show_bug.cgi?id=1216531

PR #1455, opened 2023-12-01
docs: Update models to avoid including lazy defaults in the rendered man pages

dmach:
Lazy defaults may return different results under different circumstances and we always want man pages rendered consistently.


bmwiedemann (approved):
This fixes the variations in `oscrc.5`

Issue #1453, opened 2023-11-27
Take environment variable for username and password for the initial run


AvengerMoJo:
When using osc in a new container e.g. CI in gitea action runner. 
It doesn't provide an interactive way to provide a username and password. Instead, it allows a secret to be created in the environment variable. Since it doesn't make sense to prepopulate the user credentials in the container or a public repo either. 
The possible solution is to either create an interaction script like expect to grab the secret environment variable to use osc, or better the osc cli can take a default username password environment variable before generating the oscrc configuration. 

[The example script of the workaround](https://src.opensuse.org/alexlau/bc-docker-action/src/branch/factory/osc_checkout)

dmach:
Fixed in #1463

PR #1456, opened 2023-12-04
Add env var as credential input in cli

AvengerMoJo:
#1453 

Taking env var as consideration when the osc is being called the first time. 
This allows automation in the workflow without depending on an interactive session. 

The implementation is simply checking if OSC_USERNAME, OSC_PASSWORD or OSC_CREDENTIAL 
is present and prioritized over the interactive input. 
README is also updated accordingly.

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -1999,10 +1999,14 @@ def interactive_config_setup(conffile, apiurl, initial=True):
         print()
 
     apiurl_no_scheme = urlsplit(apiurl)[1] or apiurl
-    user_prompt = f"Username [{apiurl_no_scheme}]: "
-    user = raw_input(user_prompt)
-    pass_prompt = f"Password [{user}@{apiurl_no_scheme}]: "
-    passwd = getpass.getpass(pass_prompt)
+    user = os.environ.get("OSC_USERNAME", None)
+    if not user:
+        user_prompt = f"Username [{apiurl_no_scheme}]: "
+        user = raw_input(user_prompt)
+    passwd = os.environ.get("OSC_PASSWORD", None)
+    if not passwd:
+        pass_prompt = f"Password [{user}@{apiurl_no_scheme}]: "
+        passwd = getpass.getpass(pass_prompt)
     creds_mgr_descr = select_credentials_manager_descr()
     if initial:
         config = {'user': user, 'pass': passwd}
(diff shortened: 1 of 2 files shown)


osc/conf.py
```diff
@@ -1995,10 +1995,16 @@ def interactive_config_setup(conffile, apiurl, initial=True):
         print()
 
     apiurl_no_scheme = urlsplit(apiurl)[1] or apiurl
-    user_prompt = f"Username [{apiurl_no_scheme}]: "
-    user = raw_input(user_prompt)
-    pass_prompt = f"Password [{user}@{apiurl_no_scheme}]: "
-    passwd = getpass.getpass(pass_prompt)
+    if os.environ.get("OSC_USERNAME"):
+        user = os.environ.get("OSC_USERNAME")
+    else:
```
dcermak, line 2000:
```suggestion
    user = os.environ.get("OSC_USERNAME", None):
    if not user:
```

osc/conf.py
```diff
@@ -1995,10 +1995,16 @@ def interactive_config_setup(conffile, apiurl, initial=True):
...
+        user_prompt = f"Username [{apiurl_no_scheme}]: "
+        user = raw_input(user_prompt)
+    if os.environ.get("OSC_PASSWORD"):
+        passwd = os.environ.get("OSC_PASSWORD")
+    else:
```
dcermak, line 2005:
```suggestion
    passwd = os.environ.get("OSC_PASSWORD", None):
    if not passwd:
```

dmach:
@AvengerMoJo I finally got to this pull request and I really like the idea of setting credentials through the env vars.
Your patch would work just fine in the containers, but then I realized that we could extend this so you could use the variables to override osc settings at any time. It's also in sync with the current implementation of `osc.conf.get_config()` that allows you overriding settings via `OSC_APIURL=...`, `OSC_BUILD_JOBS=...` etc.
Hmm, I should mention that in the [docs](https://opensuse-commander.readthedocs.io/en/latest/oscrc.html). It's mentioned in the [library docs](https://opensuse-commander.readthedocs.io/en/latest/api/osc.conf.html), but that's not enough.

I ended up implementing this: #1463.
I'm sorry I did not post any feedback to this PR earlier, but after some prototyping I realized I'm writing something completely else that is competing with your PR rather than building on top of it and I decided to finish it and ask for your opinion.
I also had to fix some issues I caused in the previous commits first: #1462

Could you test if #1463 works for you?
Please note that setting `OSC_CONFIG=` is essential to avoid creating `oscrc`.

AvengerMoJo:
I am pulling your branch to build a customized docker image to test it against src.o.o in gitea and let you know soon .

AvengerMoJo:
#1463 should replace this PR and the function is working in the cli using OSC_CONFIG= , test with gitea action container with the osc patch https://src.opensuse.org/alexlau/bash-custom-action/actions/runs/9

PR #1460, opened 2023-12-18
Forward --ccache-maxsize CLI arg to obs-build

dcermak:
Depends on https://github.com/openSUSE/obs-build/pull/970


dmach:
The related PR got closed, it makes no sense to keep this one open.

PR #1463, opened 2024-01-04
Credentials from env

dmach:


AvengerMoJo (commented):
Using this patch in https://src.opensuse.org/alexlau/bash-custom-action/actions/runs/9 container and it is working. Thanks

Issue #1468, opened 2024-01-13
Cannot obtain linkdiff


hramrach:
**Describe the bug**
Cannot obtain linkdiff

**Versions**
- Operating system: openSUSE Leap 15.5
- Packages: osc-0.182.1-150100.3.35.1.noarch

**To Reproduce**
```
osc co home:michals/sensors.5
cd home:michals/sensors.5
osc linkdiff
```
```
diff working copy against last committed version

Server returned an error: HTTP Error 400: home michals/sensors.5/f9e18b2d22124e93b78e0ad62c16a51c  not in repository. Either not existing or misconfigured server setting for '$nosharedtrees' setting in BSConfig.pm
home:michals/sensors.5/f9e18b2d22124e93b78e0ad62c16a51c: not in repository. Either not existing or misconfigured server setting for '$nosharedtrees' setting in BSConfig.pm
```

**Expected behavior**
There is always a diff, at least when OBS itself can expand the packages.

dmach:
@hramrach could you check if the patch in the pull request fixes your problem and displays what you expect?

hramrach:
Seems to work, thanks

Issue #1437, opened 2023-10-18
linkdiff sometimes fails


hramrach:
**Describe the bug**
osc co home:michals/ipxe-internal/
cd home:michals/ipxe-internal/
home:michals/ipxe-internal> osc linkdiff
diff working copy against last committed version

Server returned an error: HTTP Error 400: home michals/ipxe-internal/1455fd057ee3971b66f343cd7c2ae099  not in repository. Either not existing or misconfigured server setting for '$nosharedtrees' setting in BSConfig.pm
home:michals/ipxe-internal/1455fd057ee3971b66f343cd7c2ae099: not in repository. Either not existing or misconfigured server setting for '$nosharedtrees' setting in BSConfig.pm

**Versions**
- Operating system: Leap 15.5
- Packages: osc-0.182.1-150100.3.35.1.noarch

**To Reproduce**
run linkdiff for the above package

**Expected behavior**
shows diff as the web UI does

PR #1470, opened 2024-01-16
Fix 'linkdiff' command error 400: prj/pac/md5 not in repository

dmach:
Fixes: #1468
Fixes: #1437


hramrach:
Now linkdiff of branched packages also works 🥇

Issue #1481, opened 2024-02-12
osc doesn't honor XDG_CONFIG_HOME


Thaodan:
**Describe the bug**
Even when `$XDG_CONFIG_HOME` is set. is set osc still writes to `$HOME/.config/osc`. I noticed it creates `$HOME/.config/osc/trusted-certs`.
I'm not sure if the trusted certs folder belongs in the folder as opposed to `$XDG_STATE_HOME`.

In addition it doesn't save options such as the permanent saving of trusted project to oscrc when `$XDG_CONFIG_HOME` is set.

**Versions**
- Operating system: openSUSE Tumbleweed
- Packages: osc-1.6.0-403.1.noarch (issue exists at least since about a year ago).

**To Reproduce**
0. export XDG_CONFIG_HOME=$PWD/foobar
1. Run command `osc -A <apiurl> ...`
2. See error

**Expected behavior**
Osc should honor XDG variables

PR #1485, opened 2024-02-14
Honor XDG_CONFIG_HOME and XDG_CACHE_HOME env vars

dmach:
Fixes: #1481

### DIFF
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -1606,14 +1606,14 @@ def get_configParser(conffile=None, force_read=False):
 
 def write_config(fname, cp):
     """write new configfile in a safe way"""
-    if os.path.exists(fname) and not os.path.isfile(fname):
-        # only write to a regular file
-        return
-
     # config file is behind a symlink
     # resolve the symlink and continue writing the config as usual
     if os.path.islink(fname):
-        fname = os.readlink(fname)
+        fname = os.path.realpath(fname)
+
+    if os.path.exists(fname) and not os.path.isfile(fname):
+        # only write to a regular file
+        return
 
     # create directories to the config file (if they don't exist already)
     fdir = os.path.dirname(fname)
--- a/osc/conf.py
+++ b/osc/conf.py
@@ -65,6 +65,7 @@ from urllib.parse import urlsplit
 from . import credentials
 from . import OscConfigParser
 from . import oscerr
+from .output import tty
 from .util import xdg
 from .util.helper import raw_input
 from .util.models import *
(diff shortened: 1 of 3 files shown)


dmach:
@Thaodan could you test if the patch works for you?

Thaodan:
Daniel Mach ***@***.***> writes:

> @Thaodan could you test if the patch works for you?

The patch works for me, the trusted certs folder is created in
XDG_CONFIG_HOME now.
However I'm not sure if that's the right place for the folder as it's
more part of the state as opposed to config.

When accepting projects as trusted the change isn't written to oscrc.
n

dmach:
> When accepting projects as trusted the change isn't written to oscrc.

The original report was about trusted **certs**.
Now you're writing about trusted **projects**. Is this what you really meant?
I just want to make clear that you didn't write it by an accident and that there might be an additional bug.

Thaodan:
> > When accepting projects as trusted the change isn't written to oscrc.
> 
> The original report was about trusted **certs**. Now you're writing about trusted **projects**. Is this what you really meant? I just want to make clear that you didn't write it by an accident and that there might be an additional bug.

The original report was also about the trusted projects:

> In addition it doesn't save options such as the permanent saving of trusted project to oscrc when $XDG_CONFIG_HOME is set.

dmach:
> The original report was also about the trusted projects:

I did not think that this is even possible, but clearly is.
I've updated this PR with a new commit that warns you if you have `~/.oscrc` on disk, because that ignores `XDG_CONFIG_HOME`  and `~/.config/osc/oscrc`.
Isn't that your case?
It would use the trusted projects from the existing `~/.oscrc` and storing new values there.

If you don't have `~/.oscrc` on disk, then I'm afraid I have no idea what's wrong with the trusted projects, everything worked fine for me when I tested this.

Thaodan:
Daniel Mach ***@***.***> writes:

>> The original report was also about the trusted projects:
>
> I did not think that this is even possible, but clearly is.
> I've updated this PR with a new commit that warns you if you have
> `~/.oscrc` on disk, because that ignores `XDG_CONFIG_HOME` and
> `~/.config/osc/oscrc`.
> Isn't that your case?
> It would use the trusted projects from the existing `~/.oscrc` and storing new values there.
>
> If you don't have `~/.oscrc` on disk, then I'm afraid I have no idea
> what's wrong with the trusted projects, everything worked fine for me
> when I tested this.

No my case is to have $XDG_CONFIG_HOME set to $HOME/.local/etc.
However I forgot to mention that oscrc is symlink to
$HOME/.local/private/etc/osc/oscrc.
The purpose of this is to separate my osc config as it can contain
confidential data.

dmach:
The symlinks seem to be the root cause after all.
I was able to reproduce not storing the trusted projects into a symlinked config, adding `osc.path.realpath()` fixed that for me.

Thaodan:
Thank you for fixing the bug.

PR #1489, opened 2024-02-17
Decode character references in commit log

tacerus:
Correctly represent special characters in the text based log output for better readability.

Example before:

```
[botdel] Package has had build problems for &gt;= 6 weeks
```

Example after:

```
[botdel] Package has had build problems for >= 6 weeks
```

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -32,6 +32,7 @@ import textwrap
 import time
 import warnings
 from functools import cmp_to_key, total_ordering
+from html import unescape
 from http.client import IncompleteRead
 from io import StringIO
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


tacerus:
Linter in https://github.com/openSUSE/osc/actions/runs/7939010125/job/21678562033 wanted me to change the syntax of the existing lines as wel so I added that to the commit.

dmach:
I'm unable to reproduce it - I tried writing the characters in the webui and view them in osc.
Could you write me a reproducer?
Also, could you check if the text isn't broken in webui as well? That would mean that unescaping is done correctly and we only display what's in the source.

tacerus:
You're right, I now tested it again as well and maybe it is not a problem with `osc` afterall.

Reason for the patch was this:

```
~> osc log -D openSUSE:Factory kopete |head -4
----------------------------------------------------------------------------
r134 | anag+factory | 2024-02-08 18:02:03 | 8b84f116348051059ea6ac6b364667a7 | 23.08.4 | rq1144747

[botdel] Package has had build problems for &gt;= 6 weeks
```

I'm not sure how to view deleted logs in the web UI but it's possible that character issue was already introduced by the software submitting that commit/request.

dmach:
Since there doesn't seem to be any osc change needed, let me close this pull request.

tacerus:
Thanks, sorry for the noise, I now submitted this in the right place.
https://github.com/openSUSE/openSUSE-release-tools/pull/3059

PR #1497, opened 2024-02-23
Allow `osc rpmlint` to infer prj/pkg from CWD.

badshah400:
Makes `osc rpmlint` infer the project and package when working from a locally checked out package dir, if two or fewer arguments are passed. All four arguments are still required when called from a location that is not an osc package working copy.

This makes the behaviour of `osc rpmlint` similar to `osc buildlogtail`.


dmach:
Looks good to me, merging.

badshah400:
Thanks for the quick response and merge.

Issue #1492, opened 2024-02-20
osc buildinfo -d produces different output in a package checkout despite calling the same API route


dcermak:
**Describe the bug**

`osc buildinfo -d` creates different output when called with a full project + package arguments than without when called in a package checkout directory, despite the debug output suggesting that the same api call is used.


**Versions**
- Operating system: `Fedora Linux 39 (Thirty Nine)`
- Packages: `osc-1.6.0-400.1.1.fc39.noarch`

**To Reproduce**
1. `osc co devel:BCI:SLCS:next`
2. `cd devel:BCI:SLCS:next/python-3.12-image`
3. call `osc buildinfo -d` just with the repo + arch: 
```
❯ osc --debug buildinfo -d  containerfile x86_64 > buildinfo_few                                                                                                                                                                                                              
DEBUG: Config option '[https://api.opensuse.org]/keyring' doesn't map to any HostOptions field                                                                                                                                                                                               
DEBUG: Config option '[https://api.suse.de]/keyring' doesn't map to any HostOptions field                                                                                                                                                                                                    
DEBUG: Config option '[general]/gitea-api-token' doesn't map to any Options field                                                                                                                                                                                                            
hostarch:  x86_64                                                                                                                                                                                                                                                                            
arg_arch:  x86_64                                                                                                                                                                                                                                                                            
arg_repository:  containerfile                                                                                                                                                                                                                                                               
arg_descr:  None                                                                                                                                                                                                                                                                             
DEBUG: makeurl: https://api.opensuse.org ['source', 'devel:BCI:SLCS:next', '_meta'] []                                                                                                                                                                                                       
DEBUG: makeurl: https://api.opensuse.org ['build', 'devel:BCI:SLCS:next', 'containerfile', 'x86_64', 'python-3.12-image', '_buildinfo'] ['debug=1']                                                                                                                                          
```
4. call `osc buildinfo -d` with the prj + pkg + repo + arch:
```
❯ osc --debug buildinfo -d devel:BCI:SLCS:next python-3.12-image containerfile x86_64 > buildinfo_full                                                                                                                                                                        
DEBUG: Config option '[https://api.opensuse.org]/keyring' doesn't map to any HostOptions field                                                                                                                                                                                               
DEBUG: Config option '[https://api.suse.de]/keyring' doesn't map to any HostOptions field                                                                                                                                                                                                    
DEBUG: Config option '[general]/gitea-api-token' doesn't map to any Options field                                                                                                                                                                                                            
[...]

PR #1500, opened 2024-02-28
Fix 'buildinfo' command

dmach:
Fixes: #1492


dcermak:
I have retried the reproducer for #1492 and there's still differences between the outputs, albeit fewer and also some of them are whitespace differences

dcermak (approved):

dmach:
I made one additional change: buildconfig gets downloaded only when needed

Issue #1511, opened 2024-03-08
`osc build --vm-type=qemu` wants root privileges


wfrisch:
**Describe the bug**
`osc build --vm-type=qemu` still asks for the root password, even though it shouldn't be necessary (see `--vm-type=kvm`).

**Versions**
osc-1.6.1

**To Reproduce**
1. Install the `qemu-arm` package
2. `osc co openSUSE:Factory:ARM/bc`
3. `cd $_`
4. `osc build standard aarch64 bc.spec --vm-type=qemu`

**Expected behavior**
The build should complete without root privileges.

**Observed behavior**
```
...
Writing build configuration
Running build
[sudo] password for root: 
```

wfrisch:
Edited the report. The steps to reproduce should actually work now.

PR #1513, opened 2024-03-11
Add rootless build support for 'qemu' VM type

wfrisch:
Fixes https://github.com/openSUSE/osc/issues/1511

Works for me. For example:
1. Install the `qemu-arm` package
2. `osc co openSUSE:Factory:ARM/bc`
3. `cd $_`
4. `osc build standard aarch64 bc.spec --vm-type=qemu`

Root is not required anymore with this small change.


dmach:
Thanks for the patch.
I've tested it, it works as expected.

PR #1517, opened 2024-03-13
[what]dependson: add --multibuild-package

andreas-schwab:


dmach:
LGTM. Thanks for the patch.

PR #1526, opened 2024-03-21
Readd makeurl debug output

bmwiedemann:
This was lost in refactoring commit 3f14cef53a853296d7f60bf7b14eb86615842cf9
yet it was very useful to discover API queries

I slightly changed the format, so it is easier to use as input
to `osc api`


dmach:
Cool, thank you.

Issue #1532, opened 2024-04-08
Failed run of "osc service manualrun" does not throw non-zero exit code


marcosbc:
**Describe the bug**

We are seeing a behavior where a failed execution of `osc service manualrun` does not throw a non-zero exit code. We know that it failed because the output of the command execution contains the following string: `Aborting: service call failed:`.

This makes it hard to determine if the service run succeeded or not, without having to parse the actual output contents. 

**Versions**
- Operating system: openSUSE Tumbleweed
- Packages: osc-1.6.1-2.1.noarch

**To Reproduce**
1. Create a package in an OBS project where you can execute the `osc` command line (e.g. `osc mkpac mypackage`).
2. Create the following `_service` file:

    ```
    <services>
      <service name="replace_using_env" mode="manual">
        <param name="eval">
    echo Forcing error
    exit 1
    </param>
        <param name="file">filename</param>
      </service>
    </services>
    ```

3. See error:

    ```
    $ osc service manualrun
    Forcing error
    Aborting: service call failed:  /usr/lib/obs/service/replace_using_env --eval
    echo Forcing error
    exit 1
     --file filename --outdir /Users/mbjoerkelund/Dev/projects/ibs/home:mbjoerkelund:buildhooks/failed-osc/tmp0gpoy46t.replace_using_env.service
    $ echo $?
    0
    ```

**Expected behavior**

It would be expected for the above execution to have thrown a non-zero exit code, which would have indicated to the shell that the command failed to be executed.

**Screenshots, console outputs**
```
$ osc service manualrun
...
ERROR:obs-service-go_modules:go mod vendor failed
Aborting: service call failed:  /usr/lib/obs/service/go_modules --archive kustomize-kustomize-v*.tar.gz --compression zst --subdir kustomize --vendorname go-vendor --outdir /path/to/project/tmpim1x0sdd.go_modules.service
$ echo $?
0
```

**Additional context**
N/A

dmach:
Thanks for the report and especially for the reproducer, it's awesome!
I only slightly modified it to return `2` because `1` is a common exit code that appears when `obs-service-replace_using_env` is not installed.

PR #1533, opened 2024-04-08
Propagate exit code from the run() and do_() commandline methods

dmach:
Fixes: #1532

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -2828,7 +2828,6 @@ Please submit there instead, or use --nodevelproject to force direct submission.
         for srid in supersede:
             change_request_state(apiurl, srid, 'superseded',
                                  f'superseded by {rid}', rid)
-        return rid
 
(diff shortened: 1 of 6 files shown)


osc/commandline.py
```diff
@@ -561,7 +561,8 @@ def main(cls, argv=None, run=True):
         cmd.load_legacy_commands()
         if run:
             args = cmd.parse_args(args=argv)
-            cmd.run(args)
+            exit_code = cmd.run(args)
+            sys.exit(exit_code)
         else:
             args = None
         return cmd, args
```
marcosbc, line 568: The `if run:` block would never reach there `return cmd, args` statement. Just wanted to point it out, not sure if that's expected.

dmach: Thanks for checking. Yes, it is expected. `run=False` is used during tests to  check the command and arguments that would be normally executed.

PR #1536, opened 2024-04-15
Move most of the OBS SCM code to obs_scm submodule

dmach:


dmach:
I'm going to ignore the issues pep8speaks reported, because I'm only moving existing code from core.py to a new submodule.

PR #1550, opened 2024-04-26
Migrate 'token' command to obs_api.Token

dmach:


SchoolGuy (approved):
Thanks for taking the time to implement this new version of my PR. I haven't tested this version of the PR but it looks promising to me.

PR #1552, opened 2024-04-29
Improve 'build' and 'buildinfo' commands by injecting *.inc files into spec that is sent to OBS for parsing BuildRequires (boo#1221340)

dmach:


mlschroe:
What's the purpose of this? Shouldn't you just send them in the cpio archive?

dmach:
> What's the purpose of this? Shouldn't you just send them in the cpio archive?

I tried sending them in the cpio archive as you suggest but it did not work.
Is it supposed to work? Does OBS support it?

This is the code I tried right after adding a spec to cpio:
```
    if topdir:
        for include_file in glob.glob(os.path.join(topdir, "*.inc")):
            fn = os.path.basename(include_file)
            print(f"Using local file: {fn}", file=sys.stderr)
            with open(include_file, "rb") as f:
                result_data.append((fn.encode("utf-8"), f.read()))
```

mlschroe:
It's not supported yet, but I have a patch. So it will work in a few days.

Issue #1560, opened 2024-05-09
osc rebuild --all $prj $pkg misbehaviour


bmwiedemann:
**Describe the bug**
I often use `osc wipebinaries --all $prj $pkg` and when I (more than once) did `osc rebuild --all $prj $pkg` I found out that it started to rebuild the whole project instead of just the one package as intended.

**Versions**
- Operating system: `(source /etc/os-release; echo $PRETTY_NAME)` openSUSE Tumbleweed-Slowroll
- Packages: `rpm -q osc` osc-1.6.2-1.1.noarch

**To Reproduce**
1. Run command `osc rebuild --all $prj $pkg`
2. See error

**Expected behavior**
When a package name is given, it should not rebuild the whole project, even with `--all`, to be consistent with the behaviour of `wipebinaries` .  A helpful warning or error message would also work.

dmach:
The `--all` option is a safeguard - if you run `osc rebuild` in a project dir, it wants you to specify `--all` to confirm that you really want to rebuild all packages.

Ignoring the package when `--all` is specified is a bug. Let me add a conflict error there.

dmach:
@bmwiedemann does the linked patch look good to you?

PR #1561, opened 2024-05-10
Fix 'rebuild' command so the '--all' option conflicts with the 'package' argument

dmach:
Fixes: #1560


bmwiedemann (approved):
works fine in my test

PR #1576, opened 2024-06-11
Add `info` command for projects

bmwiedemann:
Without this change, `osc info` in a project checkout dir returned

    Directory '.' is not an OBS SCM working copy of a package


osc/commandline.py
```diff
@@ -7868,9 +7868,11 @@ def do_info(self, subcmd, opts, *args):
...
 
         args = parseargs(args)
-        pacs = Package.from_paths(args)
-
-        for p in pacs:
+        for pdir in args:
+            if (Path(pdir) / ".osc" / "_package").is_file():
+                p = Package(pdir)
+            else:
+                p = Project(pdir, getPackageList=False, wc_check=False)
```
dmach, line 7875: There's a `Store` object that encapsulates working with `.osc` and has several useful methods we could use here. Also, `get_store()` determines the current package/project type and returns either `obs_scm.Store` or `git_scm.GitStore`.

```
from .store import get_store
store = get_store(pdir)

if store.is_package:
    ...
else:
    ...

File: osc/core.py [hunk 2]
```diff
@@ -235,6 +235,13 @@ def cmp(a, b):
 Link info: %s
 """
 
+project_info_templ = """\
+Project name: %s
+Path: %s
+API URL: %s
+Source URL: %s
+"""
+
```
dmach, line 244: It might be better to use KeyValueTable like in this case: https://github.com/openSUSE/osc/blob/master/osc/obs_api/keyinfo_pubkey.py#L39

dmach (changes requested):

osc/core.py
bmwiedemann: I copied this part from package info_templ - should that switch then as well?

osc/commandline.py
bmwiedemann: Done (slightly differently)

osc/core.py
dmach: Let's keep it this way. I'll eventually change it as part of a bigger refactoring.

PR #1577, opened 2024-06-12
add architecture filtering to osc release

lethliel:


osc/_private/api_source.py
```diff
@@ -114,6 +116,8 @@ def release(
     url_query = {"cmd": "release"}
     if repository:
         url_query["repository"] = repository
+    if architecture:
+        url_query["arch"] = architecture
```
dmach, line 120: Just a comment, because this is perfectly fine...
The `if` statement is superfluous now, because the latest osc code ignores any query parameters that are set to None.

PR #1590, opened 2024-06-30
Don't check for store version in project store

andreas-schwab:


dmach:
@andreas-schwab could you give me an example when this is needed?
What's the problem you're trying to solve?

andreas-schwab:
$ osc up
Error: "/daten/src/osc/openSUSE:Factory:RISCV" is not an osc package working copy.

dmach:
I have no idea how you ended with a project without `.osc/_osclib_version` file.
I don't think we should assume the version, shouldn't we rather recommend fixing the working copy using `osc repairwc` ?
My pull request that fixes `osc repairwc` for this use case: https://github.com/openSUSE/osc/pull/1593

andreas-schwab:
It was never created.

dmach:
#1594 confirmed that this wasn't a one-off issue.
Let me merge it to avoid forcing users running `osc repairwc` for all such occurrences.

rhabacker:
Thanks for fixing.

PR #1599, opened 2024-07-08
Add 'gitea_api' module and 'git-obs' executable

dmach:
## TODO
- [x] print a proper error message when there's no default config entry
- [x] `login list`
- [x] `login add`
  - [x] create directories, set secure perms
  - [x] Implement `--set-as-default`
- [x] `login remove`
- [x] `login update`
  - [x] Implement `--set-as-default`
- [x] `ssh-key list`
- [x] `ssh-key add`
   - [x] validate that we're sending a public ssh key
- [x] `ssh-key remove`
- [x] `clone`
  - [ ] replace `--directory` option with an optional argument to get closer to the `git clone` command?
  - [x] move under `repo` command
- [x] `fork`
  - [x] move under `repo` command
- [x] make available as `git obs` subcommand


dmach:
@dcermak could you review the code and check if it works for you?

osc/gitea_api/exceptions.py
```diff
@@ -0,0 +1,59 @@
+import http.client
+import json
+import re
+
+
+class GiteaException(Exception):
+    def __init__(self, response: http.client.HTTPResponse):
```
dcermak, line 7: The type is wrong
```suggestion
import urllib3.response


class GiteaException(Exception):
    def __init__(self, response: urllib3.response.HTTPResponse):
```

osc/gitea_api/repo.py
```diff
@@ -0,0 +1,67 @@
...
+def branch(
+    conn: Connection,
+    owner: str,
+    repo: str,
+    *,
+    new_branch_name: str,
+    old_ref_name: Optional[str] = None,
+):
+    """
+    Create a new branch in a repository.
```
dcermak, line 18: I would find it more useful, if this function would ensured that a branch is there. `ensure_branch` is not a great name, but something along those lines. This essentially requires that I have my own wrapper for this function which does:
```
try:
    branch(c, user, repo)
except BranchExists:
    pass
```

dcermak (changes requested):

osc/gitea_api/repo.py
dmach: How about to follow `os.makedirs()` and introduce `exist_ok = False` argument?

dcermak: yes that would be perfect

osc/gitea_api/repo.py
```diff
@@ -0,0 +1,67 @@
...
+    :param new_branch_name: Name of the branch to create.
+    :param old_ref_name: Name of the old branch/tag/commit to create from.
+    """
+    json_data = {
+        "new_branch_name": new_branch_name,
+        "old_ref_name": old_ref_name,
+    }
+    url = conn.makeurl("repos", owner, repo, "branches")
+    try:
+        return conn.request("POST", url, json_data=json_data)
```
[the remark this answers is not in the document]
dmach: I want to return the response in order the user has full context in case they need it.

DimStar77:
Is this mostly a re-implementation of gitea-tea to interact with a gitea instance?

dmach:
> Is this mostly a re-implementation of gitea-tea to interact with a gitea instance?

The basic functionality is going to be similar.
I'm planning to add interactive reviews similar to what osc has already
and some goodies to support the packaging workflows.
Those still need to be clarified.

IIUC @AdamMajer is working on a staging tooling that overlaps with this tool. We'll need to decide if we need both.

AdamMajer:
Interactive reviews should use "Issue comments" and "Pull Request Reviews" API to read and leave comments on a PR.

My ideal state is to use GPG signatures for approvals and not rely on Gitea as a secure record of this data. But we'll still need a tool to do actual review (ie. diffs) and communication.

PR #1606, opened 2024-08-09
Adding creq release functionality

adrianschroeter:
to become to construct a multi action release request on CLI.

Also reporting created request id when using creq command.


boiko:
Just tried it on a SUSE product use case and it worked as expected!

PR #1617, opened 2024-08-29
Fix local building in git projects

adrianschroeter:
osc did not find it's store and was unable to run a local build in a project git

### DIFF
--- a/osc/git_scm/store.py
+++ b/osc/git_scm/store.py
@@ -31,8 +31,17 @@ class GitStore:
         self.abspath = os.path.abspath(self.path)
 
         # TODO: how to determine if the current git repo contains a project or a package?
-        self.is_project = False
-        self.is_package = os.path.exists(os.path.join(self.abspath, ".git"))
+
+        # NOTE: we have only one store in project-git for all packages
+        self._toplevel = self._run_git(["rev-parse", "--show-toplevel"])
+        config_path = os.path.join(self._toplevel, "_config")
+        pbuild_path = os.path.join(self._toplevel, "_pbuild")
+        if self._toplevel == self.abspath and (os.path.isfile(config_path) or os.path.isfile(pbuild_path)):
+            self.is_project = True
+            self.is_package = False
+        else:
+            self.is_project = False
+            self.is_package = True
 
         self._package = None
         self._project = None
@@ -69,10 +78,19 @@ class GitStore:
     @property
     def project(self):
         if self._project is None:
+            if self._toplevel:
+                try:
+                    with open(os.path.join(self._toplevel, '.osc/_project')) as f:
+                        self._project = f.readline().strip()
+                except FileNotFoundError:
+                    pass
+
+        if self._project is None:
             # get project from the branch name
             branch = self._run_git(["branch", "--show-current"])
 
             # HACK: replace hard-coded mapping with metadata from git or the build service
+            # NOTE: you never know which git repo is supposed to be used in which project
             if branch == "factory":
                 self._project = "openSUSE:Factory"
             else:
--- a/osc/store.py
+++ b/osc/store.py
@@ -38,7 +38,7 @@ def get_store(path, check=True, print_warnings=False):
                 store = git_scm.GitStore(toplevel, check)
                 if print_warnings:
                     git_scm.warn_experimental()
-        except:
+        except subprocess.CalledProcessError:
             # we should distinguish between git cli fails or git is not installed
             pass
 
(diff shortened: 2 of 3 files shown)


adrianschroeter:
@wfeldt this should solve your issue

adrianschroeter:
note: this is not a good implementation and needs extensions if we want to support the experimental _subdirs way as well. However, it unblocks people for now again with local operations.

adrianschroeter:
@wfeldt can you test again if it works for you now? (it does for me)

osc/git_scm/store.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, path, check=True):
 
         # TODO: how to determine if the current git repo contains a project or a package?
         self.is_project = False
-        self.is_package = os.path.exists(os.path.join(self.abspath, ".git"))
+        self.is_package = os.path.exists(os.path.join(self.abspath, ".git")) or os.path.exists(os.path.join(self.abspath, "../.git"))
```
dmach, line 35: Are you really suggesting that package is every directory that has .git in the parent dir?
We should rather find a better way of identifying project and package git checkouts.

osc/git_scm/store.py
```diff
@@ -68,11 +68,16 @@ def apiurl(self):
 
     @property
     def project(self):
+        if self._project is None:
+            with open(os.path.join(os.path.join(self.abspath, '.osc/_project'))) as f:
```
dmach, line 72: Extra `os.path.join()`.

Is the `.osc` directory always present?
Couldn't we use it for detecting project and package names?

osc/obs_scm/project.py
```diff
@@ -199,6 +199,8 @@ def status(self, pac: str):
         if st is None and exists:
             return '?'
         elif st is None:
+            if os.path.exists(os.path.join(self.absdir, '.git')):
+                return '?'
```
dmach, line 203: This is confusing.
`?` indicates a file that is not tracked the current OBS SCM checkout.
Now you return this value even if it is tracked under a **different** SCM (git).

osc/store.py
```diff
@@ -27,6 +27,10 @@ def get_store(path, check=True, print_warnings=False):
         store = git_scm.GitStore(path, check)
         if print_warnings:
             git_scm.warn_experimental()
+    elif os.path.exists(os.path.join(path, "../.git")):
+        store = git_scm.GitStore(os.path.join(path, "../"), check)
+        if print_warnings:
+            git_scm.warn_experimental()
```
dmach, line 33: Again, I'm not convinced that doing this purely based on existence of `../.git` is the right thing to do.

dmach (changes requested):
We should make clear how to properly detect project and package checkouts that are in git and decide how to work with their metadata.

osc/obs_scm/project.py
adrianschroeter: solved different now

osc/git_scm/store.py
adrianschroeter: at least always when used "osc co". However, using git cli now to find toplevel directory as dirk recommended.

osc/store.py
adrianschroeter: done via git cli lockup now.

osc/git_scm/store.py
```diff
@@ -31,8 +31,15 @@ def __init__(self, path, check=True):
...
-        self.is_package = os.path.exists(os.path.join(self.abspath, ".git"))
+
+        # NOTE: we have only one store in project-git for all packages
+        self._toplevel = self._run_git(["rev-parse", "--show-toplevel"])
+        if self._toplevel == self.abspath:
+            self.is_project = True
+            self.is_package = False
+        else:
+            self.is_project = False
+            self.is_package = True
```
dmach, line 42: I believe this is the change that broke unit tests.
See `tests / unit - read only fixtures (pull_request) `

adrianschroeter: I am unable to find that check/failure, however I fixed a different one when executing it local.

There are still two test case failures, but IMHO the test cases are to blame here: tests/test_git_scm_store.py sets up the the work area with plain git (not osc co) and fails over missing .osc/_project later one.

IMHO this can not work, we either should set it up using "osc co" or also setup the .osc meta data.

dmach:
I have pushed a fixed version that passes the unit tests.
I still don't like how it guesses that a git repo contains a project or a package (some form of metadata would be nice), but it doesn't seem to be making things worse at least.

adrianschroeter:
obsolete hopefully by https://github.com/openSUSE/osc/pull/1669

PR #1624, opened 2024-09-11
checkout: allow to checkout obs imported sources of scmsync sources

adrianschroeter:
This allows to skip the check of obs-scm-bridge and checks out what is stored in OBS src server.

Note: This is not necessarly the same representation as in git
      (.obscpio instead of directories or sub modules).
      Also no modification can be committed.
      So this is only intended for people who understand the current
      internal implementation of the source storage. There is no
      guarantee that this won't change.

Checkout of entire project git is not working with this switch, as the packagelist of osc is empty. It may also make no sense to do this as it will not perform for large projects.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -5007,6 +5007,8 @@ Please submit there instead, or use --nodevelproject to force direct submission.
                         help='Use server side generated sources instead of local generation.')
     @cmdln.option('-l', '--limit-size', metavar='limit_size',
                         help='Skip all files with a given size')
+    @cmdln.option('--native-obs-package', action='store_true',
+                        help='Do not clone native scm repositories: Different representation and you will not be able to submit changes!')
     @cmdln.alias('co')
     def do_checkout(self, subcmd, opts, *args):
         """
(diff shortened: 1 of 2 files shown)


adrianschroeter:
should I do this large re-formating of existing code where the linter complains about?
(I suppose we should do it also for all other commands, not only for checkout then)

osc/commandline.py
```diff
@@ -5250,6 +5250,8 @@ def do_browse(self, subcmd, opts, *args):
                         help='Use server side generated sources instead of local generation.')
     @cmdln.option('-l', '--limit-size', metavar='limit_size',
                         help='Skip all files with a given size')
+    @cmdln.option('--native-obs-package', action='store_true',
+                        help='Do not clone native scm repositories: Different representation and you will not be able to submit changes!')
```
dirkmueller, line 5254: overindented (should be one space beyond the opening parentheses)

osc/commandline.py
```diff
@@ -5361,7 +5363,7 @@ def do_checkout(self, subcmd, opts, *args):
             show_project_meta(apiurl, project)
 
             scm_url = show_scmsync(apiurl, project)
-            if scm_url is not None:
+            if scm_url is not None and opts.native_obs_package is False:
```
dirkmueller, line 5366:
```suggestion
            if scm_url is not None and not opts.native_obs_package:
```

osc/core.py
```diff
@@ -3157,7 +3158,7 @@ def checkout_package(
     meta_data = b''.join(show_package_meta(apiurl, project, package))
     root = ET.fromstring(meta_data)
     scmsync_element = root.find("scmsync")
-    if scmsync_element is not None and scmsync_element.text is not None:
+    if native_obs_package is False and scmsync_element is not None and scmsync_element.text is not None:
```
dirkmueller, line 3161:
```suggestion
    if not native_obs_package and scmsync_element is not None and scmsync_element.text is not None:
```

osc/commandline.py
adrianschroeter: yes, but I don't want to become inconsistent with all other options ... let's either change all or none  (but I would do it in another commit)

PR #1625, opened 2024-09-11
Support copy of scmsync packages

adrianschroeter:
There are two ways, either copy the scmsync definition or drop it when creating a new package meta.

If we keep it, we must not ask OBS to copy sources, it is syncing it anyway.

We could add another option to skip the scmsync tag copy, but we keep it as default, because we don't want to give a different view of the the sources to the user.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -3206,7 +3206,7 @@ def checkout_package(
 
 def replace_pkg_meta(
     pkgmeta, new_name: str, new_prj: str, keep_maintainers=False, dst_userid=None, keep_develproject=False,
-    keep_lock: bool = False,
+    keep_lock: bool = False, keep_scmsync: bool = True,
 ):
     """
     update pkgmeta with new new_name and new_prj and set calling user as the
@@ -3229,6 +3229,9 @@ def replace_pkg_meta(
             root.remove(dp)
     if not keep_lock:
         for node in root.findall("lock"):
+            root.remove(node)
+    if not keep_scmsync:
+        for node in root.findall("scmsync"):
             root.remove(node)
     return ET.tostring(root, encoding=ET_ENCODING)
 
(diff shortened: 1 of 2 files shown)


osc/core.py
```diff
@@ -3748,7 +3749,15 @@ def copy_pac(
...
             print('Sending meta data...')
             u = makeurl(dst_apiurl, ['source', dst_project, dst_package, '_meta'])
-            http_PUT(u, data=src_meta)
+            http_PUT(u, data=meta)
+
+    if meta is None:
+        meta = show_package_meta(dst_apiurl, dst_project, dst_package)
+
+    root = ET.fromstring(meta)
+    if not root.find("scmsync") is None:
```
dirkmueller, line 3758:
```suggestion
    if root.find("scmsync") not is None:
```

osc/core.py
```diff
@@ -3748,7 +3749,15 @@ def copy_pac(
...
+        print("Note: package source is managed via SCM")
```
dirkmueller, line 3759: so this breaks client side copy. 

at least with copypac client side copy should actually remove the scmsync tag, otherwise you can't commit afterwards..

dirkmueller (commented):
I think this is really dubious. if you "osc copypac" a scmsynced package then imho one should be able to commit afterwards. which you can not with scmsync.

adrianschroeter:
you can not commit in the source either. In both cases you need to use the SCM tooling.

osc/core.py
adrianschroeter: I can add that ...

adrianschroeter: done

osc/core.py
```diff
@@ -3748,7 +3752,15 @@ def copy_pac(
...
             u = makeurl(dst_apiurl, ['source', dst_project, dst_package, '_meta'])
-            http_PUT(u, data=src_meta)
+            http_PUT(u, data=meta)
+
+    if meta is None:
+        meta = show_package_meta(dst_apiurl, dst_project, dst_package)
+
+    root = ET.fromstring(meta)
+    if root.find("scmsync") is not None:
+        print("Note: package source is managed via SCM")
```
dmach, line 3762: This code is broken. `show_package_meta()` returns a list and that doesn't work with `ET.fromstring(meta)`.
It should be: `root = ET.fromstring(b"".join(meta))`

dmach (changes requested):

osc/core.py
adrianschroeter: show_files_meta is maybe the more direct way, changed

dmach:
There was only a cosmetic issue - if `copy_pac()` returns None, the command-line tool prints that.
So I pushed a small change fixing that.

PR #1626, opened 2024-09-11
Store the container annotation in the "containers" directory

mlschroe:
Like the OBS worker does.

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -203,6 +203,8 @@ class Buildinfo:
         else:
             self.preinstallimage = None
 
+        self.containerannotation = root.findtext("containerannotation")
+
     def has_dep(self, name):
         for i in self.deps:
             if i.name == name:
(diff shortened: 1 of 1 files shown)


dirkmueller (approved):
wfm

osc/build.py
```diff
@@ -1320,6 +1322,12 @@ def __str__(self):
                         print("Using prefered package: " + path + "/" + filename)
                         os.unlink(tffn)
 
+        if bi.containerannotation:
+            if not os.path.exists("containers"):
+                os.makedirs("containers")
```
dmach, line 1327: Replace with `os.makedirs("containers", exist_ok=True)`.
Your code is longer and is prone to race conditions.

osc/build.py
```diff
@@ -1320,6 +1322,12 @@ def __str__(self):
...
+            with open("containers/annotation", "wb") as f:
+                f.write(bi.containerannotation.encode())
```
dmach, line 1329: Wouldn't be better to avoid encoding to bytes?
```
            with open("containers/annotation", "w", encoding="utf-8") as f:
                f.write(bi.containerannotation)
```

dmach (changes requested):

PR #1638, opened 2024-10-08
Add --no-timestamps parameter to build

Thaodan:
Add --no-timestamps parameter which already exists in build.
Help text copied from build.
Include parameters for the same parameter in buildlog for easier
use.


dmach:
Thanks for the patch.
LGTM, merging.

PR #1639, opened 2024-10-09
Support parameters on token triggers

adrianschroeter:
Add missing parameters for token triggers. Esp used for rebuild and release commands.


adrianschroeter:
@dirkmueller please test

dmach:
I have simplified `url_query`.
`makeurl()` is now smart and skips items with value that equals `None`.

Also, I fixed the `set_release_to` error reported by pylint.

dmach:
One more change: `--repo` and `--arch` seem to be widely used in osc so I used those instead of `--repository` and --architecture`. I also replaced `--target-repository` with `--target-repo` to be consistent.

adrianschroeter:
okay, cool
and I just changed one more thing, using "repository" instead of "filter_source_repository" as the api is going to be fixed here now as well (the three options made no sense and were not working properly there).

PR #1647, opened 2024-10-28
fix[git-obs]: use rather git-obs command name

mcepl:
It works with git better and uses git-something mechanism.


dmach:
Thanks for catching that, I forgot to rename the entry point.

PR #1652, opened 2024-11-02
Print buildroot directory when build or chroot finishes

e4t:
This used to be printed but was dropped with a refactoring. Of course, there is the --just-print-buildroot option, but there is no harm printing this information anyway.


dmach:
I've rebased the pull request to pull some test fixes in.

Issue #1648, opened 2024-10-29
Fails to parse results of git project


AdamMajer:
Version: HEAD

```
> osc r systemsmanagement:cockpit
Traceback (most recent call last):
  File "/home/adamm/bin/osc", line 9, in <module>
    osc.babysitter.main()
  File "/home/adamm/work/osc/osc/osc/babysitter.py", line 227, in main
    sys.exit(run(commandline.OscMainCommand()))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/adamm/work/osc/osc/osc/babysitter.py", line 70, in run
    prg.main(argv)
  File "/home/adamm/work/osc/osc/osc/commandline.py", line 296, in main
    exit_code = cmd.run(args)
                ^^^^^^^^^^^^^
  File "/home/adamm/work/osc/osc/osc/commandline_common.py", line 214, in run
    return cmd.run(args)
           ^^^^^^^^^^^^^
  File "/home/adamm/work/osc/osc/osc/commandline.py", line 240, in run
    return self.func(args.command, args, *args.positional_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/adamm/work/osc/osc/osc/commandline.py", line 5900, in do_results
    return self.do_prjresults('prjresults', opts, *args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/adamm/work/osc/osc/osc/commandline.py", line 6017, in do_prjresults
    print('\n'.join(get_prj_results(apiurl, project, hide_legend=opts.quiet,
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/adamm/work/osc/osc/osc/core.py", line 4297, in get_prj_results
    pacs = sorted(list(set(pacs)))
           ^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '<' not supported between instances of 'NoneType' and 'str'

```

PR #1656, opened 2024-11-05
Fix 'results' command for git projects

dmach:
Fixes: #1648


dmach:
Fixed in #1642 already.

PR #1664, opened 2024-11-14
Fix 'log' command to work correctly with --meta --patch options

dmach:


darix:
tested but it also tries to diff r1 against nothing, which throws an error. this could be an API/backend bug though.

dmach:
> tested but it also tries to diff r1 against nothing, which throws an error. this could be an API/backend bug though.

Thanks for testing. It's truly a backend issue: https://github.com/openSUSE/open-build-service/issues/17107

PR #1680, opened 2024-12-23
Fix progress bar code to support progressbar2

scottp-dpaw:
The progress bar code in `meter.PBTextMeter` touches some third-party class state to hide the bar after the job is completed. This works fine for the unmaintained 2018 progressbar library (https://pypi.org/project/progressbar/), however is not compatible with the later progressbar2 fork (https://pypi.org/project/progressbar2/) and causes a crash.

### DIFF
--- a/osc/meter.py
+++ b/osc/meter.py
@@ -13,6 +13,7 @@ from typing import Optional
 try:
     import progressbar as pb
     have_pb_module = True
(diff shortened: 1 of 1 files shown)


dmach:
I was slightly confused when I saw the patch for the first time because progressbar has version 2.x and progressbar2 has version 4.x :)
The code looks good to me. Thanks for the patch!

Issue #1677, opened 2024-12-06
Make the "configuration file ... may have insecure file permissions" warning configurable


johanneskastl:
**Is your feature request related to a problem? Please describe.**
On every `osc` command I issue, I get a warning that my configuration file `may have insecure file permissions`.

In my case, the file is a link, as it is being managed by home-manager. This means that the `chmod 0600` is not successful.

But as I am sure that it does not contain sensitive information (I am using kwallet), I am annoyed by this warning. Can this be configured?

**Describe the solution you'd like**
I would like to make this warning configurable by having a new option inside the configuration file. Of course, this only works if the configuration file is being parsed before checking, which I guess it currently not the case.

**Describe alternatives you've considered**
I thought about sending a PR to change the detection to not error out if it finds a link. Or to check the directory's permissions instead, if a link is found.

But that sounds like it would not prevent bad things too happen, **if someone uses a link AND has a password in the file**.

dmach:
@johanneskastl is it somehow possible to detect that the file is managed by home-manager?
Where does it point to?

johanneskastl:
> @johanneskastl is it somehow possible to detect that the file is managed by home-manager? Where does it point to?

If the file is a link and is pointing to somewhere in `/nix/store/`, chances are high this is from home-manager. In my case, the link goes to `/nix/store/m2iyn0hldpbmv6hpar6inrknwn5y7kqy-home-manager-files/.config/osc/oscrc` (which in turn is a symlink pointing to the actual file inside the nix store).

I do not know if other dotfile management systems also use links, so maybe the warning can be made configurable? Even if the file is no link, if there are no passwords in them (when using keyring/kwallet/...), the permissions should not matter much?

dmach:
@johanneskastl could you check if https://github.com/openSUSE/osc/pull/1685 works for you?

johanneskastl:
> @johanneskastl could you check if #1685 works for you?

Done, looks good to me.

PR #1685, opened 2025-01-06
Mute oscrc permissions warning when there's no password set

dmach:
Fixes: #1677


johanneskastl:
I just copied the `conf.py` from this branch to `/usr/lib/python3.11/site-packages/osc/conf.py` and tried some osc commands.

I did not get a warning due to the wrong permissions on my `oscrc`, as intended by this change. So for me this seems to work.

PR #1701, opened 2025-02-10
Fix typo in core.binary() that caused text files being detected as binary

dmach:


osc/core.py
```diff
@@ -2796,7 +2796,9 @@ def binary(data: bytes):
 
     We're using heuristics according to OBS: src/backend/BSSrcServer/filediff - look for "diff binary detection"
     """
-    binary_chars = re.findall(b"[\x00-\0x07\x0e-\x1f]", data)
+    binary_chars = re.findall(b"[\x00-\x07\x0e-\x1f]", data)
+    if b"\0" in binary_chars:
```
darix, line 2800: wondering if it would make sense to do this "easy" test before the re.findall

dmach: Makes sense. It's more bulletproof in case the regex malfunctions as before.
Force-pushed the change.

PR #1703, opened 2025-02-13
Tell the build script to disable repos/containers handling for normal…

mlschroe:
… builds

This disables following symlinks in the source copy. The options used are only supported in version 20250206 and above, so add a conflicts with older versions to the spec file as well.


osc/build.py
```diff
@@ -1349,6 +1349,10 @@ def __str__(self):
                 else:
                     os.symlink(path, tffn)
 
+    else:
+       buildargs.append("--repos-directory=-")
+       buildargs.append("--containers-directory=-")
```
dmach, line 1354: Bad indent - 3 spaces instead of 4.

dmach (changes requested):

PR #1704, opened 2025-02-13
Improve 'maintenancerequest' command to inherit description from superseded request

e4t:
Commit 083fdf3b introduced a feature for the 'submitrequest' command where the submission description is pre-populated with one of a superseded request.
Here, the same is introduced for the `maintenancerequest` command.   
The behavior of `osc mr -s 1234` is identical to `osc sr -s 1234` - see Issue 1575.

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -4259,7 +4259,14 @@ Please submit there instead, or use --nodevelproject to force direct submission.
             print(f'Using target project \'{target_project}\'{release_in}')
 
         if not opts.message:
-            opts.message = edit_message()
+            msg = ""
+            if opts.supersede:
+                from .obs_api import Request
+
+                req = Request.from_api(apiurl, opts.supersede)
+                msg = req.description + "\n"
+            opts.message = edit_message(template=f"{msg}")
+
 
         supersede_existing = False
         reqs = []


osc/commandline.py
```diff
@@ -4259,7 +4259,13 @@ def do_maintenancerequest(self, subcmd, opts, *args):
             print(f'Using target project \'{target_project}\'{release_in}')
 
         if not opts.message:
-            opts.message = edit_message()
+            if opts.supersede:
+                from .obs_api import Request
+
+                req = Request.from_api(apiurl, opts.supersede)
+                msg = req.description + "\n"
+            opts.message = edit_message(template=f"{msg}")
```
dmach, line 4267: The scanner revealed a problem with undefined variable.
How about adding `msg = ""`?

```
            msg = ""

            if opts.supersede:
                from .obs_api import Request

                req = Request.from_api(apiurl, opts.supersede)
                msg = req.description + "\n"
	
            opts.message = edit_message(template=f"{msg}")
```

dmach (changes requested):

osc/commandline.py
e4t: Yeah, I saw this as well. 
This is what happens when implementing this in the in place for easy testing it and then cut-n-paste code over into the git sources. 
Meanwhile I've pushed a fix. Sorry for the noise.

dmach:
Thanks for the patch and for fixing the issue with it quickly. Merging.

e4t:
Thank you :)

PR #1708, opened 2025-02-17
rpmlint: Add --multibuild-package option.

badshah400:
Allow passing `--multibuild-package` or `-M` to the rpmlint command to specify flavors for a multibuild package. This makes it consistent with other commands, for example `buildlog`, where this option is similarly used.

Also add a help msg for invoking the `rpmlint` command from inside a checked out PKG dir.


dmach:
@badshah400 thanks for the patch. It looks good to me.
The failing tests are unrelated -> merging.

badshah400:
Many thanks for the quick review and merge.

Issue #1707, opened 2025-02-14
osc rdelete prints a warning from git status when invoked outside of a package


dcermak:
**Describe the bug**
A warning from git is printed to stderr when calling `osc rdelete` outside of a git & osc repo.

**Versions**
- Operating system: `Fedora Linux 41 (Forty One)`
- Packages: `osc-1.12.1-432.1.1.fc41.noarch` (unreleased build, pending push to updates-testing)

**To Reproduce**
1. cd into a directory that is neither a git repo nor a osc package
2. Run `osc rdelete -m "cleanup" $prj $pkg`
3. See the output:
```
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

**Expected behavior**
No warning should be printed as this is a remote only command

**Screenshots, console outputs**

```
🕙[ 16:05:29 ] ❯ osc rdelete --recursive --force -m "cleanup" home:dancermak:branches:devel:openSUSE:Factory
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

bmwiedemann:
I get the same effect with `osc comment create request ...` outside of a checkout
with osc-1.12.1-150600.433.1.noarch

ilmanzo:
same issue with many other commands, like `osc my packages`.
seems something introduced in 1.12.1 , on 1.12.0 it's not happening

jirislaby:
@dmach perhaps 96c5a1491c6d34c2f0f32deae3f7894519fd7830 or 29508519be00a127c30f1f8f3d094e306a4b4147?

dmach:
> [@dmach](https://github.com/dmach) perhaps [96c5a14](https://github.com/openSUSE/osc/commit/96c5a1491c6d34c2f0f32deae3f7894519fd7830) or [2950851](https://github.com/openSUSE/osc/commit/29508519be00a127c30f1f8f3d094e306a4b4147) ([#1705](https://github.com/openSUSE/osc/pull/1705)) ?

Yes, that's the root cause. Improving one thing broke another.
Let me look into the code and ideally prevent osc from executing `git` so many times.

PR #1710, opened 2025-02-18
Change how git projects and packages are detected

dmach:
The code no longer traverses directories to find git's toplevel directory, it rather relies that the working directory contains a project or a package, which is consistent with the previous osc behavior.

Fixes: #1707


dmach:
@adrianschroeter does the change look good to you?
I limited the detection code to the working directory and its parent, git topdir is no longer used.
It's not that powerful, but should be behaving more nicely and also consistently with the traditional osc behavior.

osc/git_scm/store.py
```diff
@@ -29,27 +29,32 @@ def is_package_dir(cls, path):
...
-            self.toplevel = None
 
-        # TODO: how to determine if the current git repo contains a project or a package?
         self.is_project = False
         self.is_package = False
 
-        if self.toplevel:
+        self.project_obs_scm_store = None
+
+        parent_dir = os.path.join(self.abspath, "..")
```
adrianschroeter, line 38: ideally we would not only look on directory, because we have git repositories using subdir mechanic already. The config file will change, but the fact that package sources may exist in .../project_level/some/sub/directory/package   exists. IIRC git is searching upwards until / looking for .git directory. Maybe we can do the same here?

osc/git_scm/store.py
```diff
@@ -29,27 +29,32 @@ def is_package_dir(cls, path):
...
+                # there's .git and _config/_pbuild in the working directory -> it's a project
                 self.is_project = True
-                self.is_package = False
             else:
-                self.is_project = False
+                # there's .git and no _config/_pbuild in the working directory -> it's a package
+                self.is_package = True
+        elif os.path.isdir(os.path.join(parent_dir, ".git")):
+            config_path = os.path.join(parent_dir, "_config")
+            pbuild_path = os.path.join(parent_dir, "_pbuild")
```
adrianschroeter, line 51: these files are currently optional, no problem, right?

osc/git_scm/store.py
dmach: This basically means we'll be executing `git` every time osc is ran outside a workdir without `.osc` in it.
Couldn't we narrow down definition of a package? If we knew that it lives in a certain location such as `<project-dir>/packages/` or if it's easily identified with metadata, we could build smarter logic around this lookup.

To avoid calling `.git`, I could implement the traversal in osc, but it somehow doesn't seem entirely right to me.

PR #1711, opened 2025-02-18
Fix build-install CI job

JanZerebecki:
The job unit-ro is from a different file, so "needs" can not find it.
Remove it as there is no dependency.

An alternative would be to merge both files into one. It might save work
on some broken PRs.

---

Allow manual triggering of all Github CI jobs


JanZerebecki:
The line this removes was added in https://github.com/openSUSE/osc/pull/1586 ,  it might have been something like a rebase that went wrong.

All the other CI failures are unrelated to this PR.

PR #1735, opened 2025-03-13
Assume openSUSE:Factory if project cannot be determined

AdamMajer:
Don't assume "factory" branch is needed to build against Factory. This should be the default build configuration.

As-is, it's impossible to initiate a build from other branches, even when using --alternative-project= so having a default is probably advisable. Also, devel projects are being imported with "main" as their default branches.

### DIFF
--- a/osc/git_scm/store.py
+++ b/osc/git_scm/store.py
@@ -136,12 +136,8 @@ class GitStore:
                 # read project from parent directory that contains a project with .osc metadata
                 self._project = self.project_obs_scm_store.project
             if not self._project:
-                # HACK: guess project from git branch
-                branch = self._run_git(["branch", "--show-current"])
-                if branch == "factory":
-                    self._project = "openSUSE:Factory"
-                else:
-                    raise oscerr.NoWorkingCopy(f"Couldn't map git branch '{branch}' to a project")
+                # HACK: assume openSUSE:Factory project if project metadata is missing
+                self._project = "openSUSE:Factory"
         return self._project
 
     @project.setter


AdamMajer:
We should also use `openSUSE:Factory/snapshot` as the default repository , not standard. But I'm not sure where to set this.

dmach:
@AdamMajer aren't we guessing too much already?
AFAIK this wasn't meant to be a supported workflow and I'd rather see it go away than extending it.
I really don't want people running `osc build` in random git checkouts, expecting a sane build result.
They should clone a package together with project as a context to be able to build.
Also, don't forget that we're frequently working with multiple OBS instances.
This would work with `api.opensuse.org` and nothing else.

AdamMajer:
If it's run against random git checkout and it fails, then it fails. That should not be very important here. But if someone clones some random devel project and wants to build it themselves, it would make sense that *something* happens.

We don't really have a plan of putting *any* project information into packages. Package sources should not be changed to put them into different projects. We also can't really say "ok, where's the git clone of the project?" -- processing that entity can take very long time and is unworkable. But building against `openSUSE:Factory/snapshot` as default, makes perfect sense because we are suppose to follow Factory-First principle anyway.

Alternatively, people could use `osc build --alternative-project=....` and build their package too. As-is, this cannot be done at all.

For other API calls, there's the osc parameters to direct them there and we can still use --alternative-project=... parameter to get something out of `build`.

The motive here is not to be perfect or guess correctly. It's simply to allow a pure package clone to be built in the first place.

Finally, the build against `openSUSE:Factory/standard` is not necessarily stable. Checkins on the OBS side can prevent local builds from succeeding. It's better to use `/snapshot` as default instead.

dmach:
@AdamMajer I have slightly modified the code and the commit text.
TL;DR: if there's no project metadata, it defaults to openSUSE:Factory

Issue #1734, opened 2025-03-13
`osc fork` fails because ForkCommand has no attribute 'gitea_conf'


jcronenberg:
**Describe the bug**
When running `osc fork` I get the following error
```
osc fork network:dhcp kea
Using the following Gitea settings:
Traceback (most recent call last):
  File "/usr/bin/osc", line 33, in <module>
    sys.exit(load_entry_point('osc==1.13.0', 'console_scripts', 'osc')())
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/osc/babysitter.py", line 227, in main
    sys.exit(run(commandline.OscMainCommand()))
             ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/osc/babysitter.py", line 70, in run
    prg.main(argv)
    ~~~~~~~~^^^^^^
  File "/usr/lib/python3.13/site-packages/osc/commandline.py", line 289, in main
    exit_code = cmd.run(args)
  File "/usr/lib/python3.13/site-packages/osc/commandline_common.py", line 214, in run
    return cmd.run(args)
           ~~~~~~~^^^^^^
  File "/usr/lib/python3.13/site-packages/osc/commands/fork.py", line 63, in run
    self.print_gitea_settings()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/osc/commandline_git.py", line 81, in print_gitea_settings
    print(f" * Config path: {self.gitea_conf.path}", file=sys.stderr)
                             ^^^^^^^^^^^^^^^
AttributeError: 'ForkCommand' object has no attribute 'gitea_conf'
```

**Versions**
- Operating system: openSUSE Tumbleweed
- Packages: osc-1.13.0-2.1.noarch

**To Reproduce**
1. Setup gitea credentials
2. My gitea config 
```
logins:
- name: opensuse
  url: https://src.opensuse.org
  user: jcronenberg
  token: REDACTED
  ssh_key:
  default: true
```
3. Run `osc fork network:dhcp kea` and see error

dmach:
Interesting, it's definitely a bug, but it didn't show up during testing and several colleagues said they did not trigger it either.
Maybe something that popped up after upgrading python to 3.13?

PR #1742, opened 2025-03-17
Fix 'osc fork' crash in print_gitea_settings()

dmach:
Fixes: #1734


jcronenberg (approved):
I'll just confirm that this fixes the issue for me.

PR #1753, opened 2025-03-26
Fix 'osc fork' when scmsync url points to a commit rather than a branch

dmach:


dmach:
It turns out, this pull request needs more changes.

What data do we have?
- `scmsync` url where the fragment is either a `branch` or a `commit`
- `_scmsync.obsinfo` that contains the following fields: `mtime`, `commit`, `url`, `revision`, `subdir`, `projectscmsync`

If the package points to a commit, we have no branch information in either of the cases.
Knowing the commit allows us to create a new branch and add more commits on top, but there's no branch information the tools could use to make the workflow smoother, without any need to specify the base branch on creating the pull request.

The ideal workflow could be:
- `osc fork <project> <package>`
- original scmsync address is for example `https://src.opensuse.org/pool/package?branch=distro-x.y-maint&commit=00112233`
- the forked scmsync address would be: `https://src.opensuse.org/user/package?branch=for/distro-x.y-maint/feature`
  - the forked repo shouldn't be pinned to any commit, because we're adding more commits and always want to work with HEAD
- `git-obs repo clone user/package`
- `cd user/package`
- `git switch for/distro-x.y-maint/feature`
- make changes
- `git commit`
- `git push`
- `git-obs pr create` - it would work with providing the target branch explicitly!

@AdamMajer @adrianschroeter what do you think? Is there any other way of getting to the branch information without guessing or making assumptions?
Please note I made up a new scmsync url syntax to illustrate the missing branch information.

adrianschroeter:
We changed the bridge to provide the trackingbranch information from now on in the obsinfo file:

https://github.com/openSUSE/obs-scm-bridge/commit/9a7de8bbf2547a393ba9e7e49ea6dfa75a5e361d

PR #1756, opened 2025-03-31
MR check local mods

e4t:
Check local mods on MR just like on SR when done for a single package. 
This can help to avoid expensive mishaps that may not get caught before submission enters QA.


dmach:
Good idea, thanks!
I'll probably add similar functionality to `git-obs pr create` too.

e4t:
Thank you for merging!

PR #1757, opened 2025-03-31
Fix crash for fetching updateinfo.xml

adrianschroeter:
This is no rpm, but a string here. Needs to become a bytes object.

(OBS got fixed to deliver updateinfo.xml in buildinfo for productcomposer builds now leading to this crash)

### DIFF
--- a/osc/fetch.py
+++ b/osc/fetch.py
@@ -172,7 +172,7 @@ class Fetcher:
                 if pac_obj is None:
                     print('Unsupported file type: ', tmpfile, file=sys.stderr)
                     sys.exit(1)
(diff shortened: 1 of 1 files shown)


adrianschroeter:
this is just a hot fix, we still have the problem that "updateinfo.xml" is a fixed name in local cache and therefore all files overwrite each other. We should prefix it with md5sum instead, maybe via PackageQuery class handling it better.

dmach:
@adrianschroeter isn't #1759 fixing this already?

adrianschroeter:
it should

Issue #1762, opened 2025-04-01
Show a WARNING (instead of DEBUG) if SSH keys are used but ssh-keygen is not installed


juliogonzalez:
**Is your feature request related to a problem? Please describe.**
Right now, if ssh-keygen is not available, but SSH keys are used, osc returns an error doesn't help much finding the issue:

```
# osc ls
Server returned an error: HTTP Error 401: Unauthorized
```

Only if debug is enabled, the error is more clear
```
# osc --debug ls
DEBUG: makeurl: /source?
DEBUG: Using ssh key file configured in oscrc: /home/user/.ssh/id_rsa
DEBUG: Skipping signature auth because ssh-keygen is not available
Server returned an error: HTTP Error 401: Unauthorized
```

This happens because osc packaging (at least on openSUSE and SUSE Linux) recommends but does not require ssh-keygen:

```
# needed for ssh signature auth
Recommends:     %{ssh_add_pkg}
Recommends:     %{ssh_keygen_pkg}
```

So for containers, or for things like Micro systems where recommends are not enabled, the ssh packages are not installed.

**Describe the solution you'd like**

While I understand changing `Recommends` with `Requires` is not wanted, I think `osc` could provide a more human error in this case.

As we know a ssh key is configured, but the program is not available, maybe `DEBUG: Skipping signature auth because ssh-keygen is not available` should be at least a WARNING:

```
WARNING: Skipping signature auth because ssh-keygen is not available
```

Or

```
WARNING: Skipping signature auth because ssh-keygen is not available, even if a ssh key file was configured at oscrc
```

In any case something that points to ssh-key gen, as otherwise looks as if the credentials were wrong (HTTP 401) with no more info.

**Describe alternatives you've considered**

I installed the required packages, but took me a while to figure out what was going on :-)

**Additional context**
Nothing in particular

dmach:
I decided only to change the level from debug to warning, which should be quite sufficient.
If the key is not set in oscrc, there's a lot of guesswork involved and it would be uneasy to print sensible information about it.

Does that look good to you?

juliogonzalez:
OK about skipping the guess work, but maybe we can make things a bit clearer so people is not scared if they are not using SSH keys? I left a small suggestion there.

PR #1766, opened 2025-04-02
Change a debug message to a warning if ssh-keygen is not available

dmach:
Fixes: #1762

### DIFF
--- a/osc/connection.py
+++ b/osc/connection.py
@@ -734,6 +734,7 @@ class SignatureAuthHandler(AuthHandlerBase):
 
         if not self.ssh_keygen_path:
             output.print_msg("Skipping signature auth because ssh-keygen is not available", print_to="warning")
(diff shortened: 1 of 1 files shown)


osc/connection.py
```diff
@@ -733,7 +733,7 @@ def set_request_headers_after_401(self, url, request_headers, response):
             return False
 
         if not self.ssh_keygen_path:
-            output.print_msg("Skipping signature auth because ssh-keygen is not available", print_to="debug")
+            output.print_msg("Skipping signature auth because ssh-keygen is not available", print_to="warning")
```
juliogonzalez, line 736:
```suggestion
            output.print_msg("Skipping signature auth because ssh-keygen is not available", print_to="warning")
            output.print_msg("This can be ignored if you are not using SSH keys for authentication", print_to="warning")
```
Maybe a bit more of context for anyone still using passwords? Feel free to rework the message, or to just ignore this comment at all (at least with your change the situation is already better).

Issue #1787, opened 2025-04-27
git-obs command fails when EDITOR environment variable contains arguments


glaubitz:
My EDITOR environment variable is set to `/usr/bin/emacs -nw` which causes the `git-obs` command to fail:

```
glaubitz@suse-laptop:~/suse/git/python-oci-sdk> git-obs -G opensuse pr create
Using the following Gitea settings:
 * Config path: /home/glaubitz/.config/tea/config.yml
 * Login (name of the entry in the config file): opensuse
 * URL: https://src.opensuse.org
 * User: glaubitz

Creating a pull request ...
 * Local git: branch: factory, rev: 2f24b1f036c541c5b6b0e815dfe571db132a6cb1e87658b07bba6a1788cf5c27
 * Source: glaubitz/python-oci-sdk, branch: factory, rev: 2f24b1f036c541c5b6b0e815dfe571db132a6cb1e87658b07bba6a1788cf5c27
 * Target: pool/python-oci-sdk, branch: factory, rev: 30b34d8a680bc98896dad7c2c546475a5c160c805bcff129cf853916f57bae50
Traceback (most recent call last):
  File "/usr/bin/git-obs", line 34, in <module>
    sys.exit(load_entry_point('osc==1.15.1', 'console_scripts', 'git-obs')())
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/osc/commandline_git.py", line 269, in main
    GitObsMainCommand.main()
    ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/osc/commandline_git.py", line 217, in main
    exit_code = cmd.run(args)
  File "/usr/lib/python3.13/site-packages/osc/commandline_common.py", line 217, in run
    return cmd.run(args)
           ~~~~~~~^^^^^^
  File "/usr/lib/python3.13/site-packages/osc/commands_git/pr_create.py", line 184, in run
    message = edit_message(template=NEW_PULL_REQUEST_TEMPLATE.format(**locals()))
  File "/usr/lib/python3.13/site-packages/osc/commands_git/pr_create.py", line 44, in edit_message
    run_editor(f.name)
    ~~~~~~~~~~^^^^^^^^
  File "/usr/lib/python3.13/site-packages/osc/commands_git/pr_create.py", line 32, in run_editor
[...]

PR #1798, opened 2025-05-07
Fix git-obs pr create with multistring EDITOR

andrii-suse:
Fixes: #1787

### DIFF
--- a/osc/commands_git/pr_create.py
+++ b/osc/commands_git/pr_create.py
@@ -3,12 +3,14 @@ import re
 import subprocess
 import sys
 from typing import Optional
+from typing import List
 
 import osc.commandline_git
 
 
-def get_editor() -> str:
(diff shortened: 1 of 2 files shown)


osc/commands_git/pr_create.py
```diff
@@ -17,20 +18,35 @@ def get_editor() -> str:
         candidates = ["vim", "vi"]
 
     editor_path = None
+    args = None
     for i in candidates:
-        editor_path = shutil.which(i)
+        i, *args = i.split()
```
dmach, line 23: I think `shlex.split()` makes more sense here, because theoretically you can have `EDITOR=foo --option="value with spaces"`.

dmach (changes requested):
Please change splitting the EDITOR variable. Everything else looks good to me.

dmach (approved):

PR #1806, opened 2025-05-13
Implement 'git-obs pr dump' command to store pull request information on disk

dmach:


AdamMajer:
> The performance seems to be a bit sub-optimal as the command runs `git fetch` every time it updates an existing directory. Checking if the fetch is needed before doing it would be definitely nice.

You can check via `git cat-file -e <commit>` and use the head hash for commit.. If you are not doing shallow clones, then presence of commit already means that you have all its history too.

AdamMajer:
https://src.opensuse.org/api/swagger#/issue/issueGetCommentsAndTimeline -- this API could be useful here.

It will list the order of all the commits, comments, and even external comments (so comments in other issues/PR/commits that reference this issue) in chronological order. So, while a commit can can have any timestamp associated with it in git, the pull_push event on the PR timeline is timestamp when that commit was actually pushed to the PR.

osc/gitea_api/pr.py
```diff
@@ -200,6 +271,12 @@ def yes_no(value):
 
         return str(table)
 
+    def to_light_dict(self, exclude_columns: Optional[list] = None):
```
dmach, line 292: Make the name more self-explanatory?
Do we need this method at all? It only adds 2 fields to the exclude list.
Also, if we use this to create a dictionary for matching submodule PRs, we're creating a slightly different structure than for a regular PR. Shouldn't we be using the same data structures everywhere?

osc/gitea_api/pr.py
```diff
@@ -27,6 +89,9 @@ def split_id(cls, pr_id: str) -> Tuple[str, str, int]:
         Split <owner>/<repo>#<number> into individual components and return them in a tuple.
         """
         match = re.match(r"^([^/]+)/([^/]+)#([0-9]+)$", pr_id)
+        if not match:
+            match = re.match(r"^([^/]+)/([^/]+)/pulls/([0-9]+)$", pr_id)
```
dmach: We need to explain what the regular expressions do.
Shouldn't the new one be matching full url? It seems to be taking `<owner>/<repo>/pulls/<number>` only.

osc/gitea_api/issue_timeline_entry.py
```diff
@@ -74,6 +74,11 @@ def pr_number(self) -> int:
 
         return PullRequest.get_owner_repo_number(self._data["pull_request_url"])[2]
 
+    def is_empty(self) -> bool:
+        if self._data:
+            return not self._data.get("type", False)
```
dmach, line 79: This is correct, but being a true nit-picker, I'd do `get("type", None)` because that that naturally fits into the expected values of `type` and then I'd convert it to bool by applying `not`.

osc/commands_git/pr_dump.py
```diff
@@ -0,0 +1,356 @@
...
+
+                    if base_commit == head_commit:
+                        submodule_diff["unchanged"][i] = base_submodules[i]
+                        continue
+
+                    # we expect the data to be identical in base and head with the exception of the commit
+                    # we also drop `commit` and add `base_commit` and `head_commit`
+                    data = base_submodules[i].copy()
+                    if base_commit:
+                        del data["commit"]
```
dmach, line 329: Shorten to `data.pop("commit", None)` ?

osc/commands_git/pr_dump.py
```diff
@@ -0,0 +1,356 @@
...
+                    submodule_diff["removed"][i] = base_submodules[i]
+                elif i not in base_submodules and i in head_submodules:
+                    submodule_diff["added"][i] = head_submodules[i]
+                else:
+                    for key in ["branch", "path", "url"]:
+                        # we don't expect migrating packages to another paths, branches etc.
+                        assert base_submodules[i].get(key, None) == head_submodules[i].get(key, None)
+
+                    base_commit = base_submodules[i].get("commit","")
+                    head_commit = head_submodules[i].get("commit","")
```
dmach, line 319: Add space before `""`.
Shouldn't we be using `None` instead of an empty string?

osc/commands_git/pr_dump.py
```diff
@@ -0,0 +1,356 @@
...
+
+            with open(
+                os.path.join(metadata_dir, "referenced-pull-requests.json"),
+                "w",
+                encoding="utf-8",
+            ) as f:
+                json.dump(linked_prs, f, indent=4, sort_keys=True)
+
+            if warnings:
+                return 38
```
dmach: Why 38? Shouldn't we somehow unify exit codes across all commands? `git-obs pr review interactive` Has several exit codes defined already, stating that codes < 10 are reserved for errors. Exit code `1` is commonly used when an exception is raised or program terminates with an error. We could use another value from 2-9 range for warnings.

osc/gitea_api/git.py
```diff
@@ -218,3 +293,54 @@ def status(self, *, porcelain: bool = False, untracked_files: bool = False):
...
+            submodule = match.groupdict()["submodule"]
+            commit = match.groupdict()["commit"]
+            result[submodule]["commit"] = commit
+
+        remote_url = self.get_remote_url()
+        for submodule_entry in result.values():
+            url = submodule_entry["url"]
+            if not url.startswith("../../"):
+                submodule_entry["clone_url"] = url
+                continue
```
dmach, line 424: Shouldn't we set `owner` and `repo` to None?

osc/gitea_api/git.py
```diff
@@ -218,3 +293,54 @@ def status(self, *, porcelain: bool = False, untracked_files: bool = False):
...
+
+        for line in lines:
+            match = SUBMODULE_RE.match(line)
+            if not match:
+                continue
+            submodule = match.groupdict()["submodule"]
+            key = match.groupdict()["key"]
+            value = match.groupdict()["value"]
+            #if key == "url":
+            #    assert value.startswith("../../")
```
dmach, line 405: If we remove the assertion, we risk that someone uses a full URL to a package submodule and we skip it during review.

osc/commands_git/pr_dump.py
```diff
@@ -0,0 +1,356 @@
...
+            all_submodules = sorted(set(base_submodules) | set(head_submodules))
+            warnings = 0
+            for i in all_submodules:
+
+                if base_submodules[i]:
+                    url = base_submodules[i].get("url","")
+                    if not url.startswith("../../"):
+                        print(f"Warning: incorrect path ({url}) in base submodule ({i})", file=sys.stderr)
+                        warnings += 1
+                else:
```
dmach, line 298: Shouldn't this be `if head_submodules[i]:` ?

osc/commands_git/pr_dump.py
```diff
@@ -0,0 +1,356 @@
...
+                    url = base_submodules[i].get("url","")
```
dmach, line 299: replace `base_submodules` with `head_submodules`

osc/commands_git/pr_dump.py
```diff
@@ -0,0 +1,356 @@
...
+                    }
+                )
+
+            # store timeline as <history/> entries
+            timeline = gitea_api.IssueTimelineEntry.list(self.gitea_conn, owner, repo, number)
+            xml_history_list = []
+            for entry in timeline:
+                if entry.is_empty():
+                    import sys
+                    print(f"Warning ignoring empty IssueTimelineEntry", file=sys.stderr)
```
dmach: When this happens? Don't we simply want to skip them without a warning? We're skipping selected entry types already...

osc/commands_git/pr_dump.py
```diff
@@ -0,0 +1,356 @@
...
+                    data["base_commit"] = base_commit
+                    data["head_commit"] = head_commit
+                    submodule_diff["changed"][i] = data
+
+            with open(os.path.join(metadata_dir, "submodules-diff.json"), "w", encoding="utf-8") as f:
+                json.dump(submodule_diff, f, indent=4, sort_keys=True)
+
+            linked_prs = {}
+
+            # body may contain references with both https:// or without, which look indetical in UI. so we must handle both cases:
```
dmach: `indetical` ->`identical`

osc/commands_git/pr_dump.py
```diff
@@ -0,0 +1,356 @@
...
+            for url in re.findall(r"https?://[^\s]+/pulls/\d+", pr_obj.body):
+                if not self.gitea_conn.host in url:
+                    print(f"ignoring PR {url}")
+                    linked_prs[url] = None
+                    continue
+
+                print(f"Linking PR {url}...")
+                _, _, linked_id = url.partition(self.gitea_conn.host + "/")
```
dmach: Shouldn't we parse the URL properly?
Or write an alternative to `Git.get_owner_repo_from_url()`?

osc/commands_git/pr_dump.py
```diff
@@ -0,0 +1,356 @@
...
+                    data["head_commit"] = head_commit
+                    submodule_diff["changed"][i] = data
+
+            with open(os.path.join(metadata_dir, "submodules-diff.json"), "w", encoding="utf-8") as f:
+                json.dump(submodule_diff, f, indent=4, sort_keys=True)
+
+            linked_prs = {}
+
+            # body may contain references with both https:// or without, which look indetical in UI. so we must handle both cases:
+            for url in re.findall(r"https?://[^\s]+/pulls/\d+", pr_obj.body):
```
dmach: Isn't this a guess? I thought such entries should start with `PR: `

AdamMajer: We've settled on following syntax in the Project Git PRs/commit messages.

`PR: org/repo#no`

One per referenced package PR.

This is used by workflow-pr bot and the maintenance tool kit. In the review tooling, you can verify this. And since you are using the PR timeline API, you will already get all the referenced issues and PRs (so prjgit reference  in the package git PR) and comments and other things, without need to parse strings with regular expressions in the first place. :wink:

PR #1807, opened 2025-05-15
Add json output to pr list

andrii-suse:


dmach:
I'm not sure if `KeyValueTableJson` is the right approach.
The current output is not a valid json, it only prints a sequence serialized dictionaries to the stdout, while we need serialized list of dictionaries.

I'm preparing a pull request that wraps data we retrieve from Gitea API into lightweight proxy objects.
I'd probably serialize these, with the resulting dictionary having keys matching with the object properties to have similar (if not identical) structures in Python and json output.

andrii-suse:
> I'm not sure if `KeyValueTableJson` is the right approach. The current output is not a valid json, it only prints a sequence serialized dictionaries to the stdout, while we need serialized list of dictionaries.

That should be easy to fix - I just wanted to demonstrate the idea.

> I'm preparing a pull request that wraps data we retrieve from Gitea API into lightweight proxy objects. I'd probably serialize these, with the resulting dictionary having keys matching with the object properties to have similar (if not identical) structures in Python and json output.

yes, that is very related task, just it raises two concerns:
1. Will text/json output provide different set of fields and potentially different names of the columns? (E.g. "ID" vs "number", "Author" vs "user.email" etc)
2. We still need to do changes like proposed in this SR for some objects that are not planned to have json _data (e.g. obs_api/token.py , gitea_api/conf.py , obs_api/person.py ?)

PR #1808, opened 2025-05-16
Refactor gitea_api to use objects with properties instead of raw JSON data

dmach:


osc/gitea_api/pr.py
```diff
@@ -27,52 +31,149 @@ def split_id(cls, pr_id: str) -> Tuple[str, str, int]:
...
+    def head_repo(self) -> Optional[str]:
+        if not self.is_pull_request:
+            return None
+        return self._data["head"]["repo"]["name"]
+
+    @property
+    def head_branch(self) -> Optional[str]:
+        if not self.is_pull_request:
+            return None
+        return self._data["head"]["ref"]
```
dcermak, line 131: Note that this reference becomes useless once the PR is merged

osc/commands_git/pr_search.py
```diff
@@ -69,11 +69,12 @@ def run(self, args):
...
 
-        text = gitea_api.PullRequest.list_to_human_readable_string(data, sort=True)
-        if text:
-            print(text)
-            print("", file=sys.stderr)
+        if pr_obj_list:
+            pr_obj_list.sort()
+            for pr_obj in pr_obj_list:
+                print(pr_obj.to_human_readable_string())
+                print("", file=sys.stderr)
```
andrii-suse, line 78: will this print multiple empty lines to stderr? Do we want it?

osc/commands_git/pr_list.py
```diff
@@ -27,12 +27,13 @@ def run(self, args):
...
-                print(text)
-                print("", file=sys.stderr)
+            pr_obj_list = gitea_api.PullRequest.list(self.gitea_conn, owner, repo, state=args.state)
+
+            if pr_obj_list:
+                total_entries += len(pr_obj_list)
+                pr_obj_list.sort()
+                for pr_obj in pr_obj_list:
+                    print(pr_obj.to_human_readable_string())
+                    print("", file=sys.stderr)
```
andrii-suse, line 37: Will this print multiple empty lines to stderr? Do we want it?

andrii-suse (approved):
LGT with regard to the comments about the output to stderr in a loop.

Now I wonder if the format changes on server side. Before this change - the tool will continue and try own best to provide the functionality. Now - will it do the same or will it raise an exception trying to create JSON objects?

osc/gitea_api/pr.py
dmach: Yes and no :)
What happens is this:
- `head/sha` changes and is equal to `merge_commit_sha` which is also set during merge
- `merge_base` remains set to the original `head/sha` value

osc/commands_git/pr_search.py
dmach: it separates the entries printed to console
maybe we want to change it to stdout

dmach:
> Now I wonder if the format changes on server side. Before this change - the tool will continue and try own best to provide the functionality. Now - will it do the same or will it raise an exception trying to create JSON objects?

If the format changes on the server, we'll have to handle that in `osc.gitea_api` code so the python api users won't see any difference. If that's not possible, we'll have to declare an api breakage and do things differently.

Issue #1814, opened 2025-05-22
osc doesn't work with git-driven development repo


mcepl:
**Describe the bug**
```
mitmanek:lua-rpm-macros (main) $ git remote -v
myrepo	https://src.opensuse.org/lua/lua-rpm-macros.git (fetch)
myrepo	gitea@src.opensuse.org:lua/lua-rpm-macros.git (push)
mitmanek:lua-rpm-macros (main) $ cd ..
mitmanek:lua_devel (master) $ git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
mitmanek:lua_devel (master) $ git remote -v
origin	https://src.opensuse.org/lua/_ObsPrj.git (fetch)
origin	https://src.opensuse.org/lua/_ObsPrj.git (push)
mitmanek:lua_devel (master) $ cd lua-rpm-macros
mitmanek:lua-rpm-macros (main) $ osc build
Directory '/home/matej/build/lua_devel/lua-rpm-macros' is not a working copy
mitmanek:lua-rpm-macros (main) $ 
```

**Versions**
- Operating system: MicroOS/Tumbleweed packages
- Packages: `osc_wrapper.py` from master branch

**To Reproduce**
1. git clone https://src.opensuse.org/lua/_ObsPrj.git
2. mv _ObsPrj.git dev_lua && cd dev_lua
3. git submodule update --init
4. osc build

**Expected behavior**
should just make the work I ask it to do

**Screenshots, console outputs**
see above

**Additional context**

dmach:
Ok, found it.
The code expects `.git` to be a directory, while it's a file when using git submodules.

PR #1815, opened 2025-05-22
Fix git store when .git is a file with "gitdir: <path>" link

dmach:
Fixes: #1814


mcepl (approved):
Yes, this works. I had problems with this as well.

PR #1816, opened 2025-05-23
Add --export to pr list and pr search

andrii-suse:


behave/features/git-pr.feature
```diff
@@ -13,6 +13,12 @@ Background:
      And I execute "git push"
      And I execute git-obs with args "pr create --title 'Change version' --description='some text'"
 
+@destructive
+Scenario: List pull requests in json
+    When I execute git-obs with args "pr list pool/test-GitPkgA --export"
+    Then the exit code is 0
+# example stdout [{"owner": "pool", "repo": "test-GitPkgA", "requests": [{"is_pull_request": true, "id": "pool/test-GitPkgA#1", "number": 1, "title": "Change version", "body": "some text", "state": "open", "user": "Admin", "draft": false, "merged": false, "allow_maintainer_edit": false, "base_owner": "pool", "base_repo": "test-GitPkgA", "base_branch": "factory", "base_commit": "204e27b646e6e65c210f80514f5e6286a864a4e7", "base_ssh_url": "ssh://gitea@localhost:38591/pool/test-GitPkgA.git", "head_owner": "Admin", "head_repo": "test-GitPkgA", "head_branch": "factory", "head_commit": "1c2c5bd344e3c300fe562c28ece73e6742ba3c63", "head_ssh_url": "ssh://gitea@localhost:38591/Admin/test-GitPkgA.git", "url": "http://localhost:35179/pool/test-GitPkgA/pulls/1"}]}]
```
dmach: Maybe it would be better to remove the comment this completely or test if `stdout contains` some selected keys and values instead.

osc/gitea_api/pr.py
```diff
@@ -174,6 +174,35 @@ def yes_no(value):
 
         return str(table)
 
+    def dict(self, exclude_columns: Optional[list] = None):
```
dmach, line 177: pylint now throws an error that is most likely related to this:
```
osc/gitea_api/pr.py:14:29: E0601: Using variable 'dict' before assignment (used-before-assignment)
```

I think it's a false positive, but we should take a closer look.

osc/commands_git/pr_list.py
```diff
@@ -78,8 +84,20 @@ def run(self, args):
...
-                    print(pr_obj.to_human_readable_string())
-                    print()
+                if not args.export:
+                    for pr_obj in pr_obj_list:
+                        print(pr_obj.to_human_readable_string())
+                        print()
+                else:
+                    sub_res = []
+                    for pr_obj in pr_obj_list:
+                        sub_res.append(pr_obj.dict(exclude_columns=["owner", "repo"]))
```
dmach: Could you write a comment why we're excluding the 2 columns?
Also, it would be nice if you could come up with better variable names than `res_export` and `sub_res`.

osc/commands_git/pr_list.py
```diff
@@ -78,8 +84,20 @@ def run(self, args):
...
+                    d = {"owner": owner, "repo": repo, "requests": sub_res}
```
dmach: This doesn't seem to be used. Should it be removed?

dmach (changes requested):

osc/gitea_api/pr.py
andrii-suse: Yes I am certain that it is false positive. As workaround we can rename this method (`to_dict` ?) or e.g. somehow rewrite `data: dict` in `__init__` method early in the file

osc/commands_git/pr_list.py
andrii-suse: Done, the idea is to avoid duplicated info if we group by owner and repo anyway

osc/commands_git/pr_list.py
andrii-suse: The idea is that we group rows by owner and repo. Added comment few lines above

dmach (approved):
Looks good to me, thanks!

PR #1822, opened 2025-06-06
Add 'git-obs pr review accept' and 'git-obs pr review decline' commands

dmach:


dmach:
The error from the behave test is unrelated to this change. Merging.

Issue #1778, opened 2025-04-15
Cannot authenticate when gnome-askpass is not installed


hramrach:
**Describe the bug**
Cannot authenticate when gnome-askpass is not installed

**Versions**
- Operating system: `openSUSE Leap 15.6`
- Packages: `osc-1.9.0-150400.10.9.1.noarch`

**To Reproduce**
Run any osc command from a X11 terminal when cookie is expired

**Expected behavior**
Authenticate as normal

**Screenshots, console outputs**
WARNING: Ignoring config '~/.config/osc/oscrc' in XDG location, loading an existing config from ~/.oscrc instead
         To fix this, remove '~/.oscrc'
/usr/lib/ssh/ssh-askpass: line 23: /usr/lib/ssh/ksshaskpass: No such file or directory
Load key "/home/hramrach/id_rsa": incorrect passphrase supplied to decrypt private key
ssh-keygen signature creation failed: 255

hramrach:
When /usr/lib/ssh/ssh-askpass does not produce a password osc should fallback to whatever is used in absence of DISPLAY variable, as ssh does.

PR #1841, opened 2025-06-19
Do not use pipe for ssh-keygen input

hramrach:
Fixes: #1778


hramrach:
See the referenced bug.

hramrach:
*not* the case when reading input from stdin

PR #1848, opened 2025-06-24
Fix non-default origin

andrii-suse:


osc/git_scm/store.py
```diff
@@ -328,6 +328,20 @@ def last_buildroot(self, value):
     @property
     def scmurl(self):
         try:
-            return self._run_git(["remote", "get-url", "origin"])
+            return self._run_git(["remote", "get-url", self.origin])
         except subprocess.CalledProcessError:
             return None
+
+    @property
+    def origin(self):
```
dmach: Isn't calling it `origin` a bit misleading?
Shouldn't we pick a better name such as `default_remote`, `current_remote` or something like that?

dmach (changes requested):

dmach (approved):

PR #1849, opened 2025-06-25
Fix detection of gitstore

dirkmueller:


dmach:
Looks good to me, it's fixing my oversight, obviously.
I've added couple more tests to your commit, will merge it once the test runs are finished.

PR #1852, opened 2025-06-26
Fix 'osc-develproject' for opensuse:Factory packages in gitea

### LINKED ISSUE #1827: osc develproject and osc maintainer commands do not work for git-based packages
The osc develproject command produces incorrect output for packages hosted in the src.opensuse.org Gitea instance.

Current output:

osc develproject openSUSE:Factory cockpit
> pruning cache
> pruned 0 files comprised of 0 bytes
Package openSUSE:Factory/cockpit has no devel project
 

Expected output:

osc develproject openSUSE:Factory cockpit
> pruning cache
> pruned 0 files comprised of 0 bytes
systemsmanagement:cockpit/cockpit
 
 
Currently, osc uses an OBS API call to retrieve devel package metadata.  This only works for packages that use OBS for source control.  For packages that use src.opensuse.org for source control, osc should query the devel_packages file here:
https://src.opensuse.org/openSUSE/Factory/src/branch/main/pkgs/_meta/devel_packages
Cacheing this file may also be advantageous?

kenion:
Fix #1827 by pulling info from devel_package file at https://src.opensuse.org/openSUSE/Factory/src/branch/main/pkgs/_meta/devel_packages for Factory packages that are hosted in gitea.  Confirmed with @AdamMajer that this file will persist, so can be used to reliably obtain this data going forward.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -16,6 +16,7 @@ import locale
 import os
 import platform
 import re
+import requests
 import shlex
 import shutil
 import subprocess
@@ -99,6 +100,8 @@ def cmp(a, b):
 DISTURL_RE = re.compile(r"^(?P<bs>.*)://(?P<apiurl>.*?)/(?P<project>.*?)/(?P<repository>.*?)/(?P<revision>.*)-(?P<source>.*)$")
 BUILDLOGURL_RE = re.compile(r"^(?P<apiurl>https?://.*?)/build/(?P<project>.*?)/(?P<repository>.*?)/(?P<arch>.*?)/(?P<package>.*?)/_log$")
 BUFSIZE = 1024 * 1024
+
+DEVEL_PACKAGES_URL = "https://src.opensuse.org/openSUSE/Factory/raw/branch/main/pkgs/_meta/devel_packages"
 
 new_project_templ = """\
 <project name="%(name)s">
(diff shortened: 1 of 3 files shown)


kenion:
I appear to have upset some tests by importing the requests library.  I will mark this PR as a draft until I fix this.

osc/core.py
```diff
@@ -1397,6 +1400,20 @@ def show_devel_project(apiurl, prj, pac):
 
     package_obj = obs_api.Package.from_api(apiurl, prj, pac)
     if package_obj.devel is None:
+        if prj == 'openSUSE:Factory' or prj == 'openSUSE.org:opensuse:Factory':
```
AdamMajer: `openSUSE` needs to be in correct case here

osc/core.py
```diff
@@ -1397,6 +1400,20 @@ def show_devel_project(apiurl, prj, pac):
...
+            # If OBS api doesn't return a devel project, query the gitea devel_packages file
+            try:
+                response = requests.get(DEVEL_PACKAGES_URL)
+            except:
+                return None, None
+            if response.status_code == 200 and response.text:
+                # locate pac in the devel_packages file
+                for line in response.text.splitlines():
+                    parts = line.split()
```
AdamMajer: devel_pkg, _, devel_prj = line.partition(' ')

And then later you can just check of devel_pkg == pac

kenion: Done

osc/core.py
kenion: Done

dmach:
@AdamMajer @adrianschroeter didn't we want to use the `OBS:RejectBranch` package attribute for redirecting to the right devel project? What's the primary source of the data now?

AdamMajer:
> (Daniel didn't want to hardcode the path, so it can also work for other projects).

So, the logic here is to have it hardcoded for openSUSE:Factory only. We can have better solution via attributes at some later point, if required.

But I'm not really aware of any project using the devel project concept except Factory.

dmach:
> @dmach ok to merge this?

I'm not happy about this, because it's supporting a hack instead of fixing it properly,
but it seems we're not getting to a proper solution any time soon and this is a quite useful feature.
So I'm merging it and I've created https://github.com/openSUSE/openSUSE-git/issues/108 as a reminder for @adrianschroeter and @AdamMajer to fix this technical debt.

PR #1856, opened 2025-07-01
Fix detection of packages in git, again

dirkmueller:
if there wasn't a "$PWD/.git" directory, it never found the git package. which means you cannot actually build via a scm-bridge checkout.

### DIFF
--- a/osc/git_scm/store.py
+++ b/osc/git_scm/store.py
@@ -10,7 +10,6 @@ from ..util import yaml as osc_yaml
 
 
 class GitStore:
-
     @classmethod
     def is_project_dir(cls, path):
         try:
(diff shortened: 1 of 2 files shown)


dirkmueller:
> Unit tests are failing after applying your patch. I can take a look if you send me a reproducer. Either paste here or send it to me via chat.

sorry about that, fixed and added a testcase. basically this is a project git checked out via the scm-bridge, with and without subdirectories. so it has both a `.git` and a `.osc` directory in the project base dir, and no `.git` anywhere else.

dmach:
> @dmach I don't really understand the behavior failure? is it finding the "osc git" directory itself? but that shouldn't be treated as a project git because it doesn't have any of the project git defining files?

Yes, I think it found the osc git repo and tried to use it as a project repo.
The current detection code is quite messy, because osc traverses the directory tree and tries to detect project by looking for certain files (such as _config) and it also takes _subdirs in consideration.

Let me check if I can use your `test_nested_pkg_osc_project` test to write a different GitStore code that would work in this case.

dirkmueller:
> Yes, I think it found the osc git repo and tried to use it as a project repo. The current detection code is quite messy, because osc traverses the directory tree and tries to detect project by looking for certain files (such as _config) and it also takes _subdirs in consideration.

which is fine, the configuration is correct for that case as well (except that it is in subdirs, which the build service server-side supports also)

PR #1864, opened 2025-07-09
Add git-obs init command

andrii-suse:


osc/commands_git/repo_init.py
```diff
@@ -0,0 +1,139 @@
...
+    parent = "RepoCommand"
+
+    def init_arguments(self):
+        self.add_argument(
+            "-a",
+            "--anonymous",
+            action="store_true",
+            default=None,
+            help="Clone anonymously via the http protocol",
+        )
```
dmach, line 23: If we use this, we might not be able to push.

osc/gitea_api/git.py
```diff
@@ -51,13 +52,16 @@ def _run_git(self, args: List[str], mute_stderr: bool = False) -> str:
...
-        cmd = ["init"]
+        cmd = ["init", "--object-format=sha256"]
         if initial_branch:
             cmd += ["-b", initial_branch]
         if quiet:
             cmd += ["-q"]
         self._run_git(cmd, mute_stderr=mute_stderr)
 
+    def is_initialized(self):
+        return Path(f"{self.abspath}/.git").is_dir()
```
dmach: `.git` can be a file. See osc commit b6f8c6026db2785d4bc1a665c1ed00f985f86ad5

osc/git_scm/store.py
```diff
@@ -366,3 +366,63 @@ def current_remote(self):
         if result:
             return result
         return "origin"
+
+    def obs_git_init(self, template_dir):
```
dmach: It could take `default_branch="main"` argument. Or do we always want to use `main`?

behave/container-files/opt/setup/gitea.sh
```diff
@@ -95,11 +96,42 @@ function add_ssh_key {
...
+
+git init --initial-branch main
+# git commiter equals to the configured user
+git config user.name "Geeko Packager"
+git config user.email "email@example.com"
+
+echo '_build.*' >> .gitignore
+echo '*.changes merge=merge-changes' >> .gitattributes
+echo '[merge "merge-changes"]' >> .gitconfig
+echo 'driver=/usr/lib/obs/helper/bs_mergechanges %O %B %A' >> .gitconfig
```
dmach, line 118: Install `obs-git-init` in `Containerfile`?
Add `Recommends: obs-git-init` to `osc.spec`?

behave/container-files/opt/setup/gitea.sh
```diff
@@ -95,11 +96,42 @@ function add_ssh_key {
 
 
 create_org pool
+create_org_repo pool new_package main
```
dmach, line 99: I'm really not happy about putting a template among regular packages in `pool`.
It should rather be part of `templates/new-package`.

Nit-picking: why the underscore?

Open questions:
- do we expect different templates for projects and packages?
- do we expect any custom templates in organizations (devel projects?)

behave/features/steps/osc.py
```diff
@@ -111,6 +111,9 @@ class GitObs(CommandBase):
 
     def write_config(self):
         data = {
+            "general": {
+                "git_obs_repo_init_template": f"http://localhost:{self.context.podman.container.ports['gitea_http']}/pool/new_package"
+            },
```
dmach, line 116: This doesn't look right to me.
Should there be such general entry at all?
If so, shouldn't it default to `<owner>/<repo>` that is relative to each gitea instance?
Or shouldn't we create the `templates` org I've suggested elsewhere and use it as default unless overridden in a login entry or from command-line?

osc/gitea_api/conf.py
```diff
@@ -257,3 +257,33 @@ def update_login(
...
+        env_git_obs_repo_init_template = os.environ.get("GIT_OBS_REPO_INIT_TEMPLATE", "")
+        if env_git_obs_repo_init_template:
+            return env_git_obs_repo_init_template
+
+        res = None
+        general = data.get("general")
+        if general:
+            res = general.get("git_obs_repo_init_template")
+
+        for i in data.get("logins", []):
```
dmach, line 325: This duplicates `get_login()`.
Couldn't we simplify it?

osc/commands_git/repo_init.py
```diff
@@ -0,0 +1,139 @@
...
+            exit(1)
+
+        gitea_owner, gitea_repo = (None, None)
+        if "://" in template or (not Path(template).is_dir()):
+            path = template
+            # remove trailing ".git" from path
+            if path.endswith(".git"):
+                path = path[:-4]
+
+            gitea_owner, gitea_repo = path.strip("/").split("/")[-2:]
```
dmach: We should refactor this to a function, I think the code appears in too many places already.

osc/commands_git/repo_init.py
```diff
@@ -0,0 +1,139 @@
+import sys
+
+import osc.commandline_git
+
+
+class RepoInitCommand(osc.commandline_git.GitObsCommand):
+    """
+    Init a git repo
+
+    A wrapper to git-init, plus merge .gitignore and .gitattributes from a template, plus initialize .git/config from a template
```
dmach: This needs to be reworded, for example:
```
Initialize or update a package git repo, setting .gitignore, .gitattributes and other files from a template.
```

I need to look into initializing `.git/config` more closely.

dmach (changes requested):
I've completed the first review, I suppose I'll have more questions later.
We definitely need to clarify the overall design, it doesn't seem to be complete to me.

osc/gitea_api/conf.py
andrii-suse: The only "duplicate" part is a loop over login entries and I don't think about any simplification that can replace a simple loop. E.g. adding an abstraction function with a callback instead of simple loop is not much simpler.

osc/commands_git/repo_init.py
andrii-suse: That parameter is for accessing the template, the new repo will need `git add remote ...` in any case

behave/features/steps/osc.py
andrii-suse: The naming of parameters inside test container doesn't really matter - we just use those to simplify tests (i.e. to make something available, like those default repos that we create , so it is easier to test against them).

behave/container-files/opt/setup/gitea.sh
andrii-suse: I don't think we need those. The test just adds some values that we can verify later. It doesn't matter which exact values are there

behave/container-files/opt/setup/gitea.sh
andrii-suse: The naming inside test container doesn't really matter.

Issue #1894, opened 2025-07-31
osc buildlog: Make it work outside of OBS Build Project


e4t:
As described in issue #1893 , packaging work may take place more in Git trees than in checked out OBS packages. `osc buildlog` is one of the commands that do not allow to specify an OBS project/package and thus will not work in a cloned Git tree, requiring the user to check out an OBS package as well.

In general, there should be a way to obtain the project/package information when working from a Git tree as described in issue #1893 , but as a first measure, `osc buildlog` and possible all the other commands lacking this feature should receive an option to provide project/package information on the command line.

Issue #1893, opened 2025-07-31
osc: Find a Way to make osc 'Project' or 'Package' Commands work from a 'plain' Git Repo


e4t:
After the separation of the SCM from OBS many tasks can be performed in a cloned Git repository - without needing to check out the package from OBS.
The Git repository is missing the `.osc` directory which tells `osc` which project/package is referenced. Therefore, some `osc` commands will require the project & package to be specified explicitly on the command line, some commands require additional flags - like `--alternative-project`, others don't allow either and will simply not be available without an OBS package (`osc buildlog`).

To simplify packaging workflows, find a way to 'attach' an OBS project/package to a cloned Git repository. Note that a repository/branch combo might be referenced from multiple OBS build projects. It should therefore be possible to set a preference.

Fixing command like `obs buildlog` to accept a project/package combo would be a first step to lessen the burden. Still, a solution which will save a packager typing in her day to day work would be highly appreciated.

dmach:
Would a command to set such attributes in the git checkout work for you?
Currently, osc stores some metadata that would normally go to `.osc` in `.git` by calling `git config`.
I suppose the information should be set per branch so you can switch branches together with project and package definitions.

PR #1897, opened 2025-08-01
Add 'git-obs store {set,unset,list}' commands for setting store metadata in git packages and projects

dmach:
Fixes: #1893
Fixes: #1894

### DIFF
--- a/osc/commands_git/store.py
+++ b/osc/commands_git/store.py
@@ -1,0 +1,15 @@
+import osc.commandline_git
+
+
+class StoreCommand(osc.commandline_git.GitObsCommand):
+    """
+    Manage local metadata store
+    """
+
+    name = "store"
+
+    def init_arguments(self):
+        pass
+
+    def run(self, args):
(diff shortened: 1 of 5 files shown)


dmach:
`git-obs store set --project=... --package=... --apiurl=...` sets the values for the current branch in git config
and then osc retrieves them from there.
The values are set for the current branch so you can specify a `project` per branch.
Setting `apiurl` per branch makes sense if you want to maintain several branches with code to be tested against multiple OBS instances.
Also, you can set the `package`. I'd expect that this is where you want to set a default value applicable to all branches.

Please note that the `defaults` work only partially, they are not integrated with osc yet.

Also, a downside of this approach is that you lose all previously set mappings when you delete the git repo and clone a new one.

& if someone has a better idea on naming the command, please let me know.
`store` has always referenced to `.osc` directory and metadata in it, that's why I decided to extend the concept for `.git` and stick with it. But there can be a better name for the command to make it more self-explanatory for the command-line users.

Issue #1896, opened 2025-07-31
Missing support for trackingbranch in _scmsync.obsinfo


mlschroe:
There now is an optional trackingbranch entry in the _scmsync.obsinfo file which makes osc abort:
```
  File "/usr/lib/python3.6/site-packages/osc/obs_api/scmsync_obsinfo.py", line 62, in from_api
    return cls.from_file(response)
  File "/usr/lib/python3.6/site-packages/osc/obs_api/scmsync_obsinfo.py", line 44, in from_file
    return cls.from_string(data)
  File "/usr/lib/python3.6/site-packages/osc/obs_api/scmsync_obsinfo.py", line 34, in from_string
    return cls(**kwargs)
  File "/usr/lib/python3.6/site-packages/osc/util/models.py", line 413, in __init__
    raise TypeError(f"The following kwargs of '{self.__class__.__name__}.__init__()' do not match any field: {unknown_fields_str}")
TypeError: The following kwargs of 'ScmsyncObsinfo.__init__()' do not match any field: 'trackingbranch'
```

Please add
```
trackingbranch: Optional[str] = Field()
```

mlschroe:
I ran into this when I did `iosc co SUSE:SLFO:Kernel:1.0:Build kernel-livepatch-MICRO-6-0_Update_6`.

PR #1898, opened 2025-08-01
Add 'trackingbranch' field to ScmsyncObsinfo model

dmach:
Fixes: #1896


dmach:
@mlschroe @adrianschroeter could you check the patch?
Am I setting `query["trackingbranch"]` correctly?

mlschroe:
I don't think you should deduce the trackingbranch from the revision. The revision can be a tag.

PR #1904, opened 2025-08-13
osc.spec: add "Provides: git-obs"

smithfarm:
Other git plugins, such as LFS, are shipped as "git-foo" packages. A reasonable user might expect to be able to install the git obs plugin by installing a "git-obs" package, but this does not work currently.

With this patch, a user attempting to install "git-obs" will get the git obs plugin via osc.


dmach:
Makes perfect sense, thanks!

contrib/osc.spec
```diff
@@ -158,6 +158,7 @@ Provides:       %{use_python_pkg}-osc = %{version}-%{release}
 Provides:       python3-osc = %{version}-%{release}
 %endif
 %endif
+Provides:       git-obs
```
dmach: I agree with @adrianschroeter 's comment, we should version this.
Could you replace it with the following?
`Provides:       git-obs = %{version}-%{release}`

smithfarm: OK

smithfarm:
> I think you should add "= %version" behind to avoid possible trouble later.

Good point. Thanks!

PR #1925, opened 2025-09-11
Fix tracking the right branch

adrianschroeter:
We use to follow always the default branch, but we have meanwhile the trackingbranch in the scmsync url to select the right one.


dmach:
LGTM, only force-pushed several minor changes (nit-picks).

PR #1938, opened 2025-09-23
Fix lint error E1120 for Manifest constructor

andrii-suse:


osc/git_scm/manifest.py
```diff
@@ -27,7 +27,7 @@ def from_string(cls, text: str) -> "Manifest":
         obj = cls(data)
         return obj
 
-    def __init__(self, data: dict):
+    def __init__(self, data: dict = None):
```
dmach, line 30: Where exactly do we set `dict` to `None`?
Maybe we need to fix it in that place instead.

dmach:
Fixed in: #1940

PR #1957, opened 2025-10-09
Return preinstallimage.info and allow podman to use preinstallimage

mahdiyehmr:
Hello this would download and return preinstallimage.info if preinstallimage exists to be able to check the list of packages in preinstallimage without unpacking it in build environment.
It would also sllow podman to use preinstallimage to speed up the build process. I checked and nothing failed so thought maybe it was a limiting factor that old podman versions had and it's fixed.
I used this with my changes in obs-build to use preinstallimage as container base image. https://github.com/mahdiyehmr/obs-build.

I'm not familiar with your PR process please let me know if I need to apply any changes or if you have any comments.
Thanks

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -325,8 +325,9 @@ def get_preinstall_image(apiurl, arch, cache_dir, img_info, offline=False):
     .. note::
         preinstall image can be used only for new build roots!
     """
-    imagefile = ''
-    imagesource = ''
+    imagefile = ""
+    imagesource = ""
+    info_file = "preinstallimage.info"
     img_bins = []
     for bin in img_info.findall('binary'):
         img_bins.append(bin.text)
@@ -339,16 +340,17 @@ def get_preinstall_image(apiurl, arch, cache_dir, img_info, offline=False):
     img_hdrmd5 = img_info.get('hdrmd5')
     if not img_hdrmd5:
         img_hdrmd5 = img_file
-    cache_path = '%s/%s/%s/%s' % (cache_dir, img_project, img_repository, img_arch)
-    ifile_path = '%s/%s' % (cache_path, img_file)
-    ifile_path_part = '%s.part' % ifile_path
+    cache_path = "%s/%s/%s/%s" % (cache_dir, img_project, img_repository, img_arch)
+    ifile_path = "%s/%s" % (cache_path, img_file)
+    info_file_path = "%s/%s" % (cache_path, info_file)
 
     imagefile = ifile_path
     imagesource = "%s/%s/%s [%s]" % (img_project, img_repository, img_pkg, img_hdrmd5)
+    imageinfo = info_file_path
 
     if not os.path.exists(ifile_path):
         if offline:
-            return '', '', []
+            return "", "", "", []
         url = "%s/build/%s/%s/%s/%s/%s" % (apiurl, img_project, img_repository, img_arch, img_pkg, img_file)
         print("downloading preinstall image %s" % imagesource)
         if not os.path.exists(cache_path):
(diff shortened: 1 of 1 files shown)


osc/build.py
```diff
@@ -366,10 +367,27 @@ def get_preinstall_image(apiurl, arch, cache_dir, img_info, offline=False):
...
+
+        if not os.path.exists(info_file_path):
+            info_url = "%s/build/%s/%s/%s/%s/%s" % (apiurl, img_project, img_repository, img_arch, img_pkg, info_file)
+            print("downloading preinstall image info file")
+            try:
+                gr.urlgrab(info_url, filename=info_file_path_part, text='fetching info')
+                # download ok, rename partial file to final file name
+                os.rename(info_file_path_part, info_file_path)
+            except HTTPError as e:
+                print("Failed to download info file! ecode:%i reason:%s" % (e.code, e.reason))
```
dmach: I know that this follows the existing patterns in osc, but couldn't we improve the code to make it more resilient to race conditions?
Running multiple osc commands at a time might end up downloading into the same file in parallel and then copying garbage as the final file.
How about to store the downloaded file in a temporary directory with a unique name instead?

So something like: 
```
        with tempfile.TemporaryDirectory(dir=cache_path) as temp_dir:
            temp_file = os.path.join(temp_dir, info_file)
            try:
                gr.urlgrab(info_url, filename=temp_file, text='fetching info')
                # download ok, rename partial file to final file name
                os.rename(temp_file, info_file_path)
            except HTTPError as e:
                print("Failed to download info file! ecode:%i reason:%s" % (e.code, e.reason))
```
If you copy my code, please test it properly, I haven't tried it.

osc/build.py
```diff
@@ -623,7 +641,7 @@ def calculate_prj_pac(store, opts, descr):
 
 
 def calculate_build_root_user(vm_type):
-    if vm_type in ("kvm", "podman", "qemu"):
+    if vm_type in ("kvm", "qemu"):
```
dmach: Why did you remove podman from the list?

osc/build.py
```diff
@@ -1498,6 +1517,8 @@ def __str__(self):
         rpmlist.append('preinstallimage: %s\n' % imagefile)
     if imagesource:
         rpmlist.append('preinstallimagesource: %s\n' % imagesource)
+    if imageinfo:
+        rpmlist.append('preinstallimageinfo: %s\n' % imageinfo)
```
dmach: It might be nice to use a f-string instead.

osc/build.py
mahdiyehmr: I wanted to use preinstallimage in podman. This function is called in `build_as_user` function.
```
def build_as_user(vm_type=None):
    if not conf.config.su_wrapper:
        return True
    if calculate_build_root_user(vm_type):
        return True
    return False
```
 Then `build_as_user` is used at line 1215 and doesn't return preinstallimage for podman.
In my tests the comment doesn't apply to using preinstalllimage in podman. I used podman without chroot and unprivileged and it could use preinstallimage and untar it if needed.
```
    if build_as_user(vm_type):
        # preinstallimage extraction will fail because unprivileged user cannot chroot or extract devices from the tarball
        bi.preinstallimage = None
```

osc/build.py
mahdiyehmr: I just changed it in these lines. please let me know if you prefer f-string anywhere else. I didn't change it in prints or variables to keep the coding convention.

osc/build.py
mahdiyehmr: Yes, of course. Since `NamedTemporaryFile` was already used in the code I used that instead of `TemporaryDirectory`. My tests where successful. I was torn between using `delete=false` option and `os.rename()` or keeping delete the default value and copy the file instead but decided to go with the first one.
Please let me know if you prefer something else.

osc/build.py
dmach: I see. In that case, I'd recommend returning `"podman"` back to the list in `calculate_build_root_user()`.
That function determines if `osc build` happens under the current user's account or if we need sudo.
We probably don't want to change that.

I'd rather change the condition to `if build_as_user(vm_type) and vm_type != "podman":`.

mahdiyehmr: Makes sense. Changed it to your suggestion.

mahdiyehmr: It fails now if this build option is not available `--vm-use-container-preinstallimage`. I changed it to check if the build is in podman and that option is available.
```
    # Check if using container preinstallimage for podman
    using_container_preinstall = (vm_type == "podman" and opts.build_opt and 
                                 any("--vm-use-container-preinstallimage" in opt for opt in opts.build_opt))
    
    if build_as_user(vm_type) and not using_container_preinstall:
```

mahdiyehmr:
I wasn't able to run the test locally to see what was the issue but I think it might have been about f-string. Changed it so it would specifically convert them to str.

dmach (approved):

dmach:
Thanks for your patch and sorry about the delays in reviews.
Merging.

PR #1961, opened 2025-10-09
Add 'git-obs staging' command

fwag:


osc/commandline_git.py
```diff
@@ -104,8 +104,10 @@ def add_argument_owner_repo(self, **kwargs):
         )
 
     def add_argument_owner_repo_pull(self, **kwargs):
+        dest = kwargs.pop("dest", "owner_repo_pull")
+        
         return self.add_argument(
-            "owner_repo_pull",
+            dest,
```
dmach, line 112: Could you refactor this into a standalone commit?

osc/commands_git/pr_list.py
```diff
@@ -62,6 +68,15 @@ def run(self, args):
             if args.no_draft:
                 pr_obj_list = [i for i in pr_obj_list if not i.draft]
 
+            if args.labels:
+                # Keep PRs that contain all the labels specified by the user
+                required_labels = set(args.labels)
+                print(f"Filtering by labels: {', '.join(required_labels)}", file=sys.stderr)
```
dmach: I suppose this is a temporary debug print. If not, we should print such messages systematically for all filters.

osc/commands_git/pr_list.py
```diff
@@ -62,6 +68,15 @@ def run(self, args):
...
+                pr_obj_list = [
+                    pr for pr in pr_obj_list 
+                    if not required_labels.isdisjoint(set(pr.labels))
```
dmach: This is probably fine in this case, but when working with larger data, I'd recommend assigning`set(pr.labels)` into a variable before the cycle. Converting to set in each iteration might be expensive.

osc/commands_git/staging_create.copy
```diff
@@ -0,0 +1,175 @@
...
+import re
+from typing import List, Tuple, Optional
+import os
+
+## debugging
+from IPython import embed
+import pprint
+
+
+class StagingCommandGroup(osc.commandline_git.GitObsCommand):
```
dmach: Rename to `StagingGroupCommand` which matches with `git-obs staging group`.

osc/commands_git/staging_create.copy
```diff
@@ -0,0 +1,175 @@
...
+    """
+
+    name = "group"
+    aliases = []  # for compatibility with osc
+    parent = "StagingCommand"
+
+    def init_arguments(self):
+        from osc.commandline_git import complete_pr
+
+        self.add_argument_owner_repo_pull(dest="prs", nargs="+").completer = complete_pr
```
dmach: Consider renaming `prs` to `pr_list` or `pr_id_list` to make it more self-explanatory.

osc/commands_git/staging_create.copy
```diff
@@ -0,0 +1,175 @@
...
+        # re.subn() is useful because it tells us if a replacement was made.
+        updated_message, num_replacements = pattern.subn(new_pr_block, original_message, count=1)
+
+        # 4. If no replacements were made, it means there was no existing block.
+        if num_replacements == 0:
+            updated_message = new_pr_block
+
+        return updated_message.strip() 
+    
+    def clone_or_update(
```
dmach: This seems to be borrowed from `git-obs pr dump`. We probably want to deduplicate the code and move it under `gitea_api`.
If the code differs, we need to figure out how to refactor out the common code if possible.

osc/commands_git/staging_group.py
```diff
@@ -0,0 +1,264 @@
+import sys
+import osc.commandline_git
+
+import re
```
dmach: remove

osc/commands_git/staging_group.py
```diff
@@ -0,0 +1,264 @@
...
+    def init_arguments(self):
+        from osc.commandline_git import complete_pr
+
+        self.add_argument_owner_repo_pull(dest="prs", nargs="+").completer = complete_pr
+        self.add_argument('--title', required=True, help="The new title for the staging PR.")
+        self.add_argument('--branch', required=True, help="The branch to use for the staging PR.")
+        self.add_argument('--workdir', required=True, help="Working directory for git operations.")
+        self.add_argument('--force', required=False, help="Force the operation.", action='store_true')
+    
+    def clone_or_update(
```
dmach: Another `clone_or_update()` to deduplicate.

osc/commands_git/staging_create.copy
```diff
@@ -0,0 +1,175 @@
...
+        from osc.output import tty
+
+        self.print_gitea_settings()
+
+        #pr_obj._data['labels'] 
+        num_entries = 0
+        failed_entries = []
+        all_pkg_prs : List[Tuple[str, str, int]] = []  
+        
+        all_prs: List[Tuple[str, str, str]] = [(args.master_pr[0], args.master_pr[1], args.master_pr[2])] + args.prs
```
dmach: I'm not sure if the types are correct. Isn't args.prs a list of lists?
I would skip creating the new `all_prs` variable.
Why don't you move the body of the cycle below into a function and then apply to the master pr and other prs in 2 separate steps?

osc/commands_git/staging_create.copy
```diff
@@ -0,0 +1,175 @@
...
+        for owner, repo, pull in all_prs:
+            try:
+                pr_obj = gitea_api.PullRequest.get(self.gitea_conn, owner, repo, int(pull))
+                
+                if is_master:
+                    is_master = False
+                    master_owner = owner
+                    master_repo =   repo
+                    master_pr_obj = pr_obj
+                    master_body = pr_obj.body
```
dmach: Doesn't `master_pr_obj` contain all of these already?

osc/commands_git/staging_create.copy
fwag: created an utility class under gitea_api, I didn't yet changed the pr dump code as I had to make commit parameter optional

Issue #1969, opened 2025-10-22
new git workflow branches don't change build target.


Firstyear:
If I have a git repo from the new workflow, the osc build target doesn't change when I change branch.

```
git clone ..... 

git branch
# slfo-main
osc build
# osc builds for SLFO:Main
git checkout slfo-1.2
osc build
# osc builds for SLFO:Main
```

Expected behaviour is that changing git branch should change my build target.

dmach:
Is it a standalone package git repo?
Or is it part of a project (such as _ObsPrj), checked out as project's submodule?

If it's part of a project, then the project metadata is always used regardless the branch of the package.
You must switch the project to the correct branch first.

PR #1974, opened 2025-10-27
Check that the git package has the same branch as the parent project

dmach:
Fixes: #1969

### DIFF
--- a/osc/git_scm/store.py
+++ b/osc/git_scm/store.py
@@ -2,9 +2,10 @@ import fcntl
 import fnmatch
 import json
 import os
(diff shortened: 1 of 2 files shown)


adrianschroeter:
I would not abort here, but just give a warning maybe.

It is IMHO a complete valid usecase to build eg. factory source for SLFO-1.2 for local usage or to validate a submission before hand.

adrianschroeter:
the original user expectation was to build against the other project instead btw.... not sure if this is doable, since it might not be easy to find out which other project was intended though....

PR #1980, opened 2025-11-04
fork: Add --git-branch optional argument

danigm:
This new argument allow the user to fork with a different branch name, so it's easier to create multiple topic forks at the same time.

### DIFF
--- a/osc/commands/fork.py
+++ b/osc/commands/fork.py
@@ -59,6 +59,11 @@ class ForkCommand(osc.commandline.OscCommand):
         self.add_argument_new_repo_name()
 
         self.add_argument(
+            "--git-branch",
+            help="Name of the new git branch",
+        )
+
+        self.add_argument(
             "--no-devel-project",
             action="store_true",
             help="Fork the specified package instead the package from the devel project (which is the place where the package is developed)",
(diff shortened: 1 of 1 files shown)


osc/commands/fork.py
```diff
@@ -174,6 +180,14 @@ def run(self, args):
             print(f" * You may also want to delete '{e.owner}/{e.repo}' and retry")
             sys.exit(1)
 
+        # Crete the new branch if specified
+        if args.git_branch:
+            fork_branch = args.git_branch
+            parent_branch_obj = gitea_api.Branch.create(gitea_conn,
+                                                        fork_owner,
+                                                        fork_repo,
+                                                        new_branch_name=fork_branch)
```
dmach: you're not specifying `old_ref_name` - are you creating an empty branch?

osc/commands/fork.py
```diff
@@ -56,6 +56,11 @@ def init_arguments(self):
             help="Name of the org owning the fork",
         )
 
+        self.add_argument(
+            "--git-branch",
+            help="Name of the new git branch",
+        )
+
```
dmach, line 65: I'd move it below `self.add_argument_new_repo_name()` to achieve the following  order: org, repo, branch

osc/commands/fork.py
danigm: I think that it's using the default if not none, but in any case, I've added the old_ref_name to use the correct branch as base.

danigm:
> @danigm please ignore the previous comment about the lint. just a race condition as I was reviewing while you were fixing it already.
> 
> Also, could you squash the commits? They're fine for now, but the final version should have nice and tidy commits.

I've squashed everything into just one commit and also applied your review suggestions.

Issue #1993, opened 2025-11-11
FR: global read-only flag


andrii-suse:
**Is your feature request related to a problem? Please describe.**
While experimenting with or using osc there is a danger to accidentally invoke API that makes changes due to user mistake or some programming bug

**Describe the solution you'd like**
Having a global "read only" flag can forbid any PUT / POST / DELETE requests sent to the remote.

**Describe alternatives you've considered**
In addition it can be a new command as well e.g. `oscR` that will be the same as `osc` just enforce the flag.

**Additional context**
-

andrii-suse:
> How about setting an environmental variable? That would work for both CLI and the library. `OSC_CONNECTION_GET_ONLY=1`? Or do you have a better idea for the name?

https://github.com/openSUSE/osc/pull/2007

PR #2007, opened 2025-11-28
Add environment variable OSC_HTTP_MANUAL_APPROVE

andrii-suse:
Fixes: #1993

### DIFF
--- a/osc/connection.py
+++ b/osc/connection.py
@@ -342,13 +342,32 @@ def http_request(method: str, url: str, headers=None, data=None, file=None):
         http.client.print(40 * '-')
         http.client.print(method, url)
 
-    if method != "GET" and os.environ['OSC_HTTP_MANUAL_APPROVE']:
-        print("osc is going to send "+ method +" request to " + urlopen_url)
+    if method not in ("GET", "HEAD") and conf.config.http_manual_approve:
+        print(f"Sending HTTP request: {method} {urlopen_url}")
         if data:
-            print(data.decode("utf-8"))
-        y = input("press 'y' to continue or any other key to cancel")
-        if y != 'y' and y != 'Y':
-            raise Exception("Manual abort")
+            print("Data:")
+
+            if file:
+                data_str = file.read()
+                file.seek(0)
+            else:
+                data_str = data
+
+            try:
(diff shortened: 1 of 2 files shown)


osc/connection.py
```diff
@@ -342,6 +342,14 @@ def http_request(method: str, url: str, headers=None, data=None, file=None):
         http.client.print(40 * '-')
         http.client.print(method, url)
 
+    if method != "GET" and os.environ['OSC_HTTP_MANUAL_APPROVE']:
```
dmach: `os.environ['OSC_HTTP_MANUAL_APPROVE']` may raise KeyError.

osc/connection.py
```diff
@@ -342,6 +342,14 @@ def http_request(method: str, url: str, headers=None, data=None, file=None):
...
+        print("osc is going to send "+ method +" request to " + urlopen_url)
```
dmach: Could you use a f-string instead?
Maybe drop this print and rely on the prompt below?

osc/connection.py
```diff
@@ -342,6 +342,14 @@ def http_request(method: str, url: str, headers=None, data=None, file=None):
...
+        if data:
+            print(data.decode("utf-8"))
```
dmach: This may fail. It is not guaranteed that the data is utf-8 text.

osc/connection.py
```diff
@@ -342,6 +342,14 @@ def http_request(method: str, url: str, headers=None, data=None, file=None):
...
+        y = input("press 'y' to continue or any other key to cancel")
+        if y != 'y' and y != 'Y':
+            raise Exception("Manual abort")
+
```
dmach: We should use functions and exceptions osc already offers:
(please excuse the wrong indendation, I've copied the code from a different place and slightly modified it to fit the scenario)

```
               from .output import get_user_input

               reply = get_user_input(
                    f"Do you want to  make '{method}' request to '{urlopen_url}'?",
                    answers={"y": "yes", "n": "no"},
                    default_answer="n",
                )
                if reply == "n":
                    raise oscerr.UserAbort()
```

dmach (changes requested):

Issue #1999, opened 2025-11-16
pr create in a non-forked repository


fwag:
It's not possible at the moment to create a PR in a non-forked repository.
Parameters --target-owner,--target-repo are missing (the default may be the same owner/repo)

```
Traceback (most recent call last):
  File "/usr/bin/git-obs", line 34, in <module>
    sys.exit(load_entry_point('osc==1.22.0', 'console_scripts', 'git-obs')())
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/osc/commandline_git.py", line 312, in main
    GitObsMainCommand.main()
    ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/osc/commandline_git.py", line 251, in main
    exit_code = cmd.run(args)
  File "/usr/lib/python3.13/site-packages/osc/commandline_common.py", line 221, in run
    return cmd.run(args)
           ~~~~~~~^^^^^^
  File "/usr/lib/python3.13/site-packages/osc/commands_git/pr_create.py", line 106, in run
    target_owner = source_repo_obj.parent_obj.owner
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'owner'
```

temporary fix

```
        if source_repo_obj.parent_obj is None:
            target_owner = source_owner
            target_repo = source_repo
        else:
            target_owner = source_repo_obj.parent_obj.owner
            target_repo = source_repo_obj.parent_obj.repo
```

PR #2013, opened 2025-12-08
Fix 'git-obs pr create' when the source repo is not a fork

dmach:
Fixes: #1999 

Depends on: https://github.com/go-gitea/gitea/issues/36116


dmach:
If you specify --target-owner and --target-branch, it allows creating PR to the same repo.
I think this is good to go. Merging.

PR #2022, opened 2025-12-18
Error handling for project wide trigger

adrianschroeter:
We don't support this on the server side, so don't silently eat the argument.


dmach:
Force-pushed a change that fixes pylint errors about undefined `project` and `package` variables.
Those were not entirely correct, but the changed code is the right thing to do and fixes the reported errors as well.

PR #2050, opened 2026-02-13
Add pr forward command to git-obs

fwag:
Automates the workflow of forwarding sources from one branch to another (e.g., Factory to Leap/SLFO) via a PR.

The command handles:
   - Automatic forking of the target repository if it doesn't exist.
   - Smart workspace detection and cloning.
   - Syncing the target branch with upstream.
   - Merging the source branch using the 'theirs' strategy.
   - Cleaning up files not present in the source branch to ensure a clean forward.
   - Optimized LFS fetching for specific incoming commits.
   - Force pushing the synchronized state to the fork and opening the Pull Request.

### DIFF
--- a/osc/commands_git/pr_forward.py
+++ b/osc/commands_git/pr_forward.py
@@ -183,27 +183,28 @@ class PullRequestForwardCommand(osc.commandline_git.GitObsCommand):
             forward_branch = f"for/{target_branch}/forward-{source_commit_sha}"
             print(f"Using forward branch on fork: {forward_branch}", file=sys.stderr)
 
-            # Optimize LFS fetch: fetch only objects for new commits
-            # We identify commits in source_ref that are not in target_branch
-            # and fetch LFS objects for them from the appropriate remote.
-            print(f"Fetching LFS objects from {source_remote} for incoming commits ...", file=sys.stderr)
+            if args.mode not in ["sync"]:
+                # Optimize LFS fetch: fetch only objects for new commits
+                # We identify commits in source_ref that are not in target_branch
+                # and fetch LFS objects for them from the appropriate remote.
+                print(f"Fetching LFS objects from {source_remote} for incoming commits ...", file=sys.stderr)
 
-            try:
-                # Get list of commits unique to source_branch that are not in target_branch
-                commits = git._run_git(["rev-list", f"{source_remote}/{source_branch}", f"^{source_remote}/{target_branch}"]).splitlines()
+                try:
+                    # Get list of commits unique to source_branch that are not in target_branch
+                    commits = git._run_git(["rev-list", f"{source_remote}/{source_branch}", f"^{source_remote}/{target_branch}"]).splitlines()
 
-                if commits:
-                    print(f" * Found {len(commits)} commits to fetch LFS objects for.", file=sys.stderr)
-                    # Loop through commits as requested
-                    for commit in commits:
-                        print(f"   Fetching LFS for commit {commit} ...", file=sys.stderr, end="\r")
-                        git._run_git(["lfs", "fetch", source_remote, commit])
-                    print("", file=sys.stderr)  # Newline after progress
-                else:
-                    print(" * No new commits to fetch LFS objects for.", file=sys.stderr)
-            except subprocess.CalledProcessError as e:
-                # Fallback or ignore if LFS fails/missing
-                print(f"LFS fetch warning: {e}", file=sys.stderr)
+                    if commits:
+                        print(f" * Found {len(commits)} commits to fetch LFS objects for.", file=sys.stderr)
+                        # Loop through commits as requested
+                        for commit in commits:
+                            print(f"   Fetching LFS for commit {commit} ...", file=sys.stderr, end="\r")
+                            git._run_git(["lfs", "fetch", source_remote, commit])
+                        print("", file=sys.stderr)  # Newline after progress
+                    else:
+                        print(" * No new commits to fetch LFS objects for.", file=sys.stderr)
+                except subprocess.CalledProcessError as e:
+                    # Fallback or ignore if LFS fails/missing
+                    print(f"LFS fetch warning: {e}", file=sys.stderr)
 
             # Checkout forward branch (tracking upstream/target_branch)
             print(f"Creating/resetting forward branch '{forward_branch}' from 'upstream/{target_branch}'", file=sys.stderr)


osc/commands_git/pr_forward.py
```diff
@@ -0,0 +1,278 @@
...
+                except Exception:
+                    pass
+                source_ref = f"upstream/{source_branch}"
+
+            # Optimize LFS fetch: fetch only objects for new commits
+            # We identify commits in source_ref that are not in target_branch
+            # and fetch LFS objects for them from the appropriate remote.
+            lfs_remote = "source" if source_url else "upstream"
+            print(f"Fetching LFS objects from {lfs_remote} for incoming commits ...", file=sys.stderr)
+
```
AdamMajer: If the repo from source to target is the same (different branch), there should be no need to move LFS objects around.

fwag: In this case, there are two repositories: the user's forked repository and the pool repository. Normal users cannot create a PR in pool because they don't have the right permissions.

osc/commands_git/pr_forward.py
```diff
@@ -0,0 +1,278 @@
...
+        upstream_owner, upstream_repo = args.owner_repo
+        source_branch = args.source_branch
+        target_branch = args.target_branch
+        source_url = args.source_url
+
+        # Fork if not exists
+        print(f"Checking fork for {upstream_owner}/{upstream_repo} ...", file=sys.stderr)
+        try:
+            fork_obj = gitea_api.Fork.create(self.gitea_conn, upstream_owner, upstream_repo)
```
janvhs, line 82: It think it could be helpful to either postfix the repository name with _git-obs-forward_, like _gnuplot-git-obs-forward_, or don't do that and create special branches, like _factory-git-obs-forward_. This way there is less conflict potential between already existing forks.

These forks or branches can be reset without the fear of losing other work a user potentially did.

I'm not sure if this is a problem, but I might work on a fork on the factory branch to update a package from say version 3 to 4 and want to submit this to openSUSE later.

In the meantime, I need to upstream the released state of factory to Leap or SLFO, so this script would (I did not validate this) reuse my factory branch, which has development commits, and use it to submit to Leap/SLFO. This would lead to wrong commits going into Leap/SLFO.

Using special naming convention for branches, we could delete the existing branch beforehand (maybe prompt user or have a force flag?) and therefore make sure only the commits we actually want to have land into Leap/SLFO.

Does this thought make sense?

osc/commands_git/pr_forward.py
```diff
@@ -0,0 +1,278 @@
...
+            title = args.title or f"Forward {source_branch} to {target_branch}"
+            description = args.description or f"Automated forward of {source_branch} to {target_branch} using git-obs."
+            
+            if args.dry_run:
+                print(f"[DRY RUN] Would create PR: {title}", file=sys.stderr)
+                return
+
+            print("Creating Pull Request ...", file=sys.stderr)
+            try:
+                pr_obj = gitea_api.PullRequest.create(
```
janvhs, line 318: Would it be possible to have an `${EDITOR}` open, so one can refine the commit message? `osc` does this when creating requests and often time one wants to add details like _"closes # `${ISSUE}`"_ or write a proper commit message. The latter doesn't really work, when passing a string to a parameter.

osc/commands_git/pr_forward.py
```diff
@@ -0,0 +1,278 @@
...
+            # Case B: args.workdir contains the repo (args.workdir/repo/.git)
+            elif os.path.isdir(os.path.join(args.workdir, fork_repo, ".git")):
+                repo_dir = os.path.join(args.workdir, fork_repo)
+            # Case C: args.workdir is a parent dir, repo doesn't exist yet -> create args.workdir/repo
+            else:
+                repo_dir = os.path.join(args.workdir, fork_repo)
+                if not os.path.exists(repo_dir):
+                    os.makedirs(repo_dir)
+        else:
+            repo_dir = tempfile.mkdtemp(prefix="git-obs-forward-")
```
janvhs, line 118: Users using git's path based configuration features to override their git email with their work email might want to include something like this in their gitconfig. Maybe that can be documented somewhere?

```gitconfig
[includeIf "gitdir:/tmp/git-obs-forward-*/"]
  path = ~/.config/git/suse
```

osc/commands_git/pr_forward.py
fwag: If your "factory" branch is out of sync between your local git and the remote fork, git will block your push. That's why I added the --force parameter. In any case, using a temporary branch is a good idea.

osc/commands_git/pr_forward.py
fwag: ok added --edit, commit message is automatically created from PR title and PR description

janvhs:
Don't really have a repo to test the new changes on, but the changes look good to me.

Linking the branch name with the commit sha means that you can't reuse an already open PR, which I know my colleagues did in the past, but I'm not sure whether that's wanted anyways, nor am I a maintainer here :^)

osc/commands_git/pr_forward.py
```diff
@@ -0,0 +1,369 @@
...
+            "-e", "--edit",
+            action="store_true",
+            help="Open an editor to edit the pull request title and description.",
+        )
+        self.add_argument(
+            "--allow-unrelated-histories",
+            action="store_true",
+            help="Allow merging unrelated histories",
+        )
+
```
AdamMajer, line 69: Please don't forget to add the "--merge" as an option. Creating a PR directly from the source branch to target branch without a merge commit on top should be default.

osc/commands_git/pr_forward.py
```diff
@@ -0,0 +1,369 @@
...
+                             }: Aborting due to empty title.", file=sys.stderr)
+                    sys.exit(1)
+
+                # remove blank lines between title and body
+                while lines and not lines[0].strip():
+                    lines.pop(0)
+
+                description = "\n".join(lines).strip()
+
+            # Merge Source Branch (Theirs)
```
AdamMajer: like above, all this IFF --merge only please.

osc/commands_git/pr_forward.py
```diff
@@ -0,0 +1,369 @@
...
+                print(f"Removing {len(files_to_remove)} files not present in {source_branch} ...", file=sys.stderr)
+                # Batching git rm to avoid command line length limits
+                chunk_size = 100
+                for i in range(0, len(files_to_remove), chunk_size):
+                    chunk = files_to_remove[i: i + chunk_size]
+                    git._run_git(["rm", "-f", "--"] + chunk)
+
+                git.commit(f"Clean files not present in {source_branch}")
+            else:
+                print("No extra files to clean.", file=sys.stderr)
```
AdamMajer:
```
git merge --no-ff --no-commit -X theirs PR_HEAD_SHA
git read-tree -m PR_HEAD_SHA
git commit -m ..
git clean -fxd
```

The git-read-tree write into the index the info from that other Head and the extra files from current checkout become untracked. Hence need to clean after Merge. But then you no longer need to compare files ... especially since that would ahve to be done recursively.

dmach: `git read-tree -u -m <ref>` includes the clean step

hramrach:
> > And how do you expect to merge Factory into other branch if the source revision is not updated?
> 
> It feels wrong to me to have a `revision` element in the slfo-main branch that points to `factory` in the source repo.
> 
> I ususally have separate branches for slfo-main and factory in the source repo. More often than not, they point to the same commit, but they don't have to.

This is for the case when you want a branch (git reference) other than factory to point to the same commit, or atleast same content as factory. When that's not what you are looking for you should look elsewhere.

AdamMajer:
We probably could use `git read-tree --reset -u OTHER` when reading the other branch.

osc/commands_git/pr_forward.py
AdamMajer: `git read-tree -u --reset <ref>` may be better option, as no merging is done.

I guess we are trying to emulate `git merge -s theirs` while it doesn't exist.

fwag:
This PR did not consider Martin's use case. The purpose of this PR was mainly to report the exact state of one branch in another. In any case, let's see if we can find a common ground.

mwilck:
> If that was the case there would be nothing to do.

I was using [your words](https://github.com/openSUSE/osc/pull/2050#issuecomment-3946068938). But I should have added "or the same content", like you did.

AdamMajer:
https://src.opensuse.org/git-workflow/autogits/pulls/131

So, keep in mind, we do not always want to merge. It's project dependent. When Leap swtiches from 16.0 to 16.1 and has some maintenacne updates, PR from factory should probably switch history to factory branch instead of merging it on top. Merging decision should be maintainer *and* project specific here, so ideally, not by default, and one size fits all approach.

Source remote URL is not guaranteed to stay as-is. Only in pool. But if we have it in pool, then it's already in history, which means commit message just duplicates the information for no reason.. The only reason to have a commit message like this is when reading in a tree based on another commit and not doing a merge. But keeping history is cheap, so why discard it?

hramrach:
> @fwag I've pushed a handful of commits on top of yours. It works relatively nicely now (I hope), even with fast-forwardable branches.
> 
> Open questions:
> 
>     * Shouldn't we add `--mode=<merge|fast-forward|sync>` argument so the user explicitly selects the mode?

Sounds like a good idea. Autodetection is good but it may not always produce the expected result.

Note also the `devel` mode proposed here https://src.opensuse.org/git-workflow/autogits/pulls/131 which would require no merge, merely pushing the desired history to a branch and opening a PR. I don't know how to detect that, though. It's not merged yet which makes it impossible to test. Maybe a separate feature when it becomes available.
 
>     * Do we want to copy all LFS objects over? It doesn't make much sense to me especially with unrelated histories. In such cases, I'm not sure if we shouldn't recommend just syncing the contents with a single commit (read-tree & commit).

I think it is required to copy over all LFS objects to produce correct PR.

> 
>     * Do we want to support merging with unrelated histories?

Probably yes, because bots keep creating unrelated histories for us. Maybe make it optional and not the default so that when the bots are fixed and it's not expected anymore people see it as a warning flag. ie. pass through both the --alow-unrelated-histories flag and the error when it's missing or an equivalent.

>     * Do we want to support `--source-url` for arbitrary git repos? Or do we expect that users will work only with Gitea instance?

I would suggest this is out of scope. Opening the PR is a gitea-specific feature and cannot be done when the source is not in gitea.

>     * Are we planning to support forwarding from one OBS/Gitea instance to another?

Submit form develproject to SLFO is such case.

hramrach:
Also this is very unwieldy. With old Leap project you could use `osc se <package>`  and paste parts of output directly as parameters to `osc diff` and `osc sr` without any possibility of typo. Here getting the branch names is quite a bit of guesswork, maybe the new git search would reveal those. The output of that search is far from compatible with the parameters required by pr.

hramrach:
Also as a suggestion if the way is known it might be nice to provide an option to open the browser with a 'create pr' link similar to what pushing a branch provides but with the specified branches. That way the user can review the changes before creating a PR.

PR #2054, opened 2026-02-22
Feature/staging group no fork flow

fwag:
The 'osc git staging-group' command now supports creating and updating
pull requests for users who have direct push access to the target
repository, eliminating the need for a personal fork.

Previously, the command always operated on a fork of the target
repository. This change introduces a check for repository permissions.
If the user has push or admin rights on the target repository, the
command will create the pull request from a new branch within that
repository.

This also applies when updating an existing pull request. If the user
has push access to the head repository of the pull request, the command
will push updates directly to the head branch, even if it's a fork.

To facilitate this, a new `list_my_repos` method was added to the
Gitea API wrapper to fetch all of a user's repositories and check
their permissions.

Additionally, when updating a pull request, the title is now
automatically appended with a list of the packages that were updated.

### DIFF
--- a/osc/commands_git/staging_group.py
+++ b/osc/commands_git/staging_group.py
@@ -167,6 +167,7 @@ class StagingGroupCommand(osc.commandline_git.GitObsCommand):
                     for repo in forks:
                         if repo.owner.lower() == fork_owner.lower():
                             fork_repo = repo.repo
+                            break
                     if not fork_repo:
                         raise gitea_api.GitObsRuntimeError(f"Cannot find a matching fork of {target_owner}/{target_repo} for user {fork_owner}")
 


osc/commands_git/pr_forward.py
```diff
@@ -0,0 +1,270 @@
...
+        fork_url = fork_repo_obj.ssh_url # Prefer SSH for push if possible
+
+        # Setup Workdir
+        repo_dir = None
+        cleanup = False
+
+        if args.workdir:
+            # Smart detection of workdir
+            # Case A: args.workdir IS the repo (has .git)
+            if os.path.isdir(os.path.join(args.workdir, ".git")):
```
dmach: it can be a file

osc/commands_git/pr_forward.py
```diff
@@ -0,0 +1,270 @@
...
+                repo_dir = args.workdir
+            # Case B: args.workdir contains the repo (args.workdir/repo/.git)
+            elif os.path.isdir(os.path.join(args.workdir, fork_repo, ".git")):
```
dmach: also can be a file

osc/commands_git/pr_forward.py
```diff
@@ -0,0 +1,270 @@
...
+                print(f"Cloning {fork_owner}/{fork_repo} ...", file=sys.stderr)
+                try:
+                    git.clone(fork_url, directory=".")
+                except Exception as e:
+                     raise e
+            else:
+                print(f"Using existing git repo in {repo_dir}", file=sys.stderr)
+
+            # Configure LFS to autodetect remotes
+            git._run_git(["config", "lfs.remote.autodetect", "true"])
```
dmach: NOTE: It might make sense to add this as `_run_git()` parameter that is used every time git gets executed or set it after cloning a repo. Maybe refactoring to a function that configures git repo might also make sense.

osc/commands_git/pr_forward.py
```diff
@@ -0,0 +1,270 @@
...
+
+            # Add Upstream
+            current_remotes = git._run_git(["remote"]).splitlines()
+            if "upstream" not in current_remotes:
+                print(f"Adding upstream remote: {upstream_owner}/{upstream_repo} ({upstream_url})", file=sys.stderr)
+                git.add_remote("upstream", upstream_url)
+            else:
+                git._run_git(["remote", "set-url", "upstream", upstream_url])
```
dmach: This makes an assumption about the `upstream` remote.
Maybe we should document it.
NOTE: check how this maps to the `parent` remote we use elsewhere.

osc/commands_git/staging_group.py
```diff
@@ -205,18 +224,43 @@ def run(self, args):
             target = gitea_api.StagingPullRequestWrapper(self.gitea_conn, target_owner, target_repo, target_number, topdir=temp_dir, cache_directory=cache_dir)
             target.clone()
 
+            has_push_access = False
+            if target.pr_obj._data['head']['repo']['permissions']['push'] or target.pr_obj._data['head']['repo']['permissions']['admin']:
```
dmach: Exposing perms as a property of the PullRequest object would be nicer than this.
Maybe `can_push() -> bool` that hides the internal structures completely and simply answers true/false.

osc/gitea_api/repo.py
```diff
@@ -296,6 +296,23 @@ def list_org_repos(cls, conn: Connection, owner: str) -> List["Repo"]:
...
+    @classmethod
+    def list_my_repos(cls, conn: Connection) -> List["Repo"]:
+        """
+        List repos owned by a user.
+
+        :param conn: Gitea ``Connection`` instance.
+        """
+        q = {
+            # XXX: limit works in range 1..50, setting it any higher doesn't help, we need to handle paginated results
+            "limit": 10**6,
```
dmach, line 308: Remove?

Issue #1565, opened 2024-05-13
Cannot get binaries for a multibuild package without checkout


hramrach:
**Describe the bug**
Cannot get binaries for a multibuild package without checkout

**Versions**
- Operating system: `openSUSE Leap 15.5`
- Packages: `osc-1.6.2-lp155.2.1.noarch`

**To Reproduce**

```
hramrach@localhost:~> ibs getbinaries -d i-i/reduced home:michals:kernel-test-15 installation-images standard ppc64le
WARNING: Ignoring config '~/.config/osc/oscrc' in XDG location, loading an existing config from ~/.oscrc instead
         To fix this, remove '~/.oscrc'
no binaries found: Either the package installation-images does not exist or no binaries have been built.
hramrach@localhost:~> ibs getbinaries -d i-i/reduced -M SLES home:michals:kernel-test-15 installation-images standard ppc64le
WARNING: Ignoring config '~/.config/osc/oscrc' in XDG location, loading an existing config from ~/.oscrc instead
         To fix this, remove '~/.oscrc'
usage: osc [global opts] getbinaries [-h] [-d DIR] [-M FLAVOR] [--sources]
                                     [--debuginfo] [--ccache]
osc [global opts] getbinaries: error: The -M/--multibuild-package option can be only used from a package checkout.

```

**Expected behavior**

Can specify package flavor when package is specified

**Screenshots, console outputs**
If applicable, add screenshots or console outputs to help explain your problem.

**Additional context**

`ibs getbinaries -d i-i/reduced home:michals:kernel-test-15 installation-images:SLES standard ppc64le`

works but that's a very inconsistent interface, and the error does not provide the required information, and in the help text it's buried in description of the -M option.

PR #2063, opened 2026-03-06
Fix oscrc lookup: prefer XDG location, warn if there are multiple configs

dmach:


hramrach:
Why would you do this?!

Of course it breaks osc. After an update I can no longer access OBS.

dmach:
If it doesn't work for you, please open an issue and explain what exactly is wrong.

I'm pretty sure you've got 2 configs and that you were ignoring osc warnings fo years, see the warning in your report: https://github.com/openSUSE/osc/issues/1565

hramrach:
Yes, I did. And the advice to remove the working config in that warning is useless, so I ignored it.

And now the change reverted above broke my config, so I revert it.

This #2121 is no different from an issue except for the fact that code fix is also provided so I don't really see any value in opening one.

And as already said the oscrc is not exclusively for osc.

To not reinvent the wheel bs-upload-kernel also uses it. If you change the priority of the configs you break other tools that re-use the config to not invent their own.

hramrach:
It could be even said that the advice in the warning is malicious: it advises to break my config, without any way to recover.

PR #2081, opened 2026-03-17
Allow copying from scmsync

hramrach:
Fixes: ade04260 ("Support copying from an scmsync source, when target exists.")

The 'when target exists' restriction is superfluous.

Copying was never about tracking the source, SCM or otherwise.

### DIFF
--- a/osc/core.py
+++ b/osc/core.py
@@ -3869,14 +3869,7 @@ def copy_pac(
         if force_meta_update or not found:
             new_meta = meta
 
-    if new_meta:
-        # we are about to create a new package instance. be sure we don't blindly copy scm sources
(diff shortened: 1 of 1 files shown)


jdelvare:
The breakage affects our kernel maintenance update release process. We need to copypac the kgraft package to the release project and this is failing now. We get the following message:
> Note: the source is managed via SCM. You may want to reference directly to the same scm instead?

and the copy does not happen. The command also does NOT return an error code, which makes it impossible for a script to detect and report the problem. If the copypac fails for whatever reason, I would expect an explicit error message, and an error code.

For now, we will be using linkpac instead of copypac as a workaround, but the semantics are different and this could cause other problems. So it would be safer if copypac could be used again.

PR #2107, opened 2026-04-13
Maintainership converter command line tool

### LINKED ISSUE #311: create an utility to convert from old _maintainership.json format to new one
It should provide a parameter, such as '--try-replace-groups', to find matching group names between those imported from OBS.

Groups match when all team members are part of the same group.

ncounter:
Fixes https://github.com/openSUSE/openSUSE-git/issues/311

- [x] convert from [legacy](https://src.opensuse.org/openSUSE/git-workflow-documentation/src/branch/main/architecture/hlfs/bot-group-management.md#backward-compatibility-legacy-format) to [latest](https://src.opensuse.org/openSUSE/git-workflow-documentation/src/branch/main/architecture/hlfs/bot-group-management.md#maintainership-data-format) format
- [x] print the produced content to the CLI output
- [x] `-i` to replace the same file content with the new format

### How to use it
- print the new format to the CLI output:
`git-obs file maintainership migrate _maintainership.json` 
- replace the current `_maintainership.json` file content with the content in the new format
`git-obs file maintainership migrate -i _maintainership.json`

### DIFF
--- a/osc/commands_git/file.py
+++ b/osc/commands_git/file.py
@@ -1,0 +1,15 @@
+import osc.commandline_git
+
+
+class FileCommand(osc.commandline_git.GitObsCommand):
+    """
+    Manage files in a package
+    """
+
+    name = "file"
+
+    def init_arguments(self):
+        pass
+
+    def run(self, args):
+        self.parser.print_help()
--- a/osc/commands_git/file.py
+++ b/osc/commands_git/file.py
@@ -3,7 +3,7 @@ import osc.commandline_git
 
 class FileCommand(osc.commandline_git.GitObsCommand):
     """
-    Manage files in a package
+    Manage metadata files in a repo
     """
 
     name = "file"
--- a/osc/commands_git/file_maintainership_migrate.py
+++ b/osc/commands_git/file_maintainership_migrate.py
@@ -11,6 +11,13 @@ class FileMaintainershipMigrateCommand(osc.commandline_git.GitObsCommand):
 
     def init_arguments(self):
         self.add_argument(
+            "-i",
+            "--in-place",
+            action="store_true",
+            default=False,
+            help="Write the result back to the original file instead of printing to stdout",
+        )
+        self.add_argument(
             "path",
             nargs="?",
             default="_maintainership.json",
@@ -24,4 +31,9 @@ class FileMaintainershipMigrateCommand(osc.commandline_git.GitObsCommand):
             data = f.read()
 
         obj = maintainership.Maintainership.from_string(data)
-        print(obj.to_string())
+
+        if args.in_place:
+            with open(args.path, "w", encoding="utf-8") as f:
+                f.write(obj.to_string())
+        else:
+            print(obj.to_string())
--- a/osc/commands_git/file_maintainership_migrate.py
+++ b/osc/commands_git/file_maintainership_migrate.py
@@ -25,7 +25,14 @@ class FileMaintainershipMigrateCommand(osc.commandline_git.GitObsCommand):
         )
 
     def run(self, args):
+        import os
+        import sys
+
         from osc.gitea_api import maintainership
+
+        if not os.path.exists(args.path):
+            print(f"File not found: {args.path}", file=sys.stderr)
+            return
 
         with open(args.path, "r", encoding="utf-8") as f:
             data = f.read()
--- a/osc/commands_git/file_maintainership_migrate.py
+++ b/osc/commands_git/file_maintainership_migrate.py
@@ -1,12 +1,13 @@
 import osc.commandline_git
 
 
-class MaintainershipCommand(osc.commandline_git.GitObsCommand):
+class FileMaintainershipMigrateCommand(osc.commandline_git.GitObsCommand):
     """
     Read _maintainership.json and convert it from legacy format to the current format
     """
 
-    name = "maintainership-converter"
+    name = "migrate"
+    parent = "FileMaintainershipCommand"
 
     def init_arguments(self):
         self.add_argument(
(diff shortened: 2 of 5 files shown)


tests/test_maintainership_converter.py
```diff
@@ -0,0 +1,136 @@
...
+import io
+import tempfile
+import unittest
+
+
+class TestMaintainershipConverter(unittest.TestCase):
+    def setUp(self):
+        self.tmpdir = tempfile.mkdtemp(prefix="osc_test_maintainership_converter_")
+        self.orig_stdout = sys.stdout
+        sys.stdout = io.StringIO()
```
dmach: This works, but there's a cleaner way of doing it.
In the test, do:
```
stdout = io.StringIO()
with contextlib.redirect_stdout(stdout):
    ...
```

tests/test_maintainership_converter.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def tearDown(self):
+        sys.stdout = self.orig_stdout
+        import shutil
```
dmach: Move up among the top-level imports.
In tests, we don't care about module load times, loading everything on top makes perfect sense.

tests/test_maintainership_converter.py
```diff
@@ -0,0 +1,136 @@
...
+        cmd.run(args)
+        return sys.stdout.getvalue()
+
+    def test_legacy_format_converted(self):
+        """Legacy format is converted to v1.0 format."""
+        legacy_data = {
+            "": ["project-maintainer", "@project-group"],
+            "package1": ["alice", "@pkg1-group"],
+            "package2": ["bob"],
+        }
```
dmach: This is fine, but specifying input as a text would be probably more robust.
But I'm probably nit-picking too much.

dmach (changes requested):

tests/test_maintainership_converter.py
ncounter: Makes perfect sense, it reflect better real life behavior: I adapted all over the places where needed.

tests/test_maintainership_converter.py
ncounter: nice, thanks for the pointer :+1:

osc/commands_git/file.py
```diff
@@ -0,0 +1,15 @@
+import osc.commandline_git
+
+
+class FileCommand(osc.commandline_git.GitObsCommand):
+    """
+    Manage files in a package
```
dmach: Rephrase to "Manage metadata files in a repo" or something.
We're working with a project at this moment, not a package.

osc/commands_git/file_maintainership_migrate.py
```diff
@@ -0,0 +1,39 @@
...
+            nargs="?",
+            default="_maintainership.json",
+            help="Path to the _maintainership.json file (default: %(default)s)",
+        )
+
+    def run(self, args):
+        from osc.gitea_api import maintainership
+
+        with open(args.path, "r", encoding="utf-8") as f:
+            data = f.read()
```
dmach, line 38: This ends with `FileNotFoundError: [Errno 2] No such file or directory: '_maintainership.json'` when the file doesn't exist in the current repository.
Maybe print a warning and exit with 0?
(There's nothing to migrate, which is fine).

dmach (changes requested):

osc/commands_git/file_maintainership_migrate.py
ncounter: makes sense. fixed

hramrach:
It migrates to fully 'pretty' JSON, not lists each on one line. Is that expected?

Also the dictionary key order is not alphabetical. Users are before groups, project before packages. Is that expected?

Also there are null group entries for every package. Not even an empty list, null. That looks bogus, not even conforming to the format.

ncounter:
> It migrates to fully 'pretty' JSON, not lists each on one line. Is that expected?

Yes, the file should still be human-readable, and for readability reasons a single longer lines is worse than shorter multi lines. There could be corner cases, but the majority should fall into the covered case. In any case, JSON cares only about markers/delimiters - so to say -  `{}`, `[]`, `,`, `""`, `''`, and so on, it doesn't care about indentation or spaces or break-line char: you can rearrange at your wishes once it's in the proper shape.
 
> Also the dictionary key order is not alphabetical. Users are before groups, project before packages. Is that expected?

The `_maintainership.json` file is expected to be at project root level, therefore the project entry is expected first, yes. Packages are printed as-is from the original order. Yet another point that can be re-arranged by the owner of the file, nothing that harms the format migration.

> Also there are null group entries for every package. Not even an empty list, null. That looks bogus, not even conforming to the format. 

`null`, empty `[]` list, or no key-entry at all they mean the same: no maintainers for that slice of record.


Long story short: thanks for your feedback. In the end all the three points you raised up are safe and compliant to the expected format, so no worries at all about them. If you still feel they should be part of possible improvements, feel free to open an issue to propose them, we will take into consideration for future iterations.

mmaslano:
Finally, I understood what's the issue:

When engineering teams need to update maintainers (add/remove team members, transfer package ownership), using standard JSON tools produces massive, unreviewable diffs. Say: Team removes one maintainer from one package, the result: ENTIRE file reformatted. Git shows: 2,847 lines changed.
But, the actual change: 1 line (one name removed)  Why this matters? Cannot review what actually changed in pull requests, creates conflicts with concurrent updates from other teams, makes git history unreadable, blocks teams from maintaining their own package assignments.

Could we improve the converter or osc? @dmach I would do it as a new ticket if we want to pursue.

PR #2118, opened 2026-04-22
models: Correctly handle optional values

### LINKED ISSUE #2114: git-obs file maintainership migrate creates null groups
**Describe the bug**
git-obs file maintainership migrate creates null groups

**Versions**
- Operating system: `openSUSE Leap 16.0`
- Packages: 22a1f53f2237d8d6a1b5cadfa1d11653e0a26cc9

**To Reproduce**
git checkout https://src.suse.de/products/SLFO.git
cd SLFO
git checkout slfo-1.2
git-obs file maintainership migrate -i _maintainership.json
git diff | tail -n 10

**Expected behavior**
Format conforms to https://src.opensuse.org/git-workflow/autogits/pulls/236
That specifies that group list should be a list or absent, not null.

**Screenshots, console outputs**
If applicable, add screenshots or console outputs to help explain your problem.
+        },
+        "zypper-lifecycle-plugin": {
+            "users": [
+                "jsrain"
+            ],
+            "groups": null
+        }
+    }
+}
\ No newline at end of file

hramrach:
While the models have the notion of an optional value the optional values are not really optional, and are always present.

This is a partial solution that correctly handles the NotSet value as not present.

However, the Optional type is still defined as union with None, not NotSet.

This is the result of basing models on the failry poorly designed python typing. The Optional(x) alias for Union(X, None) is defined using functionality that is not available outside the typing module, making it impossible to define own alias for Union(X, NotSet).

Fixes: #2114

### DIFF
--- a/osc/gitea_api/maintainership.py
+++ b/osc/gitea_api/maintainership.py
@@ -19,7 +19,7 @@ class MaintainershipHeader(BaseModel):
     """
     A model representing the maintainership document header.
     """
-    document: MaintainershipDocumentType = Field()
+    document: MaintainershipDocumentType = Field(default="obs-maintainers")
     version: str = Field(default="1.0")
 
 
--- a/osc/gitea_api/maintainership.py
+++ b/osc/gitea_api/maintainership.py
@@ -40,25 +40,27 @@ class Maintainership(BaseModel):
 
         data = json.loads(text)
 
+        def ug_to_dict(users, groups):
+            result = {}
+            if users:
+                result["users"] = users
+            if groups:
+                result["groups"] = groups
+            return result
+
         if data.get("header", None) is None:
             # format: legacy
             # entry with "" name holds project maintainers
             project_data = data.pop("", [])
             p_users = [i for i in project_data if not i.startswith("@")]
             p_groups = [i[1:] for i in project_data if i.startswith("@")]
-            project = {
-                "users": p_users if p_users else None,
-                "groups": p_groups if p_groups else None,
-            }
+            project = ug_to_dict(p_users, p_groups)
             packages = {}
             for package, package_data in data.items():
                 if isinstance(package_data, list):
                     pkg_users = [i for i in package_data if not i.startswith("@")]
                     pkg_groups = [i[1:] for i in package_data if i.startswith("@")]
-                    packages[package] = {
-                        "users": pkg_users if pkg_users else None,
-                        "groups": pkg_groups if pkg_groups else None,
-                    }
+                    packages[package] = ug_to_dict(pkg_users, pkg_groups)
             data = {
                 # version is the default (latest), because we do conversion to that format
                 "header": {"document": "obs-maintainers"},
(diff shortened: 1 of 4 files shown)


osc/gitea_api/maintainership.py
```diff
@@ -40,25 +40,27 @@ def from_string(cls, text: str) -> "Maintainership":
...
             # format: legacy
             # entry with "" name holds project maintainers
             project_data = data.pop("", [])
             p_users = [i for i in project_data if not i.startswith("@")]
             p_groups = [i[1:] for i in project_data if i.startswith("@")]
-            project = {
-                "users": p_users if p_users else None,
-                "groups": p_groups if p_groups else None,
-            }
+            project = ug_to_dict(p_users, p_groups)
```
ncounter, line 57: I'd rather drop the *ug_to_dict* function that is very tailored on this specific case and not really reusable, and do the logic in line here instead

```suggestion
            project = {}
            if p_users:
                project["users"] = p_users
            if p_groups:
                project["groups"] = p_groups
```

hramrach: That is the code before the change that produces the wrong result.

hramrach: ok, it would work this way

hramrach: but this code suggestion only replaces the first use of the function, not the other one?

dmach:
To remove the entries with "null" as a value, I'd consider doing something close to what pydantic does:
```
diff --git a/osc/util/models.py b/osc/util/models.py
index 448f4717..58dfcb91 100644
--- a/osc/util/models.py
+++ b/osc/util/models.py
@@ -469,18 +469,21 @@ class BaseModel(metaclass=ModelMeta):
             return False
         return self._get_cmp_data() < other._get_cmp_data()
 
-    def dict(self):
+    def dict(self, *, exclude_none: bool = False):
+        print(self, "DICT", exclude_none)
         result = {}
         for name, field in self.__fields__.items():
             if field.exclude:
                 continue
             value = getattr(self, name)
+            if exclude_none and value is None:
+                continue
             if value is not None and field.is_model:
-                result[name] = value.dict()
+                result[name] = value.dict(exclude_none=exclude_none)
             elif value is not None and field.is_model_list:
-                result[name] = [i.dict() for i in value]
+                result[name] = [i.dict(exclude_none=exclude_none) for i in value]
             elif value is not None and field.is_model_dict:
-                result[name] = {k: v.dict() for k, v in value.items()}
+                result[name] = {k: v.dict(exclude_none=exclude_none) for k, v in value.items()}
             else:
                 result[name] = value
```

dmach (changes requested):
I'm nervous about changing a lot of underlying code related to models.
I'd rather change `.dict()` to skip the `None` values as mentioned in my comment.

hramrach:
Yes, feel free to do it differently. This breaks FromParent, and I did not find why so far.

hramrach:
Still, the fundamental problem with the models remains, even with the `exclude_none` workaround.

Optional values are not really optional, they are always present.

dmach:
Fixed in #2126 - that one doesn't refactor internals of the models and is generally safer.

PR #2119, opened 2026-04-22
Produce standard JSON formatting whan saving _maintainership.json

### LINKED ISSUE #2115: git-obs file maintainership migrate uses non-standard formatting
**Describe the bug**
git-obs file maintainership migrate uses non-standard formatting

**Versions**
- Operating system: `openSUSE Leap 16.0`
- Packages: `22a1f53f2237d8d6a1b5cadfa1d11653e0a26cc9`

**To Reproduce**
git checkout https://src.suse.de/products/SLFO.git
cd SLFO
git checkout slfo-1.2
git-obs file maintainership migrate -i _maintainership.json
git add _maintainership.json
cat _maintainership.json | jq -S . > _maintainership.json.new ; mv _maintainership.json.new _maintainership.json

**Expected behavior**
Standard format such as `jq -S` is used so that the file content can be re-formatted automatically

**Screenshots, console outputs**
If applicable, add screenshots or console outputs to help explain your problem.
     "zypper-lifecycle-plugin": {
+      "groups": null,
       "users": [
         "jsrain"
-            ],
-            "groups": null
+      ]
     }
+  },
+  "project": {
+    "groups": null,
+    "users": [
+      "autobuild-review"
+    ]
   }
 }

**Additional context**
The format is close but not quite the same as `jq -S` outputs which is supported by many tools.

hramrach:
To be able to perform automated updates to the _maintainership.json file it needs to be in a format that is standardized and deterministic so that any tool can produce bit-identical output for the same data.

The format is close to what `jq -S` produces but not quite the same.

One minor problem is different amount of indentation.

The more serious problem is bespoke ordering of dictionary keys that is specific to this particular implementation. The only widely implemented key ordering is alphabetical.

While 'as is' key ordering may be possible with some implementations it is not as widely available, and causes non-determinism when adding new package entries.

Fixes: #2115

### DIFF
--- a/osc/util/models.py
+++ b/osc/util/models.py
@@ -506,8 +506,8 @@ class BaseModel(metaclass=ModelMeta):
         import json
 
         with open(path, "w", encoding="utf-8") as f:
-            # we prefer key ordering according to the fields in the model
-            json.dump(self.dict(), f, sort_keys=False, indent=4)
+            # we prefer fixed, well-defined key ordering
+            json.dump(self.dict(), f, sort_keys=True, indent=2)
 
     @classmethod
     def from_string(cls, text: str) -> "Self":


bmwiedemann (approved):
LGTM

Additionally, a python dict normally uses randomized hash seeds (unless a caller does `export PYTHONHASHSEED=1234` ) so without sorting, it might just have random order.

dmach:
> Additionally, a python dict normally uses randomized hash seeds (unless a caller does `export PYTHONHASHSEED=1234` ) so without sorting, it might just have random order.

I don't think the statement is valid on python 3.7+ where the dictionary keeps the order of the keys as they were created. It is not guaranteed on python 3.6 but it's also implemented that way.

hramrach:
> > Additionally, a python dict normally uses randomized hash seeds (unless a caller does `export PYTHONHASHSEED=1234` ) so without sorting, it might just have random order.
> 
> I don't think the statement is valid on python 3.7+ where the dictionary keeps the order of the keys as they were created. It is not guaranteed on python 3.6 but it's also implemented that way.

It's still a problem for newly added packages, and generally for producing the same output given equivalent data structure. While the key order is preserved it's irrelevant for dictionary comparison.

dmach:
Fixed in #2126

PR #2121, opened 2026-04-23
Fix config breakage.

hramrach:
Revert "Fix oscrc lookup: prefer XDG location, warn if there are multiple configs"

This reverts commit e2bd20cd8acdb461752c211827cfa5b9e59aca11.


hramrach:
The search order inversion also breaks tools that want to make use of the osc configuration to access OBS.

PR #2126, opened 2026-04-29
Modify 'git-obs file maintainership migrate' to filter out empty entries, ident by 2 and sort entries

dmach:


hramrach:
I don't really understand the need for those parameters.

In what protocol is a null value required?

In what case are unsorted keys required?

Adding those options makes the code more complex, and I don't see when those options would be needed.

hramrach:
Also #2118 now passes tests while this still does break some.

ncounter (approved):
👍

PR #2150, opened 2026-05-19
gitea_api: pr: expose 'merged_at' field

g7:
It can be rather useful, especially to create statistics on the codebase.


dmach:
This is a no-brainer, merging immediately, thanks!

PR #2151, opened 2026-05-20
Implement support for dynamic buildrequires in 'osc build'

### LINKED ISSUE #1706: osc build needs to learn the 2-phase approach of OBS with %generate_buildrequires
**Feature Request**
RPM has, for a while now, known the section %generate_buildrequires, which can run any script/tool and extract BuildDependencies by whatever means it deems fit for purpose.

OBS itself supports this and runs a 2-phase progress on the spec file:
* Build the spec file, up to the %generate_buildrequires section
* Extract the output .src.rpm which contains the generated dependencies, and restart the build using the build-dep information provided by this.

It would be nice to build such packages locally - and not only by OBS.

There are a few test packages in https://build.opensuse.org/project/show/home:dimstar:meson

**Describe the solution you'd like**
OBS implemented support for this in https://github.com/openSUSE/open-build-service/commit/eb087cfb9ffa0eb4f012121f0ec914a517a6fd98 (and a few follow ups)

It would be great for `osc` to learn how to deal with that too

dmach:
Fixes: #1706

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -225,7 +225,7 @@ BUILD_TYPES: List[BuildType] = [
         recipes=["*.spec"],
         package_suffix="rpm",
         binary_packages_paths=["RPMS/**/*.rpm"],
-        source_packages_paths=["SRPMS/**/*.src.rpm"],
+        source_packages_paths=["SRPMS/**/*.src.rpm", "SRPMS/**/*.nosrc.rpm"],
         prefer_packages_paths=["**/*.rpm"],
         prefer_packages_exclude_paths=["*.src.rpm", "*.nosrc.rpm", "*.patch.rpm", "*.delta.rpm"],
     ),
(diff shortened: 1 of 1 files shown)


darix:
woohoo +1

darix (approved):

dmach:
I got additional feedback and I should:
- [ ] check the exit codes; we should use only 11
- [ ] don't strip the versions in dependencies, use the deps as provided
- [ ] store the generated BRs on disk (similarly to buildconfig); properly handle --clean
- [ ] use /.build.packages/OTHER/_generated_buildreqs instead of parsing deps from nosrc.rpm
- [ ] when e.g. iteration 6 fails, don't start over, use previously stored deps; but make sure that re-using the buildroot to build another package in a project works fine

mlschroe:
(I think build's exit code is 9, not 11)

darix:
it checks for 9 and 11

PR #2159, opened 2026-06-01
Make TestGitEditor more portable

AMDmi3:
Do not use hardcoded editor paths, as these differ on e.g. BSD, so this always happen:

```
SKIPPED [1] tests/test_git_editor.py:16: nano is not /usr/bin/nano
SKIPPED [1] tests/test_git_editor.py:31: nano is not /usr/bin/nano
SKIPPED [1] tests/test_git_editor.py:21: nano is not /usr/bin/nano
SKIPPED [1] tests/test_git_editor.py:26: nano is not /usr/bin/nano
SKIPPED [1] tests/test_git_editor.py:11: nano is not /usr/bin/nano
```

### DIFF
--- a/tests/test_git_editor.py
+++ b/tests/test_git_editor.py
@@ -6,17 +6,21 @@ import unittest.mock
 from osc.gitea_api.common import get_editor_command
 
 
+vim_executable = shutil.which("vim")
+nano_executable = shutil.which("nano")
+
+
 class TestGitEditor(unittest.TestCase):
 
     def test_default(self):
         with unittest.mock.patch.dict(os.environ, {"EDITOR": ""}):
             c = get_editor_command("test")
-        self.assertEqual(c, ["/usr/bin/vim", "test"])
+        self.assertEqual(c, [vim_executable, "test"])
 
     def test_custom(self):
         with unittest.mock.patch.dict(os.environ, {"EDITOR": "nano"}):
             c = get_editor_command("test")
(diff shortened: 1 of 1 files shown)


AMDmi3:
CI failure is unrelated

tests/test_git_editor.py
```diff
@@ -34,10 +38,10 @@ def test_custom_with_miltiword_params(self):
...
 
 
-if shutil.which("nano") != "/usr/bin/nano":
-    TestGitEditor = unittest.skip("nano is not /usr/bin/nano")(TestGitEditor)
-elif shutil.which("vim") != "/usr/bin/vim":
-    TestGitEditor = unittest.skip("vim is not /usr/bin/vim")(TestGitEditor)
+if nano_executable is None:
+    TestGitEditor = unittest.skip("nano executable missing")(TestGitEditor)
+elif vim_executable is None:
+    TestGitEditor = unittest.skip("vim executable missing")(TestGitEditor)
```
dmach: I know that you've only improved the existing code,
but looking at this, I'm wondering if this is a correct approach at all.
Shouldn't we rather skip individual test methods rather than the whole class?
Would you mind turning these lines into `@skipIf(...)` decorators of the relevant methods?

dmach (changes requested):

tests/test_git_editor.py
AMDmi3: Great suggestion, done

PR #2180, opened 2026-07-20
fix gitea api IssueTimelineEntry parsing

TeodoraPavlova:
ValueError: Unable to instantiate model IssueTimelineEntry from the following data: None


dmach:
@TeodoraPavlova the patch looks good, but could you also add a descriptive comment in the code so it's clear why we're filtering using `isinstance()`? If you know the reason why this happens, please mention that in the comment as well. It would be good to document known gitea quirks on our end.

Issue #2137, opened 2026-05-07
`osc` commands no longer honor API info from project/package (OBS SCM)


e4t:
**Describe the bug**
Some `osc` commands (like `results`) no longer use the API information provided in the `.osc` subdirectory.

**Versions**
- Operating system:  Leap 16.0
- Package: osc-1.26.0-bp160.475.1

**To Reproduce**
1. Run:
```
osc -A https://api.suse.de co Devel:HPC:apptainer
cd Devel\:HPC\:apptainer/apptainer.SUSE_SLE-15-SP6_Update
osc info
```
Observe that the API_URL is set correctly.
2. now run:
```
osc results
```
Observe the 404 error
3. retry with:
```
osc -A https://api.suse.de results
```
observe the desired answer being given.

**Expected Behavior**
Step 3. should return the results Step 4. did.

dmach:
I'm afraid I'm unable to reproduce this with the latest osc from master and the osc version from the 1.26.0 git tag.
I'm getting output with 2 lines even without providing `-A ...` explicitly.

e4t:
Well, `do_results()` in `commandline.py` immediately calls `get_api_url()` which gets the api url from the configuration. This is fine - except when in a (traditional OBS SCM) 'project' or 'package' directory.
This change fixes it for me:
```
    def do_results(self, subcmd, opts, *args):
        [...]
        args = slash_split(args)

        apiurl = self.get_api_url()
        [...]
        project = None
        package = None

        if len(args) == 0:
            store = osc_store.get_store(Path.cwd())
            project = store.project
            if store.is_package:
                package = store.package
+           apiurl = store.apiurl
```
At least in my package this line is missing.

e4t:
To fix this fully:

1. check for command line override with `-A`
2. if no apiurl is set, try to get it from the project/package store.
3. if still no apiurl, get it from the config.
Command line override and config should not be treated equally as it breaks common expectations - and the behavior over many years.

BTW: `osc log` has the same problem. It doesn't check for a project/package store.

lethliel:
Can you provide a zip file containing your working copy? So that we can reproduce this behaviour? 

When I do the steps to reproduce this it works without problems.

e4t:
Ok, @lethliel  figured it out. I had `apisrv` not `apiurl` set in my oscrc. This is essentially a 'no-op'. It started working once `apiurl` is set.
The confusing part about this is that apiurl can be set to any - even if I set:
```
[general]
[...]

PR #2182, opened 2026-07-28
Error out on deprecated apisrv option is set

lethliel:
check if `apisrv=` is set in the config and error out with: 

`error: apisrv in config is deprecated, use apiurl instead`

This prevents unexpected behavior whan falling back to the default apiurl. 

Prevents #2137

### DIFF
--- a/osc/commandline.py
+++ b/osc/commandline.py
@@ -192,6 +192,9 @@ class OscMainCommand(MainCommand):
         if conf.config["show_download_progress"]:
             self.download_progress = create_text_meter()
 
+        if conf.config["apisrv"]:
+            raise oscerr.ConfigError("apisrv in config is deprecated, use apiurl instead", "")
+
         if not args.apiurl:
             self.parser.error("Could not determine apiurl, use -A/--apiurl to specify one")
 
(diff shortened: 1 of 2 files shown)


osc/commandline.py
```diff
@@ -192,6 +192,9 @@ def post_parse_args(self, args):
         if conf.config["show_download_progress"]:
             self.download_progress = create_text_meter()
 
+        if conf.config["apisrv"]:
+            self.parser.error("apisrv in config is deprecated, use apiurl instead")
```
dmach: I'd rather raise a different exception, `parser.error` relates to command-line args.

osc/conf.py
```diff
@@ -511,6 +511,15 @@ def apiurl_aliases(self):
...
 
+    apisrv: Optional[str] = Field(
+            default=None,
+            description=textwrap.dedent(
+                """
+                Deprecated option. Use apiurl instead.
+                """
+            ),
+        )  # type: ignore[assignment]
+
```
dmach, line 522: I was wondering if this is needed at all, but it's actually a good idea. It's going to be rendered in the oscrc man page and we'll make clear that it's there but shouldn't be used any more.

dmach (changes requested):

osc/commandline.py
lethliel: Changed the exception to ConfigError

PR #2184, opened 2026-07-30
Use any avaiable -p/--prefer-pkgs packages in containers builds

### LINKED ISSUE #2183: No longer picking up local RPM packages in non-RPM build
**Describe the bug**

When building a package and using `--prefer-pkgs` to include local packages, recent versions of osc only pick up packages of the same build type.
This breaks inclusion of local RPM packages in Kiwi builds, for example.

**Versions**
- Operating system: openSUSE Tumbleweed
- Packages: osc-1.27.2-1.2.noarch

**To Reproduce**
1. osc co home:crameleon:test1; cd home:crameleon:test1
2. osc mkpac foo; cd foo; create foo.spec with minimal lines for a valid RPM:
```
Name: foo
Version: 1
Release: 0
License: MIT
Summary: text

%description
text

%files

%changelog
```
4. `osc build -k /tmp/my-rpms -p /tmp/my-rpms 16.0 foo.spec`
5. cd ../container
6. `osc build -k /tmp/my-rpms -p /tmp/my-rpms containers container.kiwi`6a. it will say `unresolvable: nothing provides foo` with osc 1.27.2
6b. it will say `Evaluating preferred packages .. foo (/tmp/my-rpms/foo-1-0.x86_64.rpm)` and start the build fine with osc 1.22.0 (the dummy container itself will fail, point is the build starting and considering the local RPM package)

This might be related to

```
b44928e50179 ("Consolidate build types in build.py and commandline.py")

+            for build_type in BUILD_TYPES:
+                if build_type.name != type:
+                    continue
```
as removing that "continue" makes it work again, but the code changed since.

dmach:
Fixes: #2183

### DIFF
--- a/osc/build.py
+++ b/osc/build.py
@@ -614,6 +614,14 @@ def get_repo(path):
 
 
 def get_prefer_pkgs(dirs, wanted_arch, type, cpio):
+    """
+    Gather all packages from directories specified via -p/--prefer-pkgs
+    to make them available in the build environment.
+
+    We're gathering all known package types, because they can be used
+    for building both packages and containers and we don't know
+    what's going to be used in the build recipe in the end.
+    """
     paths = []
     repositories = []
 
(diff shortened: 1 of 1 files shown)


Vogtinator (approved):
Just hit this bug as well, PR works here

